PR #191, opened 2015-06-11
62 z remapping uv grids

nicjhan:
It does the remapping for all variables on u, v grids. Altogether there are > 100 new diagnostics being remapped to Z coordinates. 

This branch includes 62-z-ALE-remapping so it could be merged instead of 62-z-ALE-remapping. There a couple things that I'm not sure about in this branch:

1)  it doesn't handle variables on the B grid, I'm not sure whether I'm doing this properly, the remapping fails with large errors. So I've just disabled those axes for the time being
2) I've had to mess with the remapping error checking a bit. I've introduced maximum (hard coded) relative and absolute errors, it's not very nice.


nicjhan:
This PR is now ready to review and merge. There should be no results change.

src/core/MOM_continuity.F90
```diff
@@ -72,7 +72,7 @@ module MOM_continuity
 
 contains
 
-subroutine continuity(u, v, hin, h, uh, vh, dt, G, CS, uhbt, vhbt, OBC, &
+subroutine continuity(u, v, hin, h, uh, vh, dt, G, CS, diag_cs, uhbt, vhbt, OBC, &
```
adcroft: Could you move the call to diag_update_target_grids() to just after the call to continuity() so that we don't need the diag_CS type passsed into continuity().

This should give just a few less calls to diag_update_target_grids() but more importantly keeps continuity look more like a computational kernel without diagnostic.

nicjhan:
I have moved the calls to diag_update_target_grids() out of continuity(). This has the benefits of keeping the continuity() interface clean, and reducing unnecessary calls to diag_update_target_grids().

@adcroft can you please review etc. when you get time.

src/ALE/MOM_remapping.F90
```diff
@@ -534,15 +534,22 @@ subroutine remapping_core( CS, n0, h0, u0, n1, dx, u1 )
...
-    write(0,*) 'h0=',h0
-    write(0,*) 'hf=',h0(1:n1)+dx(2:n1+1)-dx(1:n1)
-    write(0,*) 'u0=',u0
-    write(0,*) 'u1=',u1
-    write(0,*) 'total HU0,HUf,f-0=',totalHU0,totalHU2,totalHU2-totalHU0
-    write(0,*) 'err0,errF=',err0,err2
-    call MOM_error( FATAL, 'MOM_remapping, remapping_core: '//&
-         'Total stuff on h0 and hF differ by more than roundoff' )
+    ! Maximum relative error
+    if (abs(totalHU2-totalHU0) / totalHU2 > 1e-09) then
```
adcroft: I gather the remapping was not accurate enough? I was quite proud of the estimate of round-off and haven't found the criteria to fail so far. Perhaps a left over from debugging?

PR #208, opened 2015-07-30
Fatally error with no ocean

angus-g:
An incorrect topography configuration could cause there to be no land area. `compute_global_grid_integrals` computes an inverse of the total ocean area, which may result in a divide-by-zero. Instead of failing silently, we raise a fatal error.


adcroft:
Good idea.

PR #215, opened 2015-08-13
Additional global average flux diagnostics

jkrasting:
Original scalar flux diagnostics were implemented as
integrals. Global average versions added for convenience.
Diags will be especially useful in monitoring coupled runs.

Confirmed that timestats.gnu did not change for global_ALE/z,
but full testing is needed.

### DIFF
--- a/src/core/MOM.F90
+++ b/src/core/MOM.F90
@@ -555,6 +555,7 @@ type, public :: MOM_control_struct
   integer :: id_zossq    = -1
   integer :: id_volo     = -1
   integer :: id_ssh      = -1
+  integer :: id_ssh_ga   = -1
   integer :: id_sst      = -1 
   integer :: id_sst_sq   = -1
   integer :: id_sss      = -1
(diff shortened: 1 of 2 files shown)


src/core/MOM.F90
```diff
@@ -555,6 +555,7 @@ module MOM
   integer :: id_zossq    = -1
   integer :: id_volo     = -1
   integer :: id_ssh      = -1
+  integer :: id_ssh_ga   = -1
```
adcroft: Could you add id_sst_ga and id_sss_ga while you're at it?

jkrasting: They are already implemented as sosga and tosga.

--------------------------------
Dr. John Krasting
Physical Scientist (NOAA Federal)
NOAA/Geophysical Fluid Dynamics Laboratory
Biogeochemistry, Ecosystems, and Climate Group
Princeton University Forrestal Campus
201 Forrestal Road
Princeton, NJ 08540

P. (609) 452-5359
F. (609) 987-5063

adcroft: I forgot - didn't see it in this part of the code... Thx

PR #218, opened 2015-09-01
Fixes needed for the branch code to compile and run

nikizadehgfdl:
- The argument G need to be (inout) in the subroutines that lead to MOM_initialize_tracer_from_Z()
  I don't know how it fares with MOM6 development philosophy.
- The argument paramfiles is needed to be passed to initialize_MOM_generic_tracer()
- The call to MOM_initialize_tracer_from_Z() is added to the  initialize_MOM_generic_tracer() for the whole thing to work.


adcroft:
Bump

PR #238, opened 2015-11-06
Ray tracing

bmater:
Hi Alistair,
Here are the changes for my ray tracing stuff.
-Ben


src/diagnostics/MOM_wave_structure.F90
```diff
@@ -0,0 +1,776 @@
...
+            ! [-1/H(k-1)]e(k-1) + [1/H(k-1)+1/H(k)-lam_z]e(k) + [-1/H(k)]e(k+1) = 0,
+            ! where lam_z = lam*gprime is now a function of depth.
+            ! Frist, populate interior rows
+            do K=3,kc-1
+              row = K-1 ! indexing for TD matrix rows
+              lam_z(row) = lam*gprime(K)
+              a_diag(row) = gprime(K)*(-Igu(K))
+              b_diag(row) = gprime(K)*(Igu(K)+Igl(K)) - lam_z(row)
+              c_diag(row) = gprime(K)*(-Igl(K))
+              if(isnan(lam_z(row)))then  ; print *, "lam_z(row) is NAN" ; endif
```
adcroft: Function isnan() interface is missing. Where does it come from?

src/parameterizations/vertical/MOM_diabatic_driver.F90
```diff
@@ -25,7 +26,7 @@ module MOM_diabatic_driver
 use MOM_entrain_diffusive,   only : entrain_diffusive_end, entrain_diffusive_CS
 use MOM_EOS,                 only : calculate_density, calculate_2_densities, calculate_TFreeze
 use MOM_EOS,                 only : calculate_specific_vol_derivs
-use MOM_error_handler,       only : MOM_error, FATAL, WARNING, callTree_showQuery
+use MOM_error_handler,       only : MOM_error, FATAL, WARNING, callTree_showQuery, is_root_pe
```
adcroft: I removed this edit since it's not necessary/used.

src/diagnostics/MOM_wave_structure.F90
adcroft: I don't know how intel was able to compile/link but I have all compilers working now by adding
```
use MOM_checksums, only : isnan => is_NaN
```
which uses our own implementation of the `isnan` function.

Issue #243, opened 2015-11-25
MOM6 can not reopen existing NetCDF files for appending


Hallberg-NOAA:
Due to changes that appear to have been introduced to mpp_open at some point for files that are to be appended, MOM6 is unable to continue writing to timestats.nc in multiple run segments. This occurs for multiple compilers. Work is underway to determine how widespread this problem is, how far back it goes, and how to correct it.

nikizadehgfdl:
Zhi has provided a fix for this issue in the shared code which seems to work.  [gitlab shared branch user/z1l/dev] (http://gitlab.gfdl.noaa.gov/fms/shared/commit/98184d18a22fd01079e1eb2ca6ca813979842eb3 ).

Hallberg-NOAA:
This no longer appears to be an issue with the version of FMS that is being used by MOM6 by default, and probably has not been a problem for a long time. This issue can be closed.

PR #244, opened 2015-11-25
can not reopen existing NetCDF file for append

nikizadehgfdl:
-This addresses issue #243
-To check the content of the file (via mpp_get_info and/or mpp_get_fields)
 it has to be opened in "read" mode.
-If the check is satisfactory then open the unit in "append" mode.
-This does the job. However there is a bug somewhere else that causes
 zero values to be appended to the timestats.nc for all variables!!

### DIFF
--- a/src/framework/MOM_io.F90
+++ b/src/framework/MOM_io.F90
@@ -319,18 +319,15 @@ subroutine reopen_file(unit, filename, vars, novars, G, fields, threading, timeu
   if (.not.exists) then
     call create_file(unit, filename, vars, novars, G, fields, threading, timeunit)
   else
-    if ((thread == SINGLE_FILE) .or. .not.G%Domain%use_io_layout) then
-      call open_file(unit, filename, MPP_APPEND, MPP_NETCDF, threading=thread)
-    else
-      call open_file(unit, filename, MPP_APPEND, MPP_NETCDF, domain=G%Domain%mpp_domain)
-    endif
+    !Open file to check the content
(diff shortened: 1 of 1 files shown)


src/framework/MOM_io.F90
```diff
@@ -344,6 +341,14 @@ subroutine reopen_file(unit, filename, vars, novars, G, fields, threading, timeu
 !      !    filename,vars%name,name);
 !      !call MOM_error(NOTE,"MOM_io: "//mesg)
 !    enddo
+    !
```
nikizadehgfdl: If I "call close_file(unit)" here, the model seg. faults later at line 895 of MOM_sum_output.F90 as if the array "CS%fields" was dissociated! 
It looks like fields(1:nvar) here is a pointer that goes away if the unit is closed!

adcroft: Is that because you need to `call mpp_get_fields()` after re-opening the file?

nikizadehgfdl:
Yes, this workaround does not work. We need to get the functionality back in the shared code (by uncommenting mpp_read_meta calls in APPEND mode in mpp_io_connect.inc, lines 516 and 556).

The reason this cannot be fixed in MOM6 is that the call mpp_get_fields(unit, field(1:nvar)) encapsulates the information about the input file "unit" in the array "field" , so the array "field" is gone (changed) as soon as "unit" is closed  (reopened).
Moreover the call mpp_get_fields returns a "field" with wrong content for a unit that has not been opened for read (e.g., it has   bogus values for number of variables in the file if the unit was opened in APPEND mode).   So, the strategy of first opening the file in read-only mode then in append mode does not work.

PR #269, opened 2016-04-21
Initialise GV%max_depth

angus-g:
The vertical grid structure GV has a field called max_depth, similar to that for the ocean grid structure G. However, this field is never initialised (nor currently used). We initialise GV%max_depth to G%max_depth.


adcroft:
@angus-g You are correct that GV%max_depth is not set. However, can you move the line `GV%max_depth = G%max_depth` into `MOM_initialize_coord()` which is where GV is being set up. This is called immediately after the call to `MOM_initialize_fixed()` in MOM.F90 where G%max_depth is set. Bob has been trying to separate out the vertical grid from the grid structure because it needs to be shared with SIS2 which has a different vertical grid.

You can just update this branch and the PR should update itself...

angus-g:
@adcroft Thanks for the hint, I didn't know about the restructuring. That makes the change a lot simpler!

PR #351, opened 2016-09-08
Mods for generic abiotic tracer routine

jkrasting:
- Compute sosga and pass to generic_tracer_source()
    since routine normalizes global alkalinity by the
    ratio of local to global SSS.
  - Add model time to generic_tracer_coupler_set call
  - Wrap MAX_FIELDS in ifndef block so that it can
    be modified via CPPDEFS
  - Include block from Jasmin John to reset the COBALT
    'cased' tracer to 0. for all layers k>=2

### DIFF
--- a/config_src/dynamic/MOM_memory.h
+++ b/config_src/dynamic/MOM_memory.h
@@ -34,7 +34,9 @@
                                !    NJPROC_ is the number of processors in the
                                !  y-direction.
 
+#ifndef MAX_FIELDS_
 #define MAX_FIELDS_ 50
+#endif
                                !    The maximum permitted number (each) of
                                !  restart variables, time derivatives, etc.
                                !  This is mostly used for the size of pointer
--- a/src/tracer/MOM_generic_tracer.F90
+++ b/src/tracer/MOM_generic_tracer.F90
@@ -60,6 +60,7 @@ module MOM_generic_tracer
   use MOM_grid, only : ocean_grid_type
   use MOM_io, only : file_exists, read_data, slasher, vardesc, var_desc
   use MOM_restart, only : register_restart_field, query_initialized, MOM_restart_CS
+  use MOM_spatial_means, only : global_area_mean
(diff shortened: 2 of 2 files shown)


Hallberg-NOAA:
I really dislike the idea that we are taking steps to permit the use of compile-line CPP defs. Compile line CPP defs are not automatically documented and are not readily reproducible across different approaches to compiling. The fact that we need this strongly suggests to me the need to revisit areas of the code where CPP definitions are required, and instead consider replacing them with structures like linked lists of blocks of entries. As a bare minimum interim step, we need to to add a log_param call documenting the value of MAX_FIELDS, similarly to how we document the values of STATIC_MEMORY_ and SYMMETRIC_MEMORY_ in the MOM_parameter_doc.layout files.

adcroft:
I totally agree that a dynamic allocation method is where we should go but that would require a significant overhaul of the tracers. In the meantime, I prefer this commit to the option of changing MAX_FIELDS_ to 300 for all configurations.

Line [`#define MAX_FIELDS_ 50`](https://github.com/NOAA-GFDL/MOM6/pull/351/commits/9df5b43b3a9c2a426c86ffdcb54d66d58b7190ef#diff-36b690a5243a4120d8b67292da75c964R38) has been in MOM_memory.h for a very long time. The change @jkrasting made simply allows B-group to make `MAX_FIELDS_` larger for some configurations without cpp complaining. Infact, I think it would be a much better practice to protect all the `#define` statements with `#ifndef`'s.

Why does this PR have conflicts already?

jkrasting:
@adcroft, I don't know where the conflicts came from.  I did branch from an
older commit about 2 months ago.

-----
Dr. John Krasting, Physical Scientist
NOAA/Geophysical Fluid Dynamics Laboratory
Biogeochemistry, Ecosystems, and Climate Group
201 Forrestal Road, Princeton, NJ 08540
P. (609) 452-5359   |   F. (609) 987-5063

PR #397, opened 2016-12-15
Offline tracers in coupled configurations

ashao:
Tracers which require fluxes derived from the coupler (for example gas tracers) can now be run offline. To do this, much of the previous offline tracer code was reworked and moved from step_tracers into subroutines. Every call now to step_tracers does a little bit of vertical mixing as well as tracer source/sink functions. On the first step of the offline interval, new physical fields are read in. On the last step of the offline interval, horizontal advection and diffusion occurs.

This does not change answers in forward integrations of the Baltic, Baltic_ALE_z, double_gyre, and torus_advection_test cases. 

One remaining issue that has not been resolved is that offline tracers are not perfectly conserved in the very first time step if initialized from a cold start. I suspect that this has something to do with a flux from the coupler (likely around runoff) not being stored correctly during initialization. When started from a restart file, even when tracers are reinitialized, tracers are conserved to within numerical precision.


ashao:
Closing this PR due to a bad merge in a previous commit.

PR #438, opened 2017-03-15
Add routines to check column integrals within roundoff

ashao:
With @alistair, two new routines were added to MOM_debugging that can be
used to check the column integral of a field matches either a given value
(check_column_integral) or the column integral of another field
(check_column_integrals).


adcroft:
@ashao, I suspect @alistair will be surprised to be notified about these fortran routines!

PR #476, opened 2017-04-23
OBC registry

kshedstrom:
Several things:

- OBC registry type now used by several user cases.
- Deleted OBC%ubt_outer and friends, now using segment variables to load into BT_OBC.
- Obsoleted READ_OBC_XX.
- Added shelfwave test (looks horrible, oh well).


kshedstrom:
By the way, this changes answers for CCS. It does so because Matt's code loads up the segment variables while the old barotropic code was using OBC%ubt_outer instead.

adcroft:
OK, this looks right. I'll handle the PR today.

Looking at the correlation of OBC and update_OBC_CSp in arguments makes it seem that there is redundancy but as you point out, there is a circular dependency in the types themselves. Idea for the future (i.e. not needed now), we could rename things if we think of MOM_boundary_update.F90 as an "OBC driver" and the current MOM_open_boundary.F90 as the "segment provider".

kshedstrom:
There might be a way to have OBC inside the update_OBC_CSp instead. Perhaps
rename it first. I agree about saving it for later though.

adcroft:
Exactly, and agreed.

--
Dr Alistair Adcroft    (Alistair.Adcroft@noaa.gov)
Princeton University           Tel: (609) 987-5073
NOAA/GFDL, 201 Forrestal Road, Princeton, NJ 08540

PR #539, opened 2017-07-01
Re-factor of MOM_lateral_mixing_coeffs.F90 (VarMix)

### LINKED ISSUE #522: FGNV streamfunction for running MOM6 in Neverland configuration
I get the following errors when setting `KHTH_USE_FGNV_STREAMFUNCTION = True'.

FATAL from PE     0: cg1 must be associated when using FGNV streamfunction.

adcroft:
- VarMix was originally for a crude implementation of a Visbeck et al.
  diffusivity formula. It has become a catchall for resolution functions,
  stored slopes and length-scales. It infact has never contained an actual
  diffusivity. The meaning of the parameter USE_VARIABLE_MIXING was original
  "use Visbeck" but has really meant "allocate and calculate everything in
  this module even if not being used". This has now been cleaned up...
- USE_VARIABLE_MIXING still exists but does nothing.
  - It will be made obsolete shortly.
- Visbeck related fields (VarMix@L2v, ...) are allocated only if
  KhTh_Slope_Cff>0 or KhTr_Slope_Cff>0.
- The Eady growth rates are now allocated and calculated if
  either of Visbeck or MEKE schemes are in use.
- The field VarMix%Rd_dx_h is allocated and calculated if either the
  diagnostic is requested or the field needed "Passivity" or the
  resolution functions.
- The wave speed VarMix%cg1 is calculated only if needed but which
  includes resolution fns, Rd_dx, and FGNV stream fn.
  - Closes #522.
- Added logic in tracer_hordiff() to avoid referencing VarMix%L2v or
  VarMix%SN_v when VarMix%KhTh_Slope_Cff==0.
- Added logic in tracer_hordiff() to avoid referencing VarMix%L2v or
  VarMix%SN_v when VarMix%KhTr_Slope_Cff==0.

- This commit affects available_diags and MOM_parameter_doc for some
  experiments but does not changes answers.

### DIFF
--- a/src/parameterizations/lateral/MOM_lateral_mixing_coeffs.F90
+++ b/src/parameterizations/lateral/MOM_lateral_mixing_coeffs.F90
@@ -45,6 +45,15 @@ type, public :: VarMix_CS ;
                                    !! for layer thicknesses. In addition, masking at coastlines was not
                                    !! used which introduced potential restart issues.  This flag will be
                                    !! deprecated in a future release.
+  logical :: calculate_cg1        !< If true, calls wave_speed() to calculate the first
+                                  !! baroclinic wave speed and populate CS%cg1.
+                                  !! This parameter is set depending on other parameters.
+  logical :: calculate_Rd_dx      !< If true, calculates Rd/dx and populate CS%Rd_dx_h.
+                                  !! This parameter is set depending on other parameters.
+  logical :: calculate_res_fns    !< If true, calculate all the resolution factors.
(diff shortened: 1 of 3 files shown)


adcroft:
> ... but I do wonder why we use a "create_group_pass" and then a "do_group pass" ...

Agreed. I didn't add it, just carried it forward. There seems to have been a widespread replacement of single pass_var() with group passes. I prefer pass_var() when there's just the one field.

Issue #383, opened 2016-11-03
ocean-ice tests crash with Intel16 on c4


nikizadehgfdl:
dev/master/2016.10.25

I get the following crash when I try to run any ocean-ice experiment on c4 with Intel16 executable (compiled on gaea13).
```
forrtl: severe (174): SIGSEGV, segmentation fault occurred
Image              PC                Routine            Line        Source

fms_MOM6_SIS2_com  000000000077E218  ocean_model_mod_m         728  ocean_model_MOM.F90
fms_MOM6_SIS2_com  000000000077B05D  ocean_model_mod_m         488  ocean_model_MOM.F90
```

The experiments run fine on c3 with Intel15.

I tried one experiment (1/4 degree baseline) on c4 with the c3 executable (Intel15)   and it ran fine!

I'll dig in more, but here's my Intel16 exec if you want to try it on c4 

```
/lustre/f1/unswept/Niki.Zadeh/verona_mom6_2016.10.25/MOM6_SIS2_compile/ncrc4.intel16-prod-openmp/exec/fms_MOM6_SIS2_compile.x
```

nikizadehgfdl:
I think this might be an Intel16 compiler issue because: 

1. The above crash happens only when I compile MOM6 with -openmp -O3 with Intel16 (and use 1 OMP thread).

2. I cannot even compile with -openmp -O2, the compiler panics like:
```
  catastrophic error: **Internal compiler error: segmentation violation signal raised** Please report this error along with the circumstances in which it occurred in a Software Problem Report.  Note: File and line given may not be explicit cause of this error.
compilation aborted for /lustre/f1/unswept/Niki.Zadeh/verona_mom6_2016.10.25/OM4_libs_compile/src/mom6/src/MOM6/src/framework/MOM_checksums.F90 (code 1)
``` 

3. There are no such issues when I compile/run with Intel15 with or without -openmp , or  Intel16 without -openmp.

nikizadehgfdl:
Coming back to this, it is still happening in dev/master/2017.05.26, only the line numbers for traceback have changed.
```
forrtl: severe (174): SIGSEGV, segmentation fault occurred
Image              PC                Routine            Line        Source
fms_MOM6_SIS2_com  000000000073C880  ocean_model_mod_m         803  ocean_model_MOM.F90
fms_MOM6_SIS2_com  00000000007381B2  ocean_model_mod_m         507  ocean_model_MOM.F90

```

[...]

PR #589, opened 2017-09-12
Closes issue #383, fixes an openmp crash

nikizadehgfdl:
- Coupled models in which MOM6 is compiled with -openmp are crashing with a traceback

  forrtl: severe (174): SIGSEGV, segmentation fault occurred
Image              PC                Routine            Line        Source
fms_MOM6_SIS2_com  000000000073C880  ocean_model_mod_m         803  ocean_model_MOM.F90
fms_MOM6_SIS2_com  00000000007381B2  ocean_model_mod_m         507  ocean_model_MOM.F90

- The problem seems to be the "present(patm)" statement inside an ij loop
  even when the source file that has no !$OMP directives!!
  Although it's apparently the compiler being confused,
  taking the "present" outside the loop seems to fix the problem.
  And one might argue is more efficient.

- I also took the if (associated(state%frazil)) statement outside the loop to be more efficient,
  although the compiler did not have problem with that one.

- The crash occured with all versions of Intel on c3/c4 and theta.

### DIFF
--- a/config_src/coupled_driver/ocean_model_MOM.F90
+++ b/config_src/coupled_driver/ocean_model_MOM.F90
@@ -801,12 +801,20 @@ subroutine convert_state_to_ocean_type(state, Ocean_sfc, G, use_conT_absS, patm,
 
   do j=jsc_bnd,jec_bnd ; do i=isc_bnd,iec_bnd
     Ocean_sfc%sea_lev(i,j) = state%sea_lev(i+i0,j+j0)
-    if (present(patm)) &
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA:
Good catch, @nikizadehgfdl.  Unfortunately, this PR is going to have some variable naming conflicts with PR #587, which went in earlier today. Because this is a relatively small and self-contained PR, lets hold off on fixing it for now and revise it after PR #587 (which has already passed the pipeline tests and is ready to go) is completed.

PR #598, opened 2017-09-20
Added 'F' input_filename option

Hallberg-NOAA:
Added the option to set the input_filename to 'F' in MOM_input_nml to use the
    presence or absence of the MOM restart files to determine whether this is a new
    run for the ocean.  This is a potentially risky mode of operation, but it
    matches the use for other FMS component models.


adcroft:
Pipeline is at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2496

PR #607, opened 2017-09-21
Update diagnostic target grid for ALE remapping tendencies

ashao:
Diagnostics in remap_all_state_vars were using the pointer to the main
state variable 'h' which is not updated until the end of ALE_main. To fix
this, first a call to diag_update_remap_grids is made using the new
thicknesses. Second, post_data calls also needed to have the new layer
thicknesses.


adcroft:
Fixed code is being tested in https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2515

PR #616, opened 2017-09-26
Fix i,j range when updating diagnostic grids

ashao:
When updating the diagnostic grids for remapping, every column on the data
domain was being updated. This led to a scenario when called from ALE_main,
where there were garbage values 2 steps into the halo. The diagnostic
remapping does not need information more than one step into the halo and
the ranges of i and j have been updated.


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2552

PR #621, opened 2017-09-27
Generates cell method "area" from axis methods

adcroft:
- We had coded up a general implementation of the cell methods per axis
  from which anything could be done. Apparently CMIP wants an area
  cell method which is not general.
- This commit adds an area cell method based on the x,y methods:
  - "x:mean y:mean" generates "area:mean";
  - "x:sum y:sum" generates "area:sum";
  - other methods for x,y do not generate and area method.
- I also hand coded the area methods for the static variables since we have
  not generalized the approach for static diagnostics.
- Also added methods for hfgeo which is static and had been missed.


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2557

StephenGriffies:
Very good.

PR #625, opened 2017-10-02
Optional 2-d Mask for SST/SSS restoring

MJHarrison-GFDL:


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2600

PR #631, opened 2017-10-18
Makes MCT driver independent of modules inside coupled_driver

gustavo-marques:
MCT driver is now self-contained (i.e., it no longer depends on modules inside coupled_driver). This PR also includes additional changes that are summarized below:

* Fixes inequality sign in `MOM_EOS.F90` to avoid division by zero, https://github.com/NCAR/MOM6/pull/24

* Corrects axis types for outputs in the ice_shelf module, https://github.com/NCAR/MOM6/pull/24

* Adds capability to read/write restarts files following the CESM convention

* MCT driver is now fully Doxygenized. Additional documentation explaining fluxes exported/imported to/from MCT has been added at the end of the module.

* These commits do not change answers obtained with the latest dev/master.


Hallberg-NOAA:
I have quickly examined the numerous changes with this PR, and I agree that these changes are a good idea and should be merged (assuming the testing reveals no answer changes).

  The one point of concern that I would have is that in removing the public interface to get_state_pointers from ocean_model_MOM, we might inadvertently break some other coupler that is using this interface. I realize that this interface was added in July specifically to support MCT, and that this call is _not_ being used by the FMS coupler or any other coupler that I am aware of, but there is a danger that in removing this interface we may impact someone else. In this case, I think that we will be OK, but in general we do need to be cautious about removing public interfaces.

gustavo-marques:
@Hallberg-NOAA thanks for point this out and reminding that one should be cautious about removing public interfaces. As you mentioned, get_state_pointers was introduced to support MCT and that's why I removed it from ocean_model_MOM. I've tested the code on Cheyenne, using MOM-examples, and obtained answers identical to the latest record from https://github.com/adcroft/Gaea-stats-MOM6-examples.

adcroft:
In this case I think it is safe to remove - I recall saying, when we added the function, that this was a temporary measure. We should have said so in comments to be safe.

I'm just now catching up on PRs and about to evaluate this one. In a telecon later today, I'll ask @kshedstrom to take a look too.

kshedstrom:
No problem here - I'm not using the new interface.

PR #634, opened 2017-10-22
Geometric progression for ENERGYSAVEDAYS

jkrasting:
When running the coupled ESM, there is a desire to cut down on the verbosity of stdout of the production runs.  The proposal based on the prototype XMLs is to run with ENERGYSAVEDAYS = 30.

It is highly-beneficial during development of the model to examine the global energies saved out from calls to `write_energy` at the start of a model year.  The high quality of these diagnostics makes them a bellwether of whether or not the model is bit-reproducing an earlier simulation or if something catastrophically is going wrong early in a simulation.

One compromise is to start at a relatively high frequency and double the time interval between calls to `write_energy`

With this code update and following added to MOM_override:

```
#override ENERGYSAVEDAYS = 30.
#override ENERGYSAVEDAYS_GEOMETRIC = 1.
```

Global integrals would be computed at days 1, 2, 4, 8, 16, 30, 60, 90, ...

### DIFF
--- a/config_src/coupled_driver/ocean_model_MOM.F90
+++ b/config_src/coupled_driver/ocean_model_MOM.F90
@@ -283,20 +283,21 @@ subroutine ocean_model_init(Ocean_sfc, OS, Time_init, Time_in, gas_fields_ocn)
                  "The time unit for ENERGYSAVEDAYS.", &
                  units="s", default=86400.0)
   call get_param(param_file, mdl, "ENERGYSAVEDAYS",OS%energysavedays, &
-                 "The interval in units of TIMEUNIT between saves of the \n"&
-                 "energies of the run and other globally summed diagnostics.", &
+                 "The interval in units of TIMEUNIT between saves of the \n"//&
+                 "energies of the run and other globally summed diagnostics.",&
                  default=set_time(0,days=1), timeunit=Time_unit)
-
-  call get_param(param_file, mdl, "ENERGYSAVE_GEOMETRIC",OS%energysave_geometric,&
-                 "If true, globally summed diagnostics will be computed for \n"&
-                 "timestep intervals that increase in a geometric progession \n"&
-                 "until the first ENERGYSAVEDAYS interval is reached.", &
-                 default=.true.)
   call get_param(param_file, mdl, "ENERGYSAVEDAYS_GEOMETRIC",OS%energysavedays_geometric, &
-                 "The starting interval in units of TIMEUNIT for the first call \n"&
-                 "to save the energies of the run and other globally summed diagnostics. \n"&
+                 "The starting interval in units of TIMEUNIT for the first call \n"//&
+                 "to save the energies of the run and other globally summed diagnostics. \n"//&
                  "The interval increases by a factor of 2. after each call to write_energy.",&
-                 default=set_time(seconds=43200), timeunit=Time_unit)  ! 12 hours
+                 default=set_time(seconds=0), timeunit=Time_unit)
+
+  if ((time_type_to_real(OS%energysavedays_geometric) > 0.) .and. &
+     (OS%energysavedays_geometric < OS%energysavedays)) then
+         OS%energysave_geometric = .true.
+  else
+         OS%energysave_geometric = .false.
+  endif
 
   call get_param(param_file, mdl, "OCEAN_SURFACE_STAGGER", stagger, &
                  "A case-insensitive character string to indicate the \n"//&
@@ -377,6 +378,7 @@ subroutine ocean_model_init(Ocean_sfc, OS, Time_init, Time_in, gas_fields_ocn)
   ! write_energy_time is the next integral multiple of energysavedays.
   if (OS%energysave_geometric) then
     if (OS%energysavedays_geometric < OS%energysavedays) then
+
       OS%write_energy_time = Time_init + OS%energysavedays_geometric * &
         (1 + (OS%Time - Time_init) / OS%energysavedays_geometric)
 
--- a/config_src/coupled_driver/ocean_model_MOM.F90
+++ b/config_src/coupled_driver/ocean_model_MOM.F90
@@ -378,20 +378,9 @@ subroutine ocean_model_init(Ocean_sfc, OS, Time_init, Time_in, gas_fields_ocn)
   ! write_energy_time is the next integral multiple of energysavedays.
   if (OS%energysave_geometric) then
     if (OS%energysavedays_geometric < OS%energysavedays) then
-
-      OS%write_energy_time = Time_init + OS%energysavedays_geometric * &
-        (1 + (OS%Time - Time_init) / OS%energysavedays_geometric)
-
-      ! This line below, which patterns the way write_energy_time was originally coded,
-      ! yields inconsistent results depending on the case. Dividing by OS%energysavedays
-      ! introduces decimal values but the operators from FMS time manager are integer ops.
-
-      ! OS%geometric_end_time = Time_init + OS%energysavedays * &
-      !  (1 + (OS%Time - Time_init) / OS%energysavedays)
-
-      ! Algebraically equivalent statement below that yields consistent results
-      OS%geometric_end_time = OS%Time + OS%energysavedays
-
+      OS%write_energy_time = OS%Time + OS%energysavedays_geometric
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA:
Thank you, @jkrasting, for this useful suggestion. I like the direction that you are going with these ideas. 

  However, I would like to consider these ideas in the context of a broader evaluation of the design for managing the frequency of output to ascii and energy statistics files. In particular, the proposed changes only apply to the coupled_driver-driven version of MOM6, and not the ocean only version, nor to SIS2 (where similar considerations arise). Additionally, a broader redesign might consider whether calendar types might not be used to specify the frequency of output.  In some past cases, we chose to go from a 5 day interval (which was deemed to frequent) to a 73 day interval (which was almost useless) simply because 5*73=365 and we were using a no-leap calendar. Writing the output every 30 days presents the complication of having diagnostics that drift around the calendar with a 30-year (for a no-leap calendar) or irregular (with a Gregorian calendar) repeat cycle.

  Given that this pull request introduces new parameters and  changes the MOM_parameter_doc files, do you agree that it might make sense to quickly attempt a broader redesign of the management of this output frequency should be handled, instead of merging in this pull request now?

jkrasting:
@Hallberg-NOAA, sure, but my efforts on ESM4 at the moment.

Could we adopt this PR for the coupled model as a temporary solution until there is something for the solo driver and SIS2 as well?

Hallberg-NOAA:
If this were just a matter of porting the code to SIS2 and the solo driver, I would be fine with just accepting this PR, but I am hesitant to add new run-time parameters that might be immediately obsoleted.  How much of a problem would it be to just use "ENERGYSAVEDAYS = 30" to reduce the data volume until we have a chance to discuss this more fully and properly plan out the long-term solution?

  Looking at the code changes, I see that the new get_param calls omit the concatenation "//" between character strings.  I am concerned that this implied concatenation might not work on all compilers. Is there a relevant (and universally implemented) F90 standard at play here?

adcroft:
Although we haven't resolved the above debate, I was curious to see if any of the compilers complain about the missing // and none did: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2728.

adcroft:
I've tried re-launching three times but the something about this branch is causing experiments to hang: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/jobs/11306

adcroft:
I tried running interactively:
- The concurrent coupled configuration is the one that hangs
- Output frequency for all coupled models are messed up

jkrasting:
I don't think there are any alternate proposals in the near-term.  Running with ENERGYSAVEDAYS = 30. alone is not ideal, mainly since it does not provide high temporal resolution at the start of a run.  If we set it lower to something, like 1 day, then the stdout balloons.

To make this PR more palatable, I have pared it down to one new parameter, `ENERGYSAVEDAYS_GEOMETRIC`.  If this new param. is > 0. and less than `ENERGYSAVEDAYS`, the geometric progression for `write_energy` will be performed.  

The default for the new parameter is 0., so this should not impact any existing answers / timestats files.

@adcroft - can you try this commit in pipelines to see if it behaves better than the prior?

adcroft:
New pipe is at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2736

Hallberg-NOAA:
I am happy with this revision of the input parameters down to just one as a succinct solution that should fit neatly into a more general revision to the specification of the frequency of energy outputs.

Looking more closely at the code, I have a point of concern.  After the geometric period, the energy is supposed to be written at integer multiples of energysavedays from time_init. At lines 389 & 390 of the proposed revisions to ocean_model_MOM.F90, there are commented out lines that should do this, along with a cryptic comment about "inconsistent results". The uncomment line (393) is not guaranteed to give a value of geometric_end_time that is an integer multiple of energysavedays from time_init. Is this deliberate?

jkrasting:
Yes, it is intentional that geometric_end_time need not be an integer multiple.  The thought here is to facilitate comparison with existing runs.

In the older runs where ENERGYSAVEDAYS = 30., the dates where write_energy is called are:

   0, 30, 60, 90 ...

In this PR, the proposal is to simply add time resolution at the start of an experiment without altering the downstream dates.  With ENERGYSAVEDAYS still equal to 30., if ENERGYSAVEDAYS_GEOMETRIC = 1., the dates for calling write_energy are now:

  0, 1, 2, 4, 8, 16, 30, 60, 90 ...

Note how 32 is skipped and 30, 60, and 90 are still available for comparison.

By the way, did you see the code review I started on this block?  The old formulation does not always behave as intended:

```
OS%write_energy_time = Time_init + OS%energysavedays * &
      (1 + (OS%Time - Time_init) / OS%energysavedays)
```

In the test case I give above, ENERGYSAVEDAYS should be 10., but the model evaluates it as 5.

Hallberg-NOAA:
@jkrasting, there was something odd about your last comment, I don't see how ENERGYSAVEDAYS becomes 10 or 5 if ENERGYSAVEDAYS is set to 30.

I agree that if ENERGYSAVEDAYS = 30. and ENERGYSAVEDAYS_GEOMETRIC = 1., the dates for calling write_energy should be:

0, 1, 2, 4, 8, 16, 30, 60, 90 ...

  However, what happens if the model is restarted from day 5? Noting that we always write the initial energy, I think that the energies should be written at either 

5, 8, 16, 30, 60, 90, ...

or  at

5, 6, 7, 9, 13, 21, 30, 60, 90, ...

As I read the code, it looks to me like it will write out the energies at times

5, 6, 7, 9, 13, 21, 35, 65, 95, ...

Please test the new code starting from a restart at day 5. If the results are written at either of the first two sequences of times, I will have no further concerns about accepting this PR.

Again, I think that this PR is a very good idea, I just want to make sure that we get the right behavior.

Hallberg-NOAA:
Also please try starting from a restart at day 25 with ENERGYSAVEDAYS = 30 and ENERGYSAVEDAYS_GEOMETRIC = 1. I think that something like

25, 26, 27, 29, 30, 33, 41, 60, 90, ...

or

25, 30, 60, 90, ...

would be acceptable, whereas

25, 26, 27, 29, 33, 41, 55, 85, ...

would not.  In my view, the geometric progression should not cause any times where the regular progression would be saved to be skipped. Whether the additional geometric times should only start at the beginning of a run or at the start of all run-segments is something where I am willing to defer to your judgement.

jkrasting:
@Hallberg-NOAA, this is starting to make more sense to me now.  I was operating under the thought that ENERGYSAVEDAYS were not preserved across restarts and were relative time step zero of the current run segment.  If ENERGYSAVEDAYS are preserved across restarts, then yes I agree that my commit would not allow for this behavior.  Now I also see even more why ENERGYSAVEDAYS = 30. is a bad idea.  The Julian days will different from year to year.  

ENERGYSAVEDAYS = 73. with ENERGYSAVEDAYS_GEOMETRIC = 1. might be a much better combination.

jkrasting:
@Hallberg-NOAA, I think I have addressed your concerns with the latest commit and these four tests:

1. 45-day coupled test with ENERGYSAVEDAYS = 20., ENERGYSAVEDAYS_GEOMETRIC = 0.
2. 45-day coupled test with ENERGYSAVEDAYS = 20., ENERGYSAVEDAYS_GEOMETRIC = 1.
3. 13-day coupled test with ENERGYSAVEDAYS = 20., ENERGYSAVEDAYS_GEOMETRIC = 1.
4. 32-day coupled test, restarting from above, with ENERGYSAVEDAYS = 20., ENERGYSAVEDAYS_GEOMETRIC = 1.

I've attached the stats files for each of these tests.  In all cases, the original ENERGYSAVEDAYS interval is preserved.

[ocean.stats.case1.txt](https://github.com/NOAA-GFDL/MOM6/files/1423570/ocean.stats.case1.txt)
[ocean.stats.case2.txt](https://github.com/NOAA-GFDL/MOM6/files/1423571/ocean.stats.case2.txt)
[ocean.stats.case3.txt](https://github.com/NOAA-GFDL/MOM6/files/1423572/ocean.stats.case3.txt)
[ocean.stats.case4.txt](https://github.com/NOAA-GFDL/MOM6/files/1423573/ocean.stats.case4.txt)

Hallberg-NOAA:
This is looking good to me now.  Thanks @jkrasting.

Issue #641, opened 2017-11-01
z-remapped frazil heat diagnostic is wrong


StephenGriffies:
Frazil heating is diagnosed insided MOM_diabatic_driver.F90.  There are two places that the diagnostic needs to be computed since frazil is called at both the start and end of diabatic processes.  

Unfortunately, there are nontrivial differences between the native level grid version of the depth-summed version of frazil_heat_tendency, indicating there is a problem with the remapped diagnostics

native version of FRAZIL_HEAT_TENDENCY[k=@sum]  NOT EQUAL    FRAZIL_HEAT_TENDENCY[k=@sum] z-remapped

I suspect there is a thickness issue.  The subtle point is that the frazil diagnostic is called twice, corresponding to the twice called frazil heating adjustment.  Presumably the layer thickness changes between these two calls.  

Can I get some help from @ashao on this issue?  He has debugged this issue elsewhere, and I suspect his expertise could find the bug quickly...

As it stands, this bug makes the heat budget for ocean_z not close.  The native layer output does close.

ashao:
I'll have some time today to to take a look at this.

ashao:
I think I know the reason why this diagnostic is wrong. It's slightly more
complicated than just a wrong layer thicknesses. The frazil heat flux is
the sum of the first and second call to make_frazil, but the diagnostic is
only posted at the end. In between those calls the thicknesses have all
changed.

I think the right thing to do is to call post_data both times (with the
correct thicknesses) making sure that the diagnostic manager's averaging is
for half the diabatic timestamp, and also calculating the tendency for half
the timestep.

On Nov 1, 2017 9:31 AM, "Andrew Shao" <aeshao@gmail.com> wrote:

I'll have some time today to to take a look at this.

adcroft:
That sounds about right. I think it would still have to have yet another
carefully crafted thickness for the remapping.

ashao:
Ok, what I previously described is only partly what's going on. I just
tried doing the call to make_frazil once (and changing the tendency routine
too) and the remapped still isn't quite right. I'll keep you posted when I
come up with another lead.

ashao:
Tracked it down- the v_extensive flag also needed to be set. Because this issue has come up before, I suggest that we add a "v_extensive" line to available_diags in much the same way that cell_methods is attached to some of the diagnostics.

[...]

PR #644, opened 2017-11-02
Fixes frazil heatflux diagnostic when remapped

ashao:
This addresses #641. I'm pretty sure I set the averaging windows correctly, but could I get another set of eyes to double check? (I don't want to have accidentally disabled all the thermodynamic diagnostics!)

The frazil heatflux was not being conserved under remapping for two reasons
-The v_extensive flag needed to be set to true
-make_frazil is called multiple timss, but only the accumulated tendency is
posted at the end despite layer thicknesses changing in between calls. To
fix this, enable/disable averaging calls were added to organize the
diabatic algorithm into the following sections:
Time0 to Time0 + 0.5*dt: First call to make_frazil
Time0 to Time0 + dt: Everything but make_frazil
Time0+0.5 to Time0 + dt: Last call to make_frazil

### DIFF
--- a/src/core/MOM.F90
+++ b/src/core/MOM.F90
@@ -474,6 +474,7 @@ subroutine step_MOM(forces, fluxes, sfc_state, Time_start, time_interval, CS)
   integer :: i, j, k, is, ie, js, je, Isq, Ieq, Jsq, Jeq, nz, n
   integer :: isd, ied, jsd, jed, IsdB, IedB, JsdB, JedB
 
+  type(time_type) :: Time_end_thermo ! Time at end of the thermodynamics step for diagnostic averaging
   real :: dt              ! baroclinic time step (sec)
   real :: dtth            ! time step for thickness diffusion (sec)
   real :: dtdia           ! time step for diabatic processes (sec)
@@ -686,8 +687,7 @@ subroutine step_MOM(forces, fluxes, sfc_state, Time_start, time_interval, CS)
 
       ! The end-time of the diagnostic interval needs to be set ahead if there
       ! are multiple dynamic time steps worth of thermodynamics applied here.
-      call enable_averaging(dtdia, Time_local + &
-                                   set_time(int(floor(dtdia-dt+0.5))), CS%diag)
+      Time_end_thermo = Time_local + set_time(int(floor(dtdia-dt+0.5)))
 
       if (CS%debug) then
         call uvchksum("Pre set_viscous_BBL [uv]", u, v, G%HI, haloshift=1)
@@ -712,14 +712,13 @@ subroutine step_MOM(forces, fluxes, sfc_state, Time_start, time_interval, CS)
       call cpu_clock_begin(id_clock_thermo)
 
       ! Apply diabatic forcing, do mixing, and regrid.
-      call step_MOM_thermo(CS, G, GV, u, v, h, CS%tv, fluxes, dtdia)
+      call step_MOM_thermo(CS, G, GV, u, v, h, CS%tv, fluxes, dtdia, Time_end_thermo)
       do_pass_kv_turb = associated(CS%visc%Kv_turb)
 
       ! The diabatic processes are now ahead of the dynamics by dtdia.
       CS%t_dyn_rel_thermo = -dtdia
       if (showCallTree) call callTree_waypoint("finished diabatic_first (step_MOM)")
 
-      call disable_averaging(CS%diag)
       call cpu_clock_end(id_clock_thermo)
 
     endif ! end of block "(CS%diabatic_first .and. (CS%t_dyn_rel_adv==0.0))"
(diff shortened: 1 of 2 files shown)


ashao:
Also, do I get the distinction of submitting the commit from "sea water" (I'm on the ferry from Anacortes, WA to Sidney, B.C. currently 48.5386813N, 123.0182250W).

ashao:
Thanks for the helpful comments.

In OM4_05, I checked that the sum of the frazil heatflux in both the native
and remapped z coordinate was the same. For completeness, I'll do your
suggested check and other sanity checks in OM4_025 today.

I see what you're saying about the timestep. I'll sit with the arithmetic
to make sure that the diagnostic returns the heat flux averaged over the
entire thermodynamic timestep while being posted in two half averaging
intervals.

I'll add a comment later today once I've dealt with both issues.

adcroft:
Test started at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2801

StephenGriffies:
@ashao please point me to the directory where you have the diagnostics saved for OM4_05 or OM4_025.  I wish to be sure the 3d heat budget closes with the new frazil diagnostic code.  

Please be sure the diag_table has "months" changed to "days" so that you will generate some output.  

I will check on Tuesday...so no hurry.

ashao:
@StephenGriffies , I've convinced myself that you're right and that we should use the full DT_THERM to calculate the frazil tendency, but to split the averaging interval into two to get the correct z* tendency remapping.

I've included 2 days of OM4_025 output at /work/aes/OM4_025/frazil_diag

ashao:
Just checked the sum of frazil_heat_tendency and hfsifrazil and the two are off by a factor of 2. I'll delve into this more.

ashao:
So as it turns out the initial 0.5*dt_therm is the correct way to calculate the frazil heat flux so that the average over dt_therm is correct.

Define A1 and A2 as the change in heat for a cell which, for simplicity, assume does not change thickness. The diagnostic manager calculates averages as a weighted mean over the requested averaging interval (in this case dt_therm). In the case of two averaging intervals of equal length the weight for each term is 0.5*dt_therm/dt_therm = 0.5. The equation for the weighted-average tendency is then [A1/(0.5*dt_therm)]*0.5 + [A2/(0.5*dt_therm]*0.5} which simplifies to the expected tendency (A1+A2)/dt_therm.

For the sake of transparency I've reverted the previous commit instead of just rewriting the git history.

ashao:
Could you clarify if the output that is stored at /work/aes/OM4_025/frazil_diag no longer allows you to close the heat budget? In 19000101.ocean_month.nc, the frazil_heat_tendency_2d is equal to hfsifrazil and matches the vertical sum of the remapped frazil_heat_tendency in 19000101.ocean_month_z.nc to within ~0.0002W.

One change in this PR that I wanted to emphasize is that the original code accumulated the change in heat from both calls to make_frazil, calculated the flux by dividing by dt_therm, and made one call to post_data where the averaging window was the entire length of the thermodynamic timestep. Now within the thermodynamic timestep, post_data for frazil_heat_tendency is called twice (after each make_frazil call) where the averaging window is half of the thermodynamic timestep which is also the same dt used to calculate a flux from the change in heat. Does that help clarify the origin of the 0.5 factor?

StephenGriffies:
I verified that the frazil term now is consistent across the native and z-regridded output.  Thanks for your help.  

Unfortunately, the 3d heat budget still does not close on the z-grid.  The problem is in the upper 100m of the ocean.  I believe the fault is with the diagnostic boundary_forcing_heat_tendency and boundary_forcing_salt_tendency.  I looked at that code and am very puzzled with what it is doing. I will post a new issue.

So in short, I think you nailed the Frazil bug.  At least the frazil diagnostic is now self-consistent.  Thanks @ashao

ashao:
I had been able to get the 3D salt budget to close with the remapped
diagnostics, so I'm not sure that the boundary_forcing_heat/salt_tendency
terms are the culprit. Let me dig up the script and verify that salt still
does close.

Hallberg-NOAA:
This PR changed the available_diags files for some of the test cases, but they were not updated in MOM_examples.  Whenever a code change alters the output of out test cases, please be sure to update that output at the same time as we accept a PR.  Thank you.

ashao:
@Hallberg-NOAA : Could you clarify if you had detected that some diagnostics had been added/removed? This PR should only have changed cell_methods for frazil_heat_tendency diagnostics.

Also you raise some interesting questions about PR procedures. If a PR is answer-changing, we seem to have been doing a pretty good job of elucidating that in the PR comments. However, I have been pretty lax about mentioning if any of the other runtime-generated output changes. This inherently puts the burden on the person who accepts the pull-request to detect these changes and update MOM6-examples. Short of requiring that any code-contributor runs their own regression tests and always pairs a PR to MOM6 with one to MOM6-examples, perhaps the easiest way forward so that this responsibility is shared would be to require three bullets at the end of each PR:
- Changes ocean.stats: True/False
- Changes available_diags: True/False
- Changes MOM_parameter_doc.all: True/False

kshedstrom:
Some of us are only doing PRs for MOM6, not MOM6-examples, in part because
I'm not running all of MOM6-examples. I haven't been changing
available_diags, though.

StephenGriffies:
This is my fault, as I approved the pull request. 

But my assumption was that if all tests passed, then that meant all is fine,and ready to merge. Apparently this assumption is wrong. In which case the testing process needs further automation. 

Or at least I need to be properly scolded and told what to do before hitting the green buttons... @Alistair please advise...

PR #645, opened 2017-11-02
Override T and S for remapped transport diagnostics

ashao:
This PR addresses concerns that I raised in #642 

The remapping was working as expected for remapping transport diagnostics
onto a coordinate that does not depend on T/S of the model. However, for
transport on state-dependent coordinates (e.g sigma-2), the target
coordinate was being built using the pre-dynamics thicknesses, but the
current T/S.

This commit adds two optional arguments in diag_update_remap_grids to
override T and S. The override is currently only needed for the transport
diagnostics.

### DIFF
--- a/src/core/MOM.F90
+++ b/src/core/MOM.F90
@@ -506,9 +506,11 @@ subroutine step_MOM(forces, fluxes, sfc_state, Time_start, time_interval, CS)
     v, & ! v : meridional velocity component (m/s)
     h    ! h : layer thickness (meter (Bouss) or kg/m2 (non-Bouss))
 
-  ! Store the layer thicknesses before any changes by the dynamics. This is necessary for remapped
-  ! mass transport diagnostics
+  ! Store the layer thicknesses, temperature, and salinity before any changes by the dynamics.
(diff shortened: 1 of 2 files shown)


adcroft:
Test started at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2800

ashao:
Over the weekend I thought of a different way to do this if we start having more diagnostics that need this kind of treatment- Change diag_update_remap_grids to optionally return the updated diagnostic grids which can then be stored and overridden later. You then have to store NUM_DIAG_COORDS worth of 3D fields but at least it prevents you from having to recalculate the grids as well. Because currently only the transport diagnostics need this done, I don't think it's worth implementing, but something to keep in mind.

adcroft:
I was thinking the same thing. There's a catch which is we don't know shapes, how many coordinates there are, etc. so we'd need even more code to implement something that would need to be hidden in diag_mediator. :frowning_face:

We'll take the current approach for now. @Hallberg-NOAA is moving the diagnostics calls downward out of MOM6.F90 and after that we can revisit the overall approach.

Issue #654, opened 2017-11-18
Heat and salt budgets do not close on native grid


ashao:
Something in 6da5f6b9bfd73f07e4204bb634646e4ec7556199 changed such that heat and salt budgets no longer close when using native-space diagnostics.

StephenGriffies:
Nice catch!

PR #655, opened 2017-11-18
Add enable averaging before ALE routines

ashao:
Commit 6da5f6b inadvertently disabled posting of tendency terms due to
vertical remapping because the subroutine 'diabatic' handles the averaging
intervals for itself due to the frazil diagnostic needing to be posted
twice. All post_data calls in ALE_main were thus called while averaging
was disabled. To fix this, the call to enable_averaging is moved from the
top of step_MOM_thermo and down to where the ALE routines are called. This
resolves issue #654


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2856 (passed)

PR #660, opened 2017-11-21
Cray compilation found an issue with intent for G

nikizadehgfdl:
- G should be inout since it's passed to pass_var


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2865

PR #661, opened 2017-11-22
Fixes intent for args being updated in apply_sponge()

### LINKED ISSUE #659: Bug in MOM_sponge.F90: un-initialized ea & eb
I've noticed that the two variables, ea and eb, defined in subroutine apply_sponge under module MOM_sponge within MOM_sponge.F90, have not been initialized as 0 before they are first used in line 502 & 499 (for i-mean sponge setup):

` eb(i,j,k-1) = eb(i,j,k-1) + w_int(i,j,K)`
` ea(i,j,k) = ea(i,j,k) - w_int(i,j,K)`

as well as in the if-else clause of whether bulkmixedlayer is used in line 567 & 568 (if bulk mixed layer is used):

`eb(i,j,k) = eb(i,j,k) + wpb`
`ea(i,j,k) = ea(i,j,k) - wm`

or in line 615/616 (if bulk mixed layer is not used):

`eb(i,j,k) = eb(i,j,k) + wpb`
`ea(i,j,k) = ea(i,j,k) - wm`

adcroft:
- Issue #659 reported uninitialized values in apply_sponge().
  The pertinent code increments variables passed into apply_sponge()
  as arguments with intent(out). The arguments should be intent(inout)
  in which case the values are properly initialized.
- Note: none of the three compilers we use caught this mismatch
  and they are evidently treating intent(out) as intent(inout). Something
  to be wary of.
- Closes #659. @hlkong Thanks for report.

### DIFF
--- a/src/parameterizations/vertical/MOM_sponge.F90
+++ b/src/parameterizations/vertical/MOM_sponge.F90
@@ -364,10 +364,10 @@ subroutine apply_sponge(h, dt, G, GV, ea, eb, CS, Rcv_ml)
   type(verticalGrid_type),                  intent(in)    :: GV   !< The ocean's vertical grid structure
   real, dimension(SZI_(G),SZJ_(G),SZK_(G)), intent(inout) :: h    !< Layer thicknesses, in H (usually m or kg m-2)
   real,                                     intent(in)    :: dt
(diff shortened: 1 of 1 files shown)


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/2871

PR #672, opened 2017-12-15
Changes to support H rescaling

Hallberg-NOAA:
This PR covers a large set of changes that will prevent answers from changing when H_TO_M or H_TO_KG_M2 are rescaled by integer powers of 2.  All answers in the test cases are bitwise identical when H_TO_M is not rescaled.


Hallberg-NOAA:
Being tested by pipeline #3009.

PR #698, opened 2018-01-24
Restructure MOM diagnostics

Hallberg-NOAA:
This PR includes substantial rearrangement of the diagnostics that were formerly in MOM.F90, in preparation for even more dramatic restructuring of MOM.F90 and the addition of alternate time-stepping interfaces to MOM to address coupled instabilities.  All solutions and diagnostics should be unchanged by this PR, but number and order of diagnostics in the MOM available_diags files do change.


Hallberg-NOAA:
This PR is being tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/3157

PR #721, opened 2018-02-27
Add a parameter to avoid comparing restart checksums

nikizadehgfdl:
-Sometimes it is necessary to prevent the model from comparing the
 restart checksums to avoid crash (because the checksums do not match due to a false positive)
 For example if the restarts are from an original run with a different
 processor mask_table than the current run, the checksums won't match
 because they are mask_table dependent (since restart variable "h" has a non-zero value on land and
 a missing value on the masked pes).
 Eventually we have to fix the original issue of mask_table dependence of restart vars.
 But to let our current models continue over  a layout change (hence mask_table change)
 we need to be able to switch  the checksum comparisons off (e.g., at the first segment of the run after layout change).


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/3418

PR #736, opened 2018-03-16
Make grid structure intent(in) in adjust_area_mean_to_zero

angus-g:
There's never a modification of the grid structure parameter `G`, and having it as `intent(inout)` can prevent compilation, depending on whether `G` can be modified at the call site. This shouldn't affect the compilation or answers of any existing configurations.


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/3546

PR #803, opened 2018-06-29
Eliminate MOM_OCMIP2_CO2calc.F90

Hallberg-NOAA:
Eliminated the file MOM_OCMIP2_CO2calc.F90, which is not being called by any
MOM6 code.  This file violates numerous MOM6 coding and licensing conventions,
including some that break the automatic generation of documentation by MOM6, and
it is an unused duplicate of capabilities that are found in the generic tracer
code. All answers in test cases are bitwise identical.


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/4788

PR #812, opened 2018-07-10
Doxygen fixes for src/initialization

adcroft:
- Fixes doxygen errrors in src/initialization


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/4902

PR #816, opened 2018-07-11
More "final" doxygen fixes

adcroft:
Although the last PR was called "final" there are still doxygen issues to fix.
- Fixes the pre-processor problem in MOM_barotropic.F90
- Cleans up the lengh of titles/summary lines that appear in the full list of annotated types. Some of these had entire essays appearing as the summary.


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/4916

PR #833, opened 2018-08-14
Increase precision of time handling

Hallberg-NOAA:
Replaced the use of set_time and get_time with real_to_time_type and
time_type_to_real to increase the precision of handling time by allowing the
sub-second ticks to be used effectively, and thereby permitting the use of
fractional second timesteps within MOM6.  This modification could change answers
in cases where times are not an integer number of seconds.  However, all
existing test cases only use times that are an integer number of seconds, so the
answers are bitwise identical in the test cases.


Hallberg-NOAA:
This PR was tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/5367 and passed.

PR #854, opened 2018-09-26
Correct the grid used by the ice shelf for allocate_forcing_type

Hallberg-NOAA:
Use the correct grid type in allocate_forcing_type and allocate_mech_forcing
calls inside initialize_ice_shelf, and added error checks to verify that
compatible grids are being used by the ice shelf and ocean.  All answers in the
existing MOM6 test cases are bitwise identical, but this corrects a problem in
which symmetric memory ocean models fail for the ISOMIP test case unless the
ice shelf grid also uses symmetric memory.


Hallberg-NOAA:
This PR is being tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/5845.

PR #858, opened 2018-10-05
Bug fix to compute Rd_dx whenever MEKE is used

MFJansen:
The default config of MEKE uses Rd_dx to compute barotropic and bottom energy fractions.
 If Rd_dx not computed here, it will be set to zero in MEKE and as a result gamma_b=gamma_t=1.


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/5950

PR #867, opened 2018-10-24
(*)Fixed dimensional scaling for Boussinesq Flather OBCs

Hallberg-NOAA:
Corrected dimensional rescaling for Boussinesq Flather OBCs, so the code now
gives identical answers in the circle_OBCS test case for various values of
H_RESCALE_POWER and partly corrected problems that will arise with the
non-Boussinesq version of the Flather OBCs, although this mode still is not
working.  All answers in the test cases are bitwise identical.


Hallberg-NOAA:
This PR is being evaluated with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/6108.

PR #873, opened 2018-10-26
Adds patches made to dev/master candidate back to dev/gfdl

adcroft:
These five commits were patches made to dev-master-candidate-2018-10-15 in order to pass tests at EMC and NCAR.

- Several intent issues (need to understand why the three compilers we use at GFDL do not catch these?);
- Missing OpenMP directives (we skipped a step in when submitting the candidate!);
- A fix in the MCT cap.


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/6117

PR #909, opened 2019-04-16
Re-factor of Travis-CI usage

adcroft:
- Uses stages to parallelize testing process
- Adds a doxygen job that tests for doxygen errors
- Reduces dependency on other repositories
  - No longer uses scripts from MOM6-examples/tools/tests/Travis-MOM6/
  - Still uses configurations from MOM6-examples/ocean_only/
- Adds test building/using a debug executable
- Adds a .testing/ directory
- Uses the latest xenial image on Travis (with openmpi)
- Example of new report is at https://travis-ci.org/adcroft/MOM6/builds/520870218

Todo:
- Add code coverage (needs more experiments)
- Break dependence on MOM6-examples (long term project)
- Add other portable tests (e.g. parameter scaling, rotation, etc)

### DIFF
--- a/.testing/Makefile
+++ b/.testing/Makefile
@@ -1,0 +1,77 @@
+# Makefile steps to run on Travis-CI
+# e.g. make MEMORY_SHAPE=dynamic_symmetric REPRO=1 OPENMP=1
+
+# Versions to use
+FMS_COMMIT ?= xanadu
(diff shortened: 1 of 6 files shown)


marshallward:
Looks like I also have permissions to do PR merges.

PR #919, opened 2019-05-03
MEKE: Prevent div-by-zero by bathyT in beta calc

marshallward:
The calculation of beta in the MEKE module had an explicit division by
zero when computing the lateral topography derivatives, which were
raising floating point exceptions in the debug builds.  This occurs in
the MEKE_equilibrium and MEKE_lengthScales functions.

This issue was not observed in the production tests due to
MEKE_TOPOGRAPHIC_BETA always being set to zero.  When FPEs are disabled,
the 0 * (1./0.) operation produces a NaN which is passed to a max()
function, which ignores the NaN and always returns the other value.

We resolve this by explicitly checking for zero values in bathyT and
setting the topographic beta to zero when this term is zero.  While this
could potentially change the value of the Rhines scale, these values
only occur over land, which are in general masked, and should not affect
the calculation.

The unoptimized expressions were retained, but recommended changes which
reduce the number of divisions were included in comments.

No value changes were observed in our test suite, and the patch should
be bitwise reproducible.

Minor changes:
- We do not calculate the topographic beta term if the scaling factor,
  MEKE_TOPOGRAPHIC_BETA is zero

- the default value of beta was unset in MEKE_lengthScales was unset
  when CS%use_old_lscale is True, so we set this to zero.

- Minor whitespace and index syntax changes


Hallberg-NOAA:
This PR is being tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/7909 .

Hallberg-NOAA:
This PR was tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/7912, and passed (although there are MOM_parameter_doc files that need to be updated to reflect a recent change).

PR #921, opened 2019-05-09
* Bugfix: Double diffusive correction to Kd

marshallward:
The double-diffusive salinity correction to diffusivity was incorrectly
using an exponent in its term:

          Kd_lay(i,j,k-1) = Kd_lay(i,j,k-1) + 0.5**KS_extra(i,K)

The exponent has been replaced with a multiplication.


MJHarrison-GFDL:
Marshall,

Good catch!  Obviously none of the test cases are using double diffusion (I thought we had turned it on for OM4).

Hallberg-NOAA:
This PR is being tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/7983.

PR #952, opened 2019-07-17
(*) Remap opacity diagnostic to tanh()

marshallward:
This patch remaps the opacity diagnostic to a tanh function, i.e.

    op -> 1/L * tanh(op * L)

where L is arbitrarily set to 10^-10 (1 Angstrom).  For op << 1/L, the
diagnostic is nearly equivalent to the model opacity.  For values
comparable and larger than L, the diagnostic approaches 1/L, a
sufficiently large value to reproduce the effects of a divergent
opacity.

This allows us to safely manipulate and store the opacity while also
avoiding infinite values and floating point overflow.

This change will modify the opacity diagnostic, but should not affect
the dynamic state.


marshallward:
GitLab test: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/8494

marshallward:
GitLab tests have passed (excepting parameters, which is waiting on an update)

PR #979, opened 2019-08-19
WIP: using doxyrest for documentation

raphaeldussin:
just wanted to put it out there that the current doc pipeline doesn't work with more recent versions of sphinx and some dependencies haven't changed in years.

This is exploring an alternative solution for the doc pipeline. Some issues need to be worked out however.


raphaeldussin:
There's still some weird stuff going on but it's getting better: https://mom6-doctesting.readthedocs.io/en/doxyrest_doc/

The issues with parameters tables comes from the use of markdown style for tables in the F90 files. Rewritting in latex style fixes the problem.

adcroft:
Any progress on this exercise?

PR #989, opened 2019-08-30
TC reconfig

adcroft:
Some refactoring .testing
- Cleaned up .gitignore
- Reduced diag_tables
- Made tc3 smaller
- Added a variant of tc1 (tc1.a)


marshallward:
Change is in the test suite, so no Gaea testing is needed.

Still not sure why it's not reporting a relative code coverage to dev/gfdl, but at least we're getting coverage reports now.

PR #1033, opened 2019-11-18
User/jml/operation fms2 the reckoning

wrongkindofdoctor:
This is a draft PR for the MOM6 code with the new fms-io calls so that MOM6 developers can see what the preliminary round of changes looks like. The code compiles with gnu7 in debug on gaea, but still needs more thorough testing. I will change the PR status ( and branch head, if the MOM6 team would prefer to have separate branch to test the new IO similar to  [fms-io-dev ](https://github.com/NOAA-GFDL/FMS/tree/fms-io-dev) in the FMS repo)  when I have finished regression testing. The PR for the alpha tag will also include thorough descriptions of the major changes.


marshallward:
@wrongkindofdoctor can you update the FMS hash of `.testing/Makefile` to point to the version of FMS that you plan to test this against?  A branch name may also work, though I haven't tested it.

marshallward:
Also is the really >5000 commits!!?  Might need to do some squashes here...

marshallward:
Sorry, noticed that this is an alpha PR, disregard my last comment re:commits...

I also noticed that the test Makefile is not set up to read `fms_platform.h`, which I need to sort out.  So I'll try updating the FMS hash and fixing the include statement myself, and will update you tomorrow.

wrongkindofdoctor:
@MJHarrison-GFDL The new interface routine is called `open_file`, so mpp_open_file is now referenced explicitly by modules with non-netcdf reads/write calls.

Create_file and reopen_file have been removed and replaced with calls to open_file, register_axis, register_field, read_data or write_data, and close_file for netcdf file IO, so the axis_data_type is defined at the module level for simplicity.

src/initialization/MOM_grid_initialize.F90
```diff
@@ -255,16 +283,24 @@ subroutine set_grid_metrics_from_mosaic(G, param_file, US)
...
 
-  ! Read X from the supergrid
-  tmpZ(:,:) = 999.
-  call MOM_read_data(filename, 'x', tmpZ, SGdom, position=CORNER)
+  ! open the file for domain-decomposed read
+  file_open_success = open_file(fileObjRead, filename, "read", SGdom%mpp_domain, is_restart=.false.)
+  ! tmpZ is defined on the data domain
+  tmpZ(:,:) = 999.0
+  ! register the global axes
+  call register_axis(fileObjRead, 'nxp','x', domain_position=EAST_FACE)
```
adcroft: nxp should be something that can be inquired for. The names of dimensions should not be hard coded for a read.

Hallberg-NOAA:
Thank you, @wrongkindofdoctor, for this heads-up on what you envision for MOM6 changes related to FMS2.  Here are a few broad comments:

1. I think that adding more than 5300 commits (>1 commits per line changed!) related to this rather targeted task would make the MOM6 commit log-file pretty useless for backtracking to identify when, where, and why changes came in.  Would it be possible to consolidate these to many commits into a few commits that actually track the major changes? Also have you verified that the code with essentially every commit that appears in the MOM6 code log compiles correctly and passes all of our regression testing?

2. These changes to FMS should not require any changes to the MOM6 code outside of the src/framework directory without very strong justification.  The intention in the design of MOM6 is that we would write wrappers that can map onto  _any_ infrastructure package (e.g., fms, ESMF, fms2, etc.) by changing only the contents of this one directory. As a corollary, the subroutine names that are exposed to MOM6 code outside of the MOM6 src/framework directory should not be changed to mpp_ something.  Where there is a simple pass-through of a subroutine to an FMS routine with a different name, this should be done via renaming in the routine within the MOM6 src/framework directory that changes it.

3. If there are going to be infrastructure interface changes that are required to use FMS2, and in particular ones that will be manifest widely in the MOM6 code, this should be done by interface changes that are tested and work with the current version of FMS, and it should occur only after careful and thorough discussions with the MOM6 development and design team, and adoption of these interface changes to the MOM6 dev/master branch. Doing so will help avoid breaking existing MOM6 applications on all of its various branches. In the discussions of these changes, I expect there will be an intrinsic bias toward retaining the existing MOM6 infrastructure interfaces unless there is a consensus that there is a compelling reason to change.

  I think that at this point, some detailed in-person discussions of these proposed changes to the MOM6 src/framework interfaces and src/framework code in preparation for using FMS2 would be in order.

src/initialization/MOM_grid_initialize.F90
wrongkindofdoctor: This potential solution queries the file for the global dimension names, loops through each one, and registers the dimensions that match the sizes of temporary buffers that will hold each array. 

```
ndims = get_num_dimensions(fileObjRead)
  allocate(dimNames(ndims))
  call get_dimension_names(fileObjRead, dimNames)
  ! register the global axes
  do i=1,ndims

    call get_dimension_size(fileObjRead, dimNames(i), globalDimSize)

    if (globalDimSize .eq. size(tmpz,1)) then
      call register_axis(fileObjRead, trim(dimNames(i)),'x', domain_position=CENTER)
    elseif (globalDimSize .eq. size(tmpz,2)) then
      call register_axis(fileObjRead, trim(dimNames(i)),'y', domain_position=CENTER)
    elseif (globalDimSize .eq. size(tmpV,1)) then
      call register_axis(fileObjRead, trim(dimNames(i)),'x', domain_position=EAST_FACE)
    elseif (globalDimSize .eq. size(tmpU,2)) then
      call register_axis(fileObjRead, trim(dimNames(i)),'y', domain_position=NORTH_FACE)
    endif
  enddo

  deallocate(dimNames)
```

wrongkindofdoctor:
@MJHarrison-GFDL @adcroft @marshallward 
All Travis CI tests have passed, meaning that MOM6 w/new FMS IO is now FINALLY ready for review!
As requested, I have encapsulated the new read and write routines in the framework modules, mainly MOM_io.F90 and MOM_restart.F90. The sole exception is that I replaced a call to `save_restart` in MOM.F90, with a [call](https://github.com/wrongkindofdoctor/MOM6/blob/47e7f3c4f618e31ce3e636ec07d6ffb41dc6de7d/src/core/MOM.F90#L2519) to a new [`write_initial_conditions` routine](https://github.com/wrongkindofdoctor/MOM6/blob/47e7f3c4f618e31ce3e636ec07d6ffb41dc6de7d/src/framework/MOM_restart.F90#L1130) in MOM_restart. The routines are structurally similar, but `write_initial_conditions` routines calls the non-restart FMS IO `register_field` and `write_data` calls, while save_restart calls the restart file-specific `register_field` and `save_restart` routines.

In addition to the Travis CI tests, I the ran the MOM6 regression and restart testing suite with the new IO on Gaea with Gnu7 and Intel19. Currently, the Makefile points to the FMS master branch, but I will update it to point to 2020.01 when we issue the release.

marshallward:
Hi Jess, thanks for going over this with us today.  I think there are a lot of positive changs here which will gives us more flexibility when working with files.  I've put the new build into the Gaea regression test suite, but we currently cannot build SIS2 due to the changes in API.  The first point of fail is with `open_file`, detailed below.

SIS2 inherits its IO subroutines from MOM, e.g. [SIS_dyn_cgrid.F90, L27](https://github.com/NOAA-GFDL/SIS2/blob/024255855a4dce8e86f21c1c4e2fcc5c9ca3bb95/src/SIS_dyn_cgrid.F90#L27)
```
use MOM_io,           only : open_file, APPEND_FILE, ASCII_FILE, MULTIPLE, SINGLE_FILE
```
But MOM6 has been updated to the new API and SIS2 has not, e.g. [L1799](https://github.com/NOAA-GFDL/SIS2/blob/024255855a4dce8e86f21c1c4e2fcc5c9ca3bb95/src/SIS_dyn_cgrid.F90#L1798-L1799):
```
      call open_file(CS%v_file, trim(CS%v_trunc_file), action=APPEND_FILE, &
                     form=ASCII_FILE, threading=MULTIPLE, fileset=SINGLE_FILE)
```
so the interfaces do not match and the compiler fails.

Unfortunately SIS2 will also need to be updated before we can move forward with this patch.

wrongkindofdoctor:
@marshallward That's fine; I already planned on updating SIS2 once the new MOM6 IO passed the Travis tests and another review. Tagging @nikizadehgfdl to confirm that I will begin working on implementing the new FMS IO in SIS2, with a target completion date by 02/28/20.

src/tracer/tracer_example.F90
```diff
@@ -181,11 +182,12 @@ subroutine USER_initialize_tracer(restart, day, G, GV, h, diag, OBC, CS, &
...
     if (len_trim(CS%tracer_IC_file) >= 1) then
 !  Read the tracer concentrations from a netcdf file.
-      if (.not.file_exists(CS%tracer_IC_file, G%Domain)) &
-        call MOM_error(FATAL, "USER_initialize_tracer: Unable to open "// &
+      if (.not.file_exists(CS%tracer_IC_file)) &
+        call MOM_error(FATAL, "USER_initialize_tracer: Unable to find "// &
                         CS%tracer_IC_file)
       do m=1,NTR
         call query_vardesc(CS%tr_desc(m), name, caller="USER_initialize_tracer")
+        if (m == 1) &
```
marshallward: Is this a bug?  It looks like only the first tracer is being read.

marshallward:
Given the size of this PR, I think it would be helpful to minimize as many of the API changes as we can which do not change functionality.  Could I ask that you revert the following?

* Undo the renaming of `timelevel` to `timelevel`
* Remove any whitespace changes
* Retain the `open_file` function name where possible.  I understand that the namespace conflict will need to be managed, so this may require something like this in `MOM_io`:
```
use mpp_io_mod, only: open_file => mpp_open
use fms2_io_mod, only: open_ncfile => open_file
```
(I made up `open_ncfile`, but any name will do)

I think these three changes alone would greatly reduce the number of files in the diff, and would allow for a more targeted discussion of the changes in `MOM_io.F90` and `MOM_restart.F90`.

Substantive changes which remove unused arguments, such as `file_exists` are probably OK, however.

These changes can be re-introduced later if there's a need to conform to an overall style.

src/tracer/tracer_example.F90
wrongkindofdoctor: Looks like it. I've removed it.

marshallward:
Hi Jess, given that there are still lots of commits coming in to this PR, and since it will also require a squash at some point, can I request that we close this and re-open once it's ready?

wrongkindofdoctor:
@marshallward Sorry about that! I had to resolve issues incorporating the new thread affinity module in the latest FMS updates. but this should be it. If all of the Travis CI tests pass, you, @adcroft, and @MJHarrison-GFDL can proceed with the review of the code when you are able. In the mean time, I will start updating SIS2.

wrongkindofdoctor:
Closed while I work through updates for internal gitlab CI tests.

PR #1041, opened 2019-12-06
+Simplified handling of time with forcing type

Hallberg-NOAA:
Refactored how time-averaging of fluxes in forcing types that span multiple
timesteps and flux diagnostics are handled, and rescaled the units of
fluxes%dt_buoy_accum from [s] to [T].  This involved changing the arguments to
fluxes_accumulate, forcing_accumulate, mech_forcing_diags and
forcing_diagnostics, but because of the differing types of the arguments, an
incompatible mix of code will not compile.  Also changed the units of dt as
passed to accumulate_net_input, and made a minor change to extractFluxes1d to
avoid the possibilty of a division by zero.  All answers are bitwise identical,
but there are public interface changes, including changes that impact the mct
and nuopc driver codes.


Hallberg-NOAA:
This PR is being tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/9475

PR #1042, opened 2019-12-06
+Added the new parameter KAPPA_SHEAR_ITER_BUG

Hallberg-NOAA:
Added the new runtime parameters KAPPA_SHEAR_ITER_BUG and KD_TRUNC_KAPPA_SHEAR
to permit correction of a dimensionally inconsistent expression in the Newton's
method solver code of kappa_shear, and to allow the value of shear mixing that
is neglected compared with the background mixing to be set at run-time instead
of being hard-coded.  By default, all answers are bitwise identical, but there
are two new runtime parameters and the MOM_parameter_doc files change.


Hallberg-NOAA:
This PR is being tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/commits/pr/1042 .

PR #1052, opened 2020-01-18
(*)Fixed geothermal with vanished layers

Hallberg-NOAA:
Corrected the behavior of geothermal with vanishing layers.  With this change,
all of the MOM6-examples test cases work with DEFAULT_2018_ANSWERS=False and
ANGSTROM=0.0.  All answers in the existing MOM6-examples test cases are bitwise
identical.


Hallberg-NOAA:
This PR was tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/9664, which passed.

PR #1058, opened 2020-02-05
Obc segment flag index

marshallward:
In `setup_[uv]_point_obc`, a number of flags for each segment were assigned without index (e.g. `OBC%segment%u_values_needed = .true.`), which was causing all segments to be assigned the same value.

This could problems with, say, `[uv]_values_needed` flags, which have different values depending on if they are E-W or N-S oriented.

This patch adds the index to the assignments.

No answer changes were seen in my Gaea tests, so this may not have any repercussions.


marshallward:
Gaea regression test: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/9781

PR #1070, opened 2020-03-06
Diagnostic remapping of vertically extensive fields

ashao:
For review by @adcroft @StephenGriffies @gmacgilchrist. @gmacgilchrist, when you get a chance can you see whether this improves the closure of the budgets?

Vertically extensive diagnostics like advective transports, diffusive fluxes, etc. were being remapped onto a diagnostic grid prior to the physical process that chagne the tracer and/or layer thicknesses.
However this means that the effective operator used for each component of a tracer budget was not the same, and so budgets could never be closed cell-by-cell in a diagnostic coordinate.

This commit fixes one part of the inconsistency by setting the target diagnostic grid for all vertically extensive quantities to be one constructed at the very beginning of the timestep.

However, problems with closing the budget should still be expected
1. Conservation of column integrals cannot be expected because the target grid can be smaller than the source grid (layer thicknesses at the native grid at the current point in the algorithm). This
   leads to some fluxes being 'thrown away' due to the reintegrate algorithm.
2. To truly have a consistent operator for all terms of a budget, the source grid for all vertically extensive grids should also be the same


StephenGriffies:
Thanks @ashao for this code.  But I observe that it does not compile so @gmacgilchrist cannot test it. 

I am puzzled by your comments. 

1/ Conservation of column integrals cannot be expected because the target grid can be smaller than the source grid (layer thicknesses at the native grid at the current point in the algorithm). This leads to some fluxes being 'thrown away' due to the reintegrate algorithm.

-->I am not familiar with the "reintegrate algorithm".  But why is this an issue with the diagnostics and not an issue with the prognostic model?  We always have changing thicknesses in columns. And why would there be only an issue if the target grid is smaller than the source grid, and not a problem with the target grid is larger than the source?  I realize we discussed this issue before, but I do not grok the basics.  

It seems you are saying we cannot conserve on the diagnostic grid.  Correct?  Where then does the following grid point equation fail?

If A = B + C  then Phi(A) = Phi(B) + Phi(C) so long as Phi is a linear operator. 

In this equation, A = "left hand side" total tendency, and B and C are the suite of "right hand side" processes such as diffusion, remapping, boundary fluxes, horizontal advection, etc.  So long as Phi, the linear remapping operator, is identically applied to all tendency terms, then this equation holds to machine precision.  The problem with the original code was that Phi was different when applied to A, B, C... and so there was no way to satisfy the above equality.  So long as we can make Phi the same, then we should be home free.  

2/ To truly have a consistent operator for all terms of a budget, the source grid for all vertically extensive grids should also be the same.

-->Are you using the term "consistent" in the formal numerical analysis sense, whereby the discrete equations reduce to the continuous equations in the continuum limit?  If so, then I do not understand how we are inconsistent.  If you instead mean "inconsistent" in some other looser sense, then should be be worried?  At this point my number one concern is conservation of the discrete remapped budgets, which is point 1/ above.

ashao:
Ah ok, I think I see where I went wrong with trying to implement this. I'll look at this again closer this afternoon and update. 

Regardless let me try to explain a little better where the problem is.

To your first point:
Define an function `Phi(A, h_source, h_dest)` such that `Phi` maps `A` (tendencies in a column) from the model grid `h_source` to a diagnostic grid `h_dest`. After our discussion last week, we had concluded that the `h_dest` in Phi has to be the same for all vertically extensive quantities. What I'm highlighting now is that we need to make sure that `h_source` is always the same as well. Currently, we're always using `h_source` from right before the process occurred. `h_source` potentially changes from process to process and so the operator is clearly not the same. (Note: in my original comments, I was referring to 'consistent' as meaning that `h_source` and `h_dest` were the same for all vertically extensive quantities)

With regard to conserving the vertical integrals, the 'reintegrate' routine (used to do remapping for vertically extensive quantities) does not require that the `SUM(h_source) == SUM(h_dest)`. If `h_dest` is smaller than `h_source`, then the vertical integral of `A` after 'reintegrating' is actually `SUM(A)*[SUM(h_dest)/SUM(h_source)].` As an extreme demonstration, consider some process that causes a column of zero thickness to inflate by some quantity `Delta h`. The thickness tendency associated with this is then `(Delta H)/(Delta t)`. However, if we remap this thickness tendency to the grid before the process occurred (i.e. `h_dest = 0` we get `(Delta H)/(Delta t)*(0/(Delta H))=0`. I vaguely remember @StephenGriffies  and @adcroft explaining why it's logical to post on the grid before the process occurred, but I don't recall the reasoning anymore. Could one of you rehash it? That being said, I remember showing this empirically because when we posted on the grid after the process, the remapped vertical integrals were different than the native coordinate column integrals.

StephenGriffies:
Thanks @ashao for the points.  

1/ To diagnose the total tendency (left hand side), we have at the end of a time step the tracer content field (vertical extensive) 

T_cont_final = T_final * h_final, 

from which we diagnose the total time tendency 

A = (T_cont_final - T_cont_init)/(delta t) = (T_final * h_final - T_init * h_init)/(delta t).  

To perform this diagnostic requires T_init * h_init at the end of the time step, which may require some added memory. Do we agree on this point?  Presumably this is how we are currently diagnosing the "left hand side".  Correct?  

2/ As you note, the individual tendency terms on the right hand side of A = B + C + D + ... (advection, diffusion, remapping), do use the same h_source.  However, since h_source is part of the tendency itself, we do not need to worry about that detail.  That is, A = B + C + D + ... is an equation for extensive fields.  We never need to unpack the individual thicknesses used to compute the operators B, C, D, etc.  Instead, the proposal is to set h_source = h_dest = h_orig so we compute 

Phi(B, h_orig, h_orig), Phi(C, h_orig, h_orig), Phi(D, h_orig, h_orig), etc.  

3/ For your point about reintegrate algorithm, how do we conserve over a column with the present method (the one we are trying to replace)?

ashao:
1. Yes that's how we're diagnosing them and I fully agree that it i the correct way of diagnosing the tendency. The question for this specifically is which grid should those tendencies be associated with? Even in native space if one was to do a cumulative integral (starting from the bottom) should they be defined on the interfaces at the beginning or end of the timestep? Currently we're doing them at the beginning of the timestep, though as demonstrated in my example this seems like it would give the wrong answer.

2. I think I see where the confusion might be happening. The reintegration does need to know the layer thicknesses defining the source grid in order to do the remapping. So to fill this out with even more clunky, notation, I'm pointing out that this is the form that the operator needs to have for all quanties, Phi(A, h_begin_native, h_begin_diag).

3. We conserve currently because source and destination grids have the same total column thicknesses (to within floating point roundoff).

ashao:
This should be ready to be merged in. However, the Travis testing suite is reporting the build as broken due to the diagnostic checksums changing (as they should for all extensive variables). @marshallward, is there a procedure for updating those and/or requesting an exception to the failing CI?

marshallward:
There's no formal process for dealing with regressions right now.  If it's different and everyone agrees that it should be different, then we just overlook it and merge it.

However, you do have trailing whitspace in the diag mediator:
```
$ ./.testing/trailer.py -e TEOS10 -l 120 src config_src
src/framework/MOM_diag_mediator.F90, line 336: Trailing space detected "  real, dimension(:,:,:), allocatable :: h_begin !< Layer thicknesses at the beginning of the timestep used "
The command "./.testing/trailer.py -e TEOS10 -l 120 src config_src" exited with 1.
```

That is a very serious problem!!

ashao:
> That is a very serious problem!!
Agreed, pushing the big red button (mapped to F5 in my vim) to blow it away.

Issue #769, opened 2018-05-10
Remove missing_value MOM6 code


adcroft:
A grep for `missing_value` returns 161 occurrences in MOM6 code. This is partly because the wrapper for `register_diag_field()` has `missing_value` as an optional argument, which we propagated upwards, but it is mainly because `missing_value` was briefly being used in code instead of using masks.

Apparently we have inconsistent values appearing in history files, presumably related to the specification of `missing_value`.

- To avoid future confusion, and inconsistent history files, we need to remove all reference to `missing_value` in model and framework code. Only the `diag_manager` should ever know anything about `missing_value` (technically `_FillValue`) which is used solely for filling space in the file where there is no data.

jkrasting:
Are there other parts of the FRE workflow / post-processing that depend on
missing_value being present?  I know that had once been true.

-----
Dr. John Krasting, Physical Scientist
Biogeochemistry, Ecosystems, and Climate Group
NOAA/Geophysical Fluid Dynamics Laboratory
201 Forrestal Road, Princeton, NJ 08540
P. (609) 452-5359   |   F. (609) 987-5063

On Thu, May 10, 2018 at 11:26 AM, Alistair Adcroft <notifications@github.com

underwoo:
I do think the post-processing itself used the missing_value of
_FillValue.  Some of the tools used in post-processing (e.g. the nco tools)
use their values.  As long as they are both set in the variable's
attribute, there should be no issue in the post-processing that will cause
an issue.

--
Seth Underwood

S/W Development Process Lead
Modeling Systems Group
GFDL/NOAA/DOC
201 Forrestal Road
Princeton, NJ 08540-6649

(609) 452-5847 Office
(304) 376-9002 Cell
(609) 987-5063 Fax
Seth.Underwood@noaa.gov

PR #1091, opened 2020-04-16
Removes missing_value argument from fns in MOM_diag_remap

adcroft:
- This is a very small step towards #769. When I looked at what it will take to deal with #769 it turns out diag_mediator needs a significant refactor that just isn't going to happen overnight.
- missing_value was an argument to several functions in MOM_diag_remap but was not used. At one point these functions incorrectly used missing_value to mask data but this is not the correct usage of missing_value nor the right way to do masking in MOM6. This was fixed 3 years ago and this commit simply removes unused arguments.


marshallward:
There are conflicts in this PR now, can you take a look when you get a chance?

marshallward:
I resolved the commit in the GitHub editor, but obviously it has now not been tested locally.  Would be good for you to review it if you get a chance.

marshallward:
Gaea regression testing: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/10237

PR #1097, opened 2020-04-28
ROTATE_INDEX, INDEX_TURNS moved to debug params

marshallward:
The index rotation flags were incorrectly handled as model parameters,
rather than debug parameters.  This patch fixes this oversight.


marshallward:
Gaea regression test: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/10294

Regression has passed (excluding parameter verification)

PR #1110, opened 2020-05-14
Reduced output during testing

marshallward:
The Makefile has been modified to reduce the amount of output during testing.  Output is generally omitted on a successful test.  When a test fails, we only display a small portion of the total output.

We also now run all tests, even if they fail, in order to give a complete profile of the test failures.

Regression testing rules have been integrated into the general rules.

Finally, Travis config has been modified to further reduce output, and to run all of tests (make -k).


marshallward:
This seems good enough to merge from my end, but there is no hurry and plenty of room for feedback.

marshallward:
Sample output from a run with answer changes:

https://travis-ci.org/github/marshallward/MOM6/jobs/687177632

PR #1113, opened 2020-05-19
Fixes an integer-kind mismatch in MOM_random, seed_from_time()

adcroft:
- gcc/8.3.0 issued `Error: Integer too big for its kind` reported in
  feedback on PR #1111. The intent was to assume kind=4 in these
  calculations but apparently our compilers were promoting
  `mod(dy + 32*(mo + 13*yr), 2147483648)` to kind=8. There were two
  mistakes in the expression:
  - the use of `2147483648` in the `mod` is not representable with kind=4;
  - the `mod` produces negative values and should have been a `modulo`.
- This commit reduces the range of the results by one number on the
  positive side and removes all the negatives.

@gustavo-marques Could you test this patch for me? It passes everything at our end but so did the original PR.

### DIFF
--- a/src/framework/MOM_random.F90
+++ b/src/framework/MOM_random.F90
@@ -143,8 +143,9 @@ integer function seed_from_time(Time)
   call get_date(Time,yr,mo,dy,hr,mn,sc)
   s1 = sc + 61*(mn + 61*hr) + 379 ! Range 379 .. 89620
   ! Fun fact: 2147483647 is the eighth Mersenne prime.
-  ! This is not the reason for using 2147483647+1 here.
-  s2 = mod(dy + 32*(mo + 13*yr), 2147483648) ! Range 0 .. 2147483647
+  ! This is not the reason for using 2147483647 here. It is the
+  ! largest integer of kind=4.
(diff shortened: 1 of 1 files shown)


marshallward:
I know this isn't the only place where this is being violated, but we really ought to stop assuming that `kind` is equal to the number of bytes.  There is at least one major compiler (NAG) which does not make this assumption.

I'd recommend using `iso_fortran_env, only : int4`.

gustavo-marques:
I can confirm it builds with gnu/8.3.0.

klindsay28:
@marshallward , it looks like `int4` was introduced in Fortran 2008.

If you want to support earlier Fortran versions, you can use `i4 = selected_int_kind(6)`. This is what CIME and CESM components use. `selected_int_kind` was introduced in Fortran 90.

Does the MOM6 development team have a position on minimal Fortran version required for model development? I don't see this at https://mom6.readthedocs.io/en/dev-gfdl/index.html or https://github.com/NOAA-GFDL/MOM6/wiki.

marshallward:
@klindsay28 I recall last we discussed this, we found FMS had already moved onto 2008, and so we've opted to do the same.  (We're already using `iso_fortran_env` types elsewhere in the code).

I agree we need to clarify our target standard somewhere in the docs.

PR #1135, opened 2020-06-16
+Improve documentation in MOM_parameter_doc files

Hallberg-NOAA:
This PR consists of a set of changes to improve the logging of modules to
the various MOM_parameter_doc files, including moving the halo size parameters
from the doc.debugging to the doc.all files, and avoiding the logging of
parameters that are specific to unused modules, although they are still being
read to avoid triggering issues if FATAL_UNUSED_PARAMS is true.  The commits
that are including in this PR include:

- NOAA-GFDL/MOM6@f5dfef8 +Do not log parameters from unused modules
- NOAA-GFDL/MOM6@85b784d +Do not log modules to parameter_doc.short files
- NOAA-GFDL/MOM6@cf80f4b +Log modules to parameter_doc.debug files
- NOAA-GFDL/MOM6@5a807f6 +Record halo sizes in MOM_parameter_doc.all files
- NOAA-GFDL/MOM6@b4d3c15 +Add 4 optional args to log_version & doc_module

### DIFF
--- a/config_src/coupled_driver/MOM_surface_forcing_gfdl.F90
+++ b/config_src/coupled_driver/MOM_surface_forcing_gfdl.F90
@@ -1272,7 +1272,7 @@ subroutine surface_forcing_init(Time, G, US, param_file, diag, CS)
 
   call write_version_number(version)
   ! Read all relevant parameters and write them to the model log.
-  call log_version(param_file, mdl, version, "")
+  call log_version(param_file, mdl, version, "", log_to_all=.true., debugging=.true.)
 
   call get_param(param_file, mdl, "INPUTDIR", CS%inputdir, &
                  "The directory in which all input files are found.", &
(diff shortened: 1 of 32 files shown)


Hallberg-NOAA:
The apparent Travis CI failure here is because a missing diagnostic was added. I do not believe that this should be considered to have actually failed.

adcroft:
Yes it passed. We have no "warning" indicator like we do on gitlab.

adcroft:
Q: Do we still need the MINIMAL_DOCUMENTATON and COMPLETE_DOCUMENTATION parameters?

Hallberg-NOAA:
A: Yes, I think that the MINIMAL_DOCUMENTATION and COMPLETE_DOCUMENTATION parameters are useful for giving users control over what is written.

Issue #1146, opened 2020-06-26
Floating point exception in OM4-025


jiandewang:
I am having Floating point exception when use MOM6-examples (dev/master eb2f3b4 Merge branch 'dev-master-candidate-2020-05-15' into dev/master) for OM4-025. Here are some details:
(1) two changes in MOM_input: dt-thermal=1800s, VERTEX_SHEAR=T  (they are 7200s and F in original setting)
(2) this happens only in "DEBUG" mode, "REPRO" mode works fine.
(3) intel 18 compiler
(4) the crash happened in the beginning of run

Below is error message:
[Orion-03-64:400365:0:400365] Caught signal 8 (Floating point exception: floating-point invalid operation)
[Orion-03-61:246749:0:246749] Caught signal 8 (Floating point exception: floating-point invalid operation)
*** Error in `/./MOM6': munmap_chunk(): invalid pointer: 0x000000001d426ab0 ***
*** Error in `./MOM6': munmap_chunk(): invalid pointer: 0x000000001ded96b0 ***
======= Backtrace: =========
/lib64/libc.so.6(+0x7f754)[0x7f12e68a9754]
======= Backtrace: =========
/usr/lib64/libucs.so.0(+0x591a0)[0x7f11e56fe1a0]
..................................
......................................
MOM6               00000000058176DB  Unknown               Unknown  Unknown
libpthread-2.17.s  00007F91427925D0  Unknown               Unknown  Unknown
MOM6               0000000001B31199  mom_diag_remap_mp         343  MOM_diag_remap.F90
MOM6               000000000116DCE7  mom_diag_mediator        3267  MOM_diag_mediator.F90
MOM6               0000000003CF4D62  mom_mp_step_mom_t        1174  MOM.F90
MOM6               0000000003CE4FF3  mom_mp_step_mom_          774  MOM.F90
MOM6               0000000002B3B355  ocean_model_mod_m         579  ocean_model_MOM.F90
MOM6               0000000000DF91E7  MAIN__                   1082  coupler_main.F90
MOM6               0000000000410E22  Unknown               Unknown  Unknown
libc-2.17.so       00007F91423D8495  __libc_start_main     Unknown  Unknown
MOM6               0000000000410D29  Unknown               Unknown  Unknown
forrtl: error (78): process killed (SIGTERM)
Image              PC                Routine            Line        Source
MOM6               00000000058176DB  Unknown               Unknown  Unknown
[...]

PR #1147, opened 2020-06-26
(*)Corrected halo size in EOS call if VERTEX_SHEAR=T

Hallberg-NOAA:
Corrected halo size in density derivative calculations in smoothed_dRdT_dRdS
This fixes an i-parallelization problem that was recently introduced (as a part
of MOM6 PR #1089) when VERTEX_SHEAR is True, and closes MOM6 issue #1146.  All
answers in the existing MOM6-examples test suite are bitwise identical, but this
does change (correct) answers when VERTEX_SHEAR is true.


Hallberg-NOAA:
This PR is being tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/pipelines/10704 .

PR #1156, opened 2020-07-08
Adds parameterizations of SGS variance to PGF and isopycnal slopes

adcroft:
This adds deterministic parameterizations of both i) the Brankart terms in the pressure gradient force (PGF) and ii) related effect on isopycnal slope in the GM parameterization. There is more to come but these changes add up such that maintaining compatibility with dev/gfdl is becoming a burden. The deterministic schemes in this PR are being evaluated in OM4.1. There are quite a few steps including:
- major re-factor of the line integrals and PGF code
  - variants of the line integrals are now grouped into MOM_density_integrals where there were scattered across PGF and EOS modules before. This happens to localize the logic of where stuff is switched but the primary reason is for being able to access data type that should not be seen by the EOS.
  - the routines always and still, worked on a single layer at a time but now take 3D inputs along with "k". This allows reconstruction to be moved into this routine and avoid 3D pre-calculation (still in progress with a new monotonized PLM scheme). For now the main advantage is to avoid the addition of numerous arguments associated with each of the SGS fields.
- adds the PPM form of Boussinesq PGF using quadrature. This was once used in 2d for published results but never extended to three dimensions. PPM has been validated by zeroing out the curvature term to bitwise reproduce the PLM results. New tests that use PPM-PGF will be coming later in MOM6-examples.
- a less major re-factor of some of the ALE reconstruction code because they are used in PGF (includes renaming ppoly_E -> edge_values as discussed a while ago)
- introduction of new equation of state functions that use SGS variance and covariance as inputs
- added new fields to the `tv` data type. This are currently determined by parameterizations but will later be optionally prognostic fields.
- updated TC2 to enable (test) both these new schemes

Interactions:
- This adds new parameters to MOM6-examples will need updating once it points to this code.
- There are no answer changes for existing configurations in MOM6-examples but tc2 and tc2.a were updated to test these schemes so a regression test will fail.


src/ALE/PLM_functions.F90
```diff
@@ -5,21 +5,177 @@ module PLM_functions
...
+    ! larger extreme values.
+    PLM_slope_wa = 0.0
+  endif
+
+  ! This block tests to see if roundoff causes edge values to be out of bounds
+  if (u_c - 0.5*abs(PLM_slope_wa) < u_min .or.  u_c + 0.5*abs(PLM_slope_wa) > u_max) then
+    PLM_slope_wa = PLM_slope_wa * ( 1. - epsilon(PLM_slope_wa) )
+  endif
+
+  ! An attempt to avoid inconsistency when the values become unrepresentable.
```
Hallberg-NOAA, line 57: Please modify this comment and the one at line 107 to note that these are dimensionally inconsistent expressions. It would be better to rewrite the code to handle underflow consistently, (e.g., via a du_underflow argument), but this can be added later. For now leaving a note should suffice.

src/equation_of_state/MOM_EOS.F90
```diff
@@ -193,6 +202,43 @@ subroutine calculate_density_scalar(T, S, pressure, rho, EOS, rho_ref, scale)
 
 end subroutine calculate_density_scalar
 
+!> Calls the appropriate subroutine to calculate density of sea water for scalar inputs
```
Hallberg-NOAA, line 205: I think that perhaps calculate_stanley_density_scalar should follow the same pattern with respect to the dimensions of its arguments and rescaling via the contents of EOS as does calculate_density_scalar, and not the pattern of calculate_density_array, as it currently does.  This is especially true if the calculate_density_array routines are eventually going to be hidden from public exposure.

Hallberg-NOAA:
I have visually examined all of the code change in this PR, and I am supportive of it. There are two minor points where I had previously commented on specific parts of the the commit. Specifically there are two places in the ALE/PLM_functions code where we need comments noting dimensionally inconsistent expressions to deal with underflow, and I have a query about the handling of the units in the corresponding calculate_stanley_density_scalar and calculate_density_scalar routines.

  I have also manually run this through the MOM6-examples test suite, and I have verified that this PR does not change answers there.  The deliberate fail on the tc2 test case has been noted.

  Once these minor points have been resolved, I will move ahead with manually merging this pull request, along with appropriate changes to the MOM_parameter_doc files.

src/core/MOM_PressureForce.F90
```diff
@@ -28,10 +28,10 @@ module MOM_PressureForce
 type, public :: PressureForce_CS ; private
   logical :: Analytic_FV_PGF !< If true, use the analytic finite volume form
                              !! (Adcroft et al., Ocean Mod. 2008) of the PGF.
-  logical :: blocked_AFV     !< If true, used the blocked version of the ANALYTIC_FV_PGF
+  logical :: blocked_FV      !< If true, used the blocked version of the ANALYTIC_FV_PGF
```
Hallberg-NOAA, line 31: blocked_FV is left over from an obsolete option and it not used anywhere in this module, so rather than being renamed, I think that it should simply be deleted.

src/core/MOM_PressureForce_FV.F90
```diff
@@ -186,7 +159,7 @@ subroutine PressureForce_AFV_nonBouss(h, tv, PFu, PFv, G, GV, US, CS, ALE_CSp, p
   EOSdom(1) = Isq - (G%isd-1) ;  EOSdom(2) = G%iec+1 - (G%isd-1)
 
   if (.not.associated(CS)) call MOM_error(FATAL, &
-       "MOM_PressureForce_AFV_nonBouss: Module must be initialized before it is used.")
+       "MOM_PressureForce_FV_nonBouss: Module must be initialized before it is used.")
```
Hallberg-NOAA, line 163: Because the Stanley option is not yet implemented for a non-Boussinesq code, it might be advisable to add an error message here if CS%Stanley_T2_det_coeff>=0.  If there are plans to add this in the very near-future this might not be necessary, but it would serve both as a reminder of a missing option and to avoid users thinking that they are using something that they are not.

src/ALE/PLM_functions.F90
```diff
@@ -5,21 +5,177 @@ module PLM_functions
...
+real elemental pure function PLM_slope_wa(h_l, h_c, h_r, h_neglect, u_l, u_c, u_r)
+  real, intent(in) :: h_l !< Thickness of left cell [units of grid thickness]
+  real, intent(in) :: h_c !< Thickness of center cell [units of grid thickness]
+  real, intent(in) :: h_r !< Thickness of right cell [units of grid thickness]
+  real, intent(in) :: h_neglect !< A negligible thickness [units of grid thickness]
+  real, intent(in) :: u_l !< Value of left cell [units of u]
+  real, intent(in) :: u_c !< Value of center cell [units of u]
+  real, intent(in) :: u_r !< Value of right cell [units of u]
+  ! Local variables
+  real :: sigma_l, sigma_c, sigma_r, u_min, u_max
```
Hallberg-NOAA, line 30: We should get into the habit of documenting the units of every real variable, and eventually enforce this for all code commits if possible.  In this case it would be easy as all have [units of u].

src/ALE/PLM_functions.F90
```diff
@@ -5,21 +5,177 @@ module PLM_functions
...
+real elemental pure function PLM_slope_cw(h_l, h_c, h_r, h_neglect, u_l, u_c, u_r)
+  real, intent(in) :: h_l !< Thickness of left cell [units of grid thickness]
+  real, intent(in) :: h_c !< Thickness of center cell [units of grid thickness]
+  real, intent(in) :: h_r !< Thickness of right cell [units of grid thickness]
+  real, intent(in) :: h_neglect !< A negligible thickness [units of grid thickness]
+  real, intent(in) :: u_l !< Value of left cell [units of u]
+  real, intent(in) :: u_c !< Value of center cell [units of u]
+  real, intent(in) :: u_r !< Value of right cell [units of u]
+  ! Local variables
+  real :: sigma_l, sigma_c, sigma_r, u_min, u_max, h_cn
```
Hallberg-NOAA, line 72: These local variables should have comments indicating their units, and variables with different units (i.e. h_cn) should be on a separate line if the variable units are being documented collectively.

src/ALE/PLM_functions.F90
```diff
@@ -5,21 +5,177 @@ module PLM_functions
...
+!> Returns a limited PLM slope following Colella and Woodward 1984.
+real elemental pure function PLM_monotonized_slope(u_l, u_c, u_r, s_l, s_c, s_r)
+  real, intent(in) :: u_l !< Value of left cell [units of u]
+  real, intent(in) :: u_c !< Value of center cell [units of u]
+  real, intent(in) :: u_r !< Value of right cell [units of u]
+  real, intent(in) :: s_l !< PLM slope of left cell [units of u]
+  real, intent(in) :: s_c !< PLM slope of center cell [units of u]
+  real, intent(in) :: s_r !< PLM slope of right cell [units of u]
+  ! Local variables
+  real :: e_r, e_l, edge, almost_two, slp
```
Hallberg-NOAA, line 120: Please document the units of these variables.

src/ALE/PLM_functions.F90
adcroft: Address in commit f0bab127117a700716f100ed11b776b2857220cf

src/equation_of_state/MOM_EOS.F90
adcroft: The problem actually goes deeper because second derivatives are always scaled so it is hard to build these Stanley version out of the existing functions. The existing `calculate_density_1d()` calls `calculate_density_array()` and to do so has to "unscale" rho_ref. This did not work for the second derivatives so I've had to write out longer versions of these functions. Addressed in commit e8adc48a36dfe4b12459651ee2893b6670c0e2a1.

src/core/MOM_PressureForce.F90
adcroft: Addressed in ec0946c9c9d7e0f50a78451be67bc80ab8bd91dd

src/core/MOM_PressureForce_FV.F90
adcroft: Addressed in 4307fa5e528778bd8668ee8caedfee2a0b5edfe6

src/ALE/PLM_functions.F90
adcroft: Addressed in 553166ae8e8a871a0d1e39a091cbe707ef8a3159

src/ALE/PLM_functions.F90
adcroft: Addressed in 553166ae8e8a871a0d1e39a091cbe707ef8a3159

src/ALE/PLM_functions.F90
adcroft: Addressed in 553166ae8e8a871a0d1e39a091cbe707ef8a3159

PR #1161, opened 2020-07-16
MOM6: +(*)Add option to use time-integrated barotropic continuity

Hallberg-NOAA:
MOM6: +(*)Add option to use time-integrated barotropic continuity

  This PR includes a set of changes that introduce the new option to use the
time integrated barotropic velocities to determine the time integrated
barotropic transports that update the sea surface height or bottom pressure
directly from their initial condition.  This should give better behavior in
certain cases with vigorous gravity waves impinging on topography, and it should
improve the consistency between split and unsplit solutions.  This new option is
enabled when the new runtime parameter INTEGRAL_BT_CONTINUITY=True, SPLIT=True
and USE_BT_CONT_TYPE=True.  All of the test cases with a split time stepping
scheme give altered but qualitatively similar answers when the new option is
enabled.

  In addition there is one change setting a maximum value to the viscous
coupling coefficients so that they will always be representable as 32-bit
floating point numbers in diagnostics; this could conceivably change answers in
some cases, but it does not in the MOM6-examples test suite and I think such
changes to be very unlikely in practice.  With this change, some configurations
and diag_tables that were failing before now work again as intended.

  By default all answers in all tested cases are bitwise identical, but there
are changes to the entries in the MOM_parameter_doc files.  The commits in this
PR include:

- NOAA-GFDL/MOM6@7a6a10a36 Merge branch 'dev/gfdl' into integral_bt_solver
- NOAA-GFDL/MOM6@37a7ee0da +Reuse find_uhbt with INTEGRAL_BT_CONT
- NOAA-GFDL/MOM6@4743f8cec +Implemented integral_BT_cont options for OBCs
- NOAA-GFDL/MOM6@6374aaf77 +Add find_duhbt_dubt_int
- NOAA-GFDL/MOM6@14d63487c +Add the new parameter INTEGRAL_BT_CONTINUITY
- NOAA-GFDL/MOM6@8df3e7f55 +Eliminate uhbt_IC and ubt_IC from restart files
- NOAA-GFDL/MOM6@b3a584d43 (*)Set maximum value of au_visc

### DIFF
--- a/src/core/MOM_CoriolisAdv.F90
+++ b/src/core/MOM_CoriolisAdv.F90
@@ -256,7 +256,7 @@ subroutine CorAdCalc(u, v, h, uh, vh, CAu, CAv, OBC, AD, G, GV, US, CS)
   enddo ; enddo
 
   !$OMP parallel do default(private) shared(u,v,h,uh,vh,CAu,CAv,G,CS,AD,Area_h,Area_q,&
-  !$OMP                        RV,PV,is,ie,js,je,Isq,Ieq,Jsq,Jeq,nz,h_neglect,h_tiny,OBC)
+  !$OMP                        RV,PV,is,ie,js,je,Isq,Ieq,Jsq,Jeq,nz,h_neglect,h_tiny,OBC,eps_vel)
   do k=1,nz
 
     ! Here the second order accurate layer potential vorticities, q,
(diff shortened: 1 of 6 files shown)


adcroft:
doc files were changed - needs handling on the command line

Issue #966, opened 2019-07-31
benchmark experiment issues


marshallward:
Two minor issues with the `benchmark`:

1. ~Vertical level `NK` must be at least 6, since there's an explicit `do itt=1,6` loop in `src/user/benchmark_initialization.F90`.~ This was incorrect, the issue is that `NK` must satisfy the constraint `NK > NKML + NKBL` when `BULKMIXEDLAYER = .true.`.  For the tc1 verification test, `NKML + NKBL + 1` was 5.  Using `NK = 5` does not raise the error seen at `NK = 4`. The reference to 6 iterations was just a coincidence.

2. Vertical level needs to be even higher (in my case 8, but probably depends on other details) due to an unbounded `do` loop which can search for points outside of bounds before satisfying its loop termination condition.

First can probably be fixed by requiring ~`NK >= 6`~ `NK > NKML + NKBL`.  The second may need something a bit more specific to the numerics, or at least a bounds check.

marshallward:
#1186 has resolved the issue seen in tc1 when `NK = 6`, but other issues remain.

For `NK = 4`, we see the following error:
```
At line 363 of file ../../../src/initialization/MOM_coord_initialization.F90
Fortran runtime error: Index '5' of dimension 1 of array 't0' above upper bound of 4

Error termination. Backtrace:
#0  0x5555564950b2 in set_coord_from_ts_range
	at ../../../src/initialization/MOM_coord_initialization.F90:363
#1  0x55555649902a in __mom_coord_initialization_MOD_mom_initialize_coord
	at ../../../src/initialization/MOM_coord_initialization.F90:96
#2  0x555555b1ee0c in __mom_MOD_initialize_mom
	at ../../../src/core/MOM.F90:2309
#3  0x55555579465c in mom_main
	at ../../../config_src/solo_driver/MOM_driver.F90:296
#4  0x555555796b42 in main
	at ../../../config_src/solo_driver/MOM_driver.F90:27
```

For `NK = 5` we see a floating point exception:
```
Thread 1 "MOM6" received signal SIGFPE, Arithmetic exception.
0x000055555629fcaf in benchmark_initialization::benchmark_initialize_thickness (h=..., g=..., gv=..., us=..., param_file=..., eqn_of_state=0x555558ad66f0, 
    p_ref=20000000, just_read_params=.FALSE.) at ../../../src/user/benchmark_initialization.F90:168
[...]

PR #1186, opened 2020-08-21
Fix epipycnal pairing array size declarations

Hallberg-NOAA:
Corrected the size of 4 arrays used to describe layer pairings in
tracer_epipycnal_ML_diff, to avoid the possibility of segmentation faults
when there are very few interior layers compared with the number of mixed and
buffer layers.  Also corrected a number of spelling errors in comments.  In
runs that previously worked, all answers should be bitwise identical, and they
are identical in all MOM6-examples test cases.

### DIFF
--- a/src/tracer/MOM_tracer_hor_diff.F90
+++ b/src/tracer/MOM_tracer_hor_diff.F90
@@ -36,7 +36,7 @@ implicit none ; private
 
 public tracer_hordiff, tracer_hor_diff_init, tracer_hor_diff_end
 
-!> The ocntrol structure for along-layer and epineutral tracer diffusion
+!> The control structure for along-layer and epineutral tracer diffusion
 type, public :: tracer_hor_diff_CS ; private
   real    :: KhTr           !< The along-isopycnal tracer diffusivity [L2 T-1 ~> m2 s-1].
   real    :: KhTr_Slope_Cff !< The non-dimensional coefficient in KhTr formula [nondim]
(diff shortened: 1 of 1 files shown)


marshallward:
This appears to resolve #966.

Regression testing on Gaea: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11074

marshallward:
My last comment was not quite correct.  This resolves the error with `NK = 6` in tc1 but not `NK = 4`.

marshallward:
If I am reading this right, then it seems that `num_srt` (which sets the upper bound on iteration) is itself bounded by `2*nkmb`, or `2*(nkml + nkbl)`.

https://github.com/NOAA-GFDL/MOM6/blob/f093750526283c909885c346ccbd8ce9ee524ebe/src/tracer/MOM_tracer_hor_diff.F90#L761-L778

In the case of tc1, 2*(nkml + nkbl) is 8, and I hit the out-of-bound index at 7, which is larger than NK but still within bounds.

These declarations also appear to work correctly:
```
 685   integer, dimension(2*GV%nk_rho_varies) :: &                                   
 686     kbs_Lp, &   ! The sorted indices of the Left and Right columns for          
 687     kbs_Rp      ! each pairing.                                                 
 688   logical, dimension(2*GV%nk_rho_varies) :: &                                   
 689     left_set, &  ! If true, the left or right point determines the density of   
 690     right_set    ! of the trio.  If densities are exactly equal, both are true. 
```
Would it be better to use this rather than `2*NK`?

----

Also, the `NK = 4` fail is already violating the requirement that `NK >= NKML + NKBL + 1` when `BULKMIXEDLAYER = .true.`, so that constraint needs to be checked somewhere.  But that can perhaps be addressed separately.

Hallberg-NOAA:
I am pretty sure the upper limit on num_srt is PEmax_krho, which in turn can be as large as GV%ke. With this, I think that the most restrictive limit on the number pairings in this code is `GV%ke+GV%nk_rho_varies`, but as this is less than `2*GV%ke` and I know for certain that there will be fewer pairings than this, and also because these are only 1-d arrays, it might be OK to leave this as `2*GV%ke`

marshallward:
Sorry, I misread the second loop, yes you are right.  Disregard my comment.

Issue #1185, opened 2020-08-20
Change Default Behavior of BBL_USE_EOS


breichl:
BBL_USE_EOS defaults as False.  In layer mode this is okay, since it uses the layer values to set BBL stratification.  In ALE mode this is problematic as it can result in random numbers being used to compute BBL stratification, which can lead to unreasonable viscosities in the model.

After discussions with @Hallberg-NOAA we are proposing to change the default to avoid potential pitfalls in the future.

Two possible fixes are:
1) Change the default to True.  This probably won't have negative effects in layer mode (besides excess computation).
2) Change the setting so it is always "True" in ALE mode, but takes the value of BBL_USE_EOS in layer mode.

Perhaps the second makes the most sense, as allowing this to be False in ALE mode does not seem like a feature worth maintaining.  However, this may result in answer changes.

Hallberg-NOAA:
Fortunately there are no ALE test cases in the MOM6-examples suite that use ALE (i.e., have USE_REGRIDDING = True) that do not also explicitly BBL_USE_EOS=True.  However, there are a number of layer-mode test cases where explicitly setting BBL_USE_EOS=True would change answers, including ocean_only/benchmark, ice_ocean_SIS2/Baltic, the 4 ice_ocean_SIS2/SIS2* cases, and all three cases under coupled_AM2_LM3_SIS2.

@breichl's second proposed action seems like it should be safe, in that it makes no sense to use the layer target values of Rlay to determine the bottom boundary layer properties in ALE mode. However in the interest of taking the safest (interim) course of action, a third choice is that we could set the default value of BBL_USE_EOS True in ALE mode, but leave it false (for now) in layer mode, but keep it around in all cases to enable the current solutions to be recovered. Ultimately, BBL_USE_EOS is a good candidate to become obsolete, because setting this to True will always do the right thing because there is already a test to handle the case when there is no equation of state in pure layer mode.

In the mean time, it might make sense to explicitly set BBL_USE_EOS=True in the layer mode test cases where this would change answers.

breichl:
Okay, I will work up a solution that should preserve answers (for now), while changing the default in ALE so it doesn't rely on being explicitly set at run-time.

Hallberg-NOAA:
This issue was addressed with PR#1190, and can now be closed. In addition, BBL_USE_EOS has been added to the list of candidates for default changes the next time we come to that.

PR #1190, opened 2020-08-25
Updating the default setting for BBL_USE_EOS to match USE_REGRIDDING

breichl:
I tested 1 ALE and 1 layer config for reproducing answers, but I did not sweep the repository for any cases that use the now prevented combination of USE_REGRIDDING = True and BBL_USE_EOS = False.  @Hallberg-NOAA suggests that there are no such cases, which we should confirm soon.

I decided to add "FATAL" error for using these in combination because (1) WARNINGS usually get buried in output text, and (2) I don't see a good reason to allow the model to proceed in this situation.  But we can revisit this.


breichl:
This addresses issue #1185

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11079

Issue #1157, opened 2020-07-09
Unrealistic temperatures when using restart file for  a run


hmkhatri:
I am running an idealized channel model in ALE Z* mode (on a lat-lon grid) forced with steady winds, surface temperature restoration, and sponge at the north. The model runs fine if I start a new simulation. However, if I continue the simulation from a restart file (does not matter at what simulation time), temperatures increase drastically (see the attached domain-mean temperature and surface temperature plots, I continued the run from 7300 days using a restart file).  The data in the restart file looks fine. I do not see any missing or nan values. I have tested with the latest version of MOM6 as well as the early Jan version of MOM6 but the same issue appears. I have run multiple tests with different configurations. Switching off forcing terms, sponge did not help either.
![snapshot](https://user-images.githubusercontent.com/17830036/87069092-9c777480-c1e4-11ea-8133-5c6c7f9f5981.png)
![T](https://user-images.githubusercontent.com/17830036/87069093-9d100b00-c1e4-11ea-9d78-7f82b8374388.png)

Any idea what could be going wrong?

Issue #850 seems to report a similar problem. I have set VERTEX_SHEAR=True.  Could this be causing the problem?

Hallberg-NOAA:
This problem seems to me to be unlikely to be due to VERTEX_SHEAR, but you could also check on this specific question by trying it out in a run with VERTEX_SHEAR=False and see whether the problem goes away. If it does not, we have well-established techniques for tracking down the source of non-reproducibility across restarts, which I will describe briefly.

  The first thing that you will want to do to debug your experiment is to find a cheap and quick case that reproduces these problems. Ideally you will have a case that can be run interactively within less than a minute or so.  Try seeing whether a 2-day run reproduces a pair of 1-day runs with a restart in the middle. If this case is too large to conveniently run interactively, try a lower resolution version of this case to see whether it exhibits the same pathology of differing across restarts.

[...]

PR #1192, opened 2020-08-29
Always set VARIABLE_BUOYFORCE = True with RESTOREBUOY = True

hmkhatri:
Here is a minor addition in the `MOM_surface_forcing.F90` module. For `RESTOREBUOY = True`, modified code always sets `VARIABLE_BUOYFORCE` to be True (default option in MOM6), even if the user specifies the opposite in `MOM_input` file. A warning message is also printed.

Setting `VARIABLE_BUOYFORCE = False` with `RESTOREBUOY = True` leads to diverging solutions, e.g. in #1157. 

@Hallberg-NOAA Could you please have a quick look?


marshallward:
Hi Hemant, could you change this so that the error is FATAL rather than a WARNING, along with a update to the error message?  It's not a good habit to allow invalid configuration files to persist.

marshallward:
Gaea regression test: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11116

marshallward (approved):

PR #1197, opened 2020-09-02
Autoconf-based build framework

marshallward:
This patch is a first draft of an autoconf-based build for MOM6.

This is a hybrid solution which relies on mkmf and list_paths to
automatically generate the model dependencies, but uses autoconf to
determine the necessary compiler flags for building the model.

Recommended instructions for testing the build system, as well as
instructions for new users, are shown below.
```
    # (Optional) Fetch mkmf and build FMS
    cd deps
    make -j
    cd ..

    # Set up the autoconf build files
    autoreconf

    # Configure and build the model in /build
    mkdir -p build
    cd build
    ../configure
    make -j
```

As with standard autoconf builds, compilers and flags can be configured
with prepended arguments, e.g.
```
    FC=mpifort FCFLAGS="-g -O0 -Wall -Wextra" ./configure
    make
```

Introduction of autoconf configuration should not interfere with any
existing mkmf-based build procedures.

A more detailed discussion is provided below.

----

At minimum, only the following instructions would be required:
```
    autoreconf
    ./configure
    make
```
but this requires that both mkmf and FMS be accessible in standard
searchable directories.

A more realistic MOM6-only build would be as shown below
```
    autoreconf
    PATH="path/to/mkmf/bin:${PATH}" \
      FCFLAGS="-Ipath/to/fms_mods" \
      LDFLAGS="-Lpath/to/fms_lib" \
      ./configure
    make
```
where `path/to/xxx` points to the location of the pre-installed content.

To alleviate this issue, a Makefile in the `deps` directory has been
provided, which checks out copies of mkmf and FMS and places them in the
`deps` directory.  If the MOM6 `./configure` script is unable to find
mkmf or FMS, then it will automatically continue to search inside of the
`deps` directory.

----

The `deps` directory provides templates for an autoconf build of FMS
resembling our MOM6 build.

The Makefile provides the following rules:

`make` or `make all`:
  Fetch mkmf and FMS, build FMS, install module files to `deps/include`
  and `libFMS.a` to `deps/lib`.

`make clean`
  Delete the compiled FMS files in `deps/fms/build` but not the
  installed libraries or module files.

`make distclean`
  Delete all files produced by the Makefile.

Much of the FMS source configuration has been moved to the `deps`
Makefile.  The available hooks and default values are shown below.
```
MKMF_URL ?= https://github.com/NOAA-GFDL/mkmf.git
MKMF_COMMIT ?= master
FMS_URL ?= https://github.com/NOAA-GFDL/FMS.git
FMS_COMMIT ?= 2019.01.03
FCFLAGS_FMS ?=
```
The FCFLAGS_FMS hook is provided to override the default autoconf value,
`-g -O2`.

Although FMS provides its own automake build framework, the 2019.01.03
release cannot be used with MOM6 due to missing components in its
libtool configuration (`random_number`).  The build time is also
significantly longer due to incorrect dependencies in the automake
configuration.

Most of the issues describe above have either been or will soon be
resolved in newer FMS releases, and we may be able to transition to
the FMS automake in a future release.

----

The .testing directory has been reworked to use the autoconf builds of
MOM6 and FMS, but should otherwise work the same as before.
```
    cd .testing
    make -j
    make -j test
```
The changes required to use autoconf have been incorporated into the
`.testing` Makefile.

This can create issues with the management of FMS, which is now
configured inside of `deps/Makefile` rather than in `.testing`.  Also, a
rebuild of FMS will require an explicit wipe of FMS in `deps`, which can
be done with the following command.
```
    make -C ../deps distclean
```

Platform-specific flags are now managed in the `config.mk` file rather
than in a mkmf template.  The following hooks are provided, with default
values shown.
```
FCFLAGS_DEBUG ?= -g -O0
FCFLAGS_REPRO ?= -g -O2
FCFLAGS_INIT ?=
FCFLAGS_COVERAGE ?=
```
The intention is to provide some correspondence to the GFDL DEBUG and
REPRO production builds.

Any FMS build produced by `make` in `.testing` will use the
`FCFLAGS_DEBUG` variable.

Although `FCFLAGS_INIT` is intended for intialization, it acts as more
of a general-purpose flag which is applied to MOM6 but not FMS.

----

Other points of interest

- This build system only uses autoconf.  It does not use automake or libtools.

- autoconf will determine any necessary flags required for building
  MOM6, but does not attempt to use optional flags for debugging or
  performance.

  Flags required for testing or production builds must still track or
  otherwise manage their own compiler flags (FCFLAGS, etc).

- The MOM6 `./configure` script requires a FMS library and module files as
  part of its build validation.  This is why we must pre-build FMS
  before calling `./configure`.

- FCFLAGS defaults to `-g -O2`.  Any explicit setting of FCFLAGS
  will remove and replace these flags.  Other flags, such as
  `-fdefault-real-8` will be appended as needed and do not need to be
  managed.

- Support M4 macros have been tested on GCC, Intel, PGI, and Cray
  compilers.  The macros can be extended to other compilers and older
  versions of these compilers if necessary.

----

Known issues

- The macro used for the MPI launcher (AX_MPI) can potentially locate a
  MPIFC (e.g. mpif90) which does not match FC.  For example, FC will
  tend to default to GFortran, but mpif90 may use Intel Fortran.

  While FC is quickly replaced with MPIFC, there are some initial tests
  which may fail if FCFLAGS was defined for the non-GFortran compiler.

  This can be resolved by explicitly setting both FC and FCFLAGS for the
  same compiler.

- Since we use git for both development and distribution, we may want to
  pre-build the configure script and support files, eliminating the
  `autoreconf` step.


marshallward:
I think there are two issues here, first with coverage (probably missing from LDFLAGS) and the other with the target build unable to find the local FMS library.  I will try to sort these out in the next commit.

adcroft:
`./configure` got stuck at
```
checking for MPI_Init...
```
This is using the usual gnu environment on gaea.

marshallward:
I saw this a few times, it appears to be a Lustre issue.  I believe it's hanging because it cannot create the test file.

marshallward:
The behavior has been modified based on feedback from @adcroft.

Autoconf configuration has now been moved into the `ac` directory.  This breaks convention as well as the expected `./configure && make` ritual, but it isolates the build configuration files from the codebase.  Pathing is occasionally awkward and redundant (e.g. `<...>/ac/../ac/<...>` but is otherwise robust to the usual autoconf pathing rules.

To create the `configure` script:
```
autoreconf -if ac/configure
```
We may also want to pre-commit the output files (`configure`, `aclocal.m4`, `autom4te.cache/`), even if they can be generated.

The `deps` directory has also been moved into this `ac` directory, isolating it from the source.  To pre-build FMS and pre-fetch mkmf:
```
make -C ac/deps
```

To build MOM6:
```
cd path/to/some/dir
${codebase}/ac/configure
```
where `${codebase}` points to wherever the MOM6 source was checked out.

MOM6 `.testing` has been updated to build its own FMS library, rather than rely on `ac/deps`, but otherwise should operate normally:
```
cd .testing
make -j
make -j test
```
While `.testing` has been refactored to use autoconf, none of the instructions above should be required to run the test suite.

marshallward:
This is passing now on Travis, but some debug/repro differences are occurring on Gaea which need further investigation.

marshallward:
Recent changes include some improved support for macOS, as well as some bugfixes in the Makefile rule dependencies.

I think this is close enough to be operational, assuming we are happy with how to use it.  But it still needs documentation.

adcroft (approved):

adcroft:
This is ready except for documentation.

ac/README.md
```diff
@@ -0,0 +1,184 @@
...
+configured for the MPI wrapper but is incompatible with the `FC` compiler.
+
+To resolve this, ensure that `FC` and `FCFLAGS` are specified for the same
+compiler.
+
+
+## Coupled builds
+
+The Autoconf build is currently only capable of building ocean-only
+executables, and cannot yet be build as part of a coupled more, nor as a
```
raphaeldussin, line 183: -more +model ?

PR #1200, opened 2020-09-03
OBC: segment parser refactoring

marshallward:
The segment configuration parser was encountering problems with GFortran
10.2, where the `fields` array update was being removed by the -O2
optimizer.

In fact, the `parse_segment_data_str` function was doing two separate
operations, where the first call would determine the number of fields
and save their names into an array, and the second call would parse the
data contents of each input field.  The presence of optional arguments
were used to effectively select the preferred operation.

It is possible that the presence of these optional arguments was
interfering with optimization, causing the removal of the `fields`
update.

While this is most likely a bug in the GFortran compiler, we address the
problem by instead splitting this function into two independent
functions, which allows us to remove the optional arguments.

When split, the function appears to work correctly under O2
optimization.

### DIFF
--- a/src/core/MOM_open_boundary.F90
+++ b/src/core/MOM_open_boundary.F90
@@ -669,7 +669,7 @@ subroutine initialize_segment_data(G, OBC, PF)
       call MOM_error(FATAL, mesg)
     endif
 
-    call parse_segment_data_str(trim(segstr), fields=fields, num_fields=num_fields)
+    call parse_segment_manifest_str(trim(segstr), num_fields, fields)
     if (num_fields == 0) then
       call MOM_mesg('initialize_segment_data: num_fields = 0')
       cycle ! cycle to next segment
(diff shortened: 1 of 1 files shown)


marshallward:
Someday it might be better to rewrite this parser to populate a more organized data structure, but I think this change is good enough for now.

kshedstrom:
Either this or another recent change is causing my seamount cases to fail:
`FATAL from PE     3: mpp_sync_self: size_recv does not match of data received`

marshallward:
@kshedstrom Can you either send me the experiment or revert to dev/gfdl and test again?

kshedstrom:
It is definitely dev/gfdl which is the problem. I'll go to an earlier version and try your patch.

kshedstrom:
If I go to origin/dev/esmg and cherry-pick your patch, all is well. It does include Andrew's tides so maybe they are at odds with one of the other updates - I'll have to investigate further.

I approve this PR.

marshallward:
Thanks for checking Kate, there was a good chance I would make a mistake with this.

adcroft:
@kshedstrom , @MJHarrison-GFDL  we definitely need you to try this branch since we only have two tests that will exercise this code.

MJHarrison-GFDL:
@marshallward . Having trouble building (cd .testing; make TEMPLATE=./linux-ubuntu-xenial-gnu.mk compile).

MJHarrison-GFDL:
@kshedstrom . seamount cases pass with a clean build.

MJHarrison-GFDL:
I am OK with these changes.

adcroft (approved):
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11152

Issue #1201, opened 2020-09-04
Fatal error - but only when optimized


kshedstrom:
After a recent update, my seamount cases are failing with:
`FATAL from PE   3: mpp_sync_self: size_recv does not match of data received`
I used "git bisect" to track it to:
```
commit e8b5d96bcb769b79f68a671e35f4f02940d7cab5
Author: Robert Hallberg <Robert.Hallberg@noaa.gov>
Date:  Wed Aug 19 18:47:35 2020 -0400
```
Running in debug mode to see if I can find out more, it doesn't fail. This is with gnu fortran 8.3.0. I have been unable to get everything to run with the intel compiler, so haven't regularly tried it.

Hallberg-NOAA:
I have tried to reproduce these errors with the dev/gfdl MOM6-examples seamount test cases and the gnu 8.3.0 compiler but have failed to do so. It is possible that this is because our compiler flags or our experiments differed. However, given the nature of the error message, I have some suggestions for how to get at what is going on here.

  The error message you quote indicates that this is a problem with a halo update, and there are two different halo updates associated with this PR. We should be able to identify which is at fault by commenting them out successively. Because the code ran before (albeit incorrectly), it should run again when we comment out the problematic halo update.  The line numbers I will refer to below are those from https://github.com/NOAA-GFDL/MOM6/pull/1184/files. and they might differ from the line numbers in the version of the code you are using.

  The call to `pass_var(fluxes%p_surf,...` on line 1256 of MOM.F90 is only used in coupled models, so it seems to be problematic, but because the test for whether to do this is based on whether a pointer is associated, if we had failed to nullify it at start-up, it might lead to problems. So we should try commenting out this pass var call.

[...]

PR #1202, opened 2020-09-08
Properly set vertex_shear in set_diffusivity_init

Hallberg-NOAA:
Always call kappa_shear_at_vertex so that the Vertex_shear element of the
control structure for the MOM_set_diffusivity module is always actively set.  It
had previously been initialized only if USE_JACKSON_PARAM is true, which led to
the issues reported in MOM6 issue #1201.  Also clarified or corrected some of
the comments in MOM_kappa_shear.  All answers will be bitwise identical in all
cases that worked previously, and this PR corrects the problems noted in issue
#1201, which can closed once it is accepted.


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11160

marshallward (approved):

marshallward:
Overriding hanging Travis build.  Otherwise everything passes.

Issue #1206, opened 2020-09-11
Temperature, salinity and tracer concentration diagnostics are posted before diabatic forcing is applied


Hallberg-NOAA:
The diagnostics of the 3-dimensional tracer states, including temperature and salinity, are being posted _before_ any boundary forcing or vertical mixing have been applied. This means that there is an O(dt) error in all of these 3-d tracer state diagnostics, and an inconsistency between fields like that sea surface temperature or surface salinity that are calculated at the correct point in the algorithm and the 3-d temperature and salinity diagnostics.

  In a carefully executed and thoughtful exercise, https://github.com/NOAA-GFDL/MOM6/pull/718 rearranged the MOM6 code to change the point in the algorithm when a number of diagnostics were posted to create remapped tracer transport budgets that close in non-native coordinates. While this exercise did achieve its objectives, in moving the diagnostics from the 3-d tracer concentrations (including temperature and salinity) from post_tracer_diagnostics to post_tracer_transport_diagnostics, it also moved these diagnostics to an earlier point in the algorithm before `diabatic()` has been called unless DIABATIC_FIRST = True (which is not the default value). There is therefore an O(dt) error that was introduced for the 3-d diagnostics of temperature and salinity.

  Thankfully this issue does not impact any of the generic tracers, such as those in biogeochemical packages like bling, TOPAZ or COBALT, because they set `registry_diags=False` in their `register_tracer()` calls, and are therefore responsible for all their own diagnostics.

  The more extensive automated regression testing of diagnostics that has been implemented for MOM6 via Travis since https://github.com/NOAA-GFDL/MOM6/pull/718 would have highlighted the changes, and I think it likely that these issues would have been addressed before the PR was accepted.

 A figure illustrating the severity of this issue by comparing the SST diagnostic with the surface value of the 3-d temperature diagnostic in images taken 6 hours apart in a variant OM4 is attached below, courtesy of Xu Li from NOAA/EMC, is included below.

![image](https://user-images.githubusercontent.com/6529173/92910226-d8ea6b00-f3f5-11ea-9e2b-3a302552a380.png)

ashao:
I agree that my previous PR was the cause of this. However, while this does result in SST and 3D-temperature not being consistent, it's still a valid 'diagnostic' it's just that within a timestep temperature is being posted at a different time. The larger problem in my view is that T, S are not consistent with the other tracers when we have taken great care within the code NOT to distinguish them from other tracers.

[...]

PR #1207, opened 2020-09-11
Post tracer concentration when synced

ashao:
Tracer concentrations were being posted at the same time as the
transport diagnostics. This leads to some unexpected behavior (e.g. SST
and 3D temperature are not consistent) because the concentrations are
potentially being posted at an intermediate state.

In addition to fixing this problem for all tracers, the subroutine has
been renamed to be more descriptive of what kinds of diagnostics might
fall into this category.

Proposed solution for  https://github.com/NOAA-GFDL/MOM6/issues/1206


ashao:
@Hallberg-NOAA is too fast for me. I was literally pushing when you made your edit. Now the MOM6 repository will never have the following comment:
> Neglected subroutine in new declaration    
    Careless me left out 'subroutine' in the declaration of
    post_tracer_diagnostics_at_sync

Hallberg-NOAA:
The pipeline testing for this PR is at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11165 .

ashao:
I'll post images of the diagnostics over the weekend.

Hallberg-NOAA:
The updated version of this PR is being tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11170.

Hallberg-NOAA (approved):
I agree that these are the right changes to address issue #1206, and that the new diagnostic to allow the state of tracers just after advection and horizontal diffusion will help in gracefully transitioning diagnostics for existing experiments.  I am therefore approving this PR.

  However, I do not agree with the change to the name of the subroutine post_tracer_diagnostics to post_tracer_diagnostics_at_sync, and I will be putting in a follow up pull request to revert this to its previous name.  This should be the primary routine for diagnostics of the state of tracers, and I think that the newly extended subroutine name just makes this key point confusing, as was discussed in a call concerning this PR.

PR #1210, opened 2020-09-18
Tc4 ice shelf

MJHarrison-GFDL:
@marshallward . tc4.rotate is failing due to intermediate timestats diverging although the final stats are identical.
image


adcroft:
It's not just the rotation test that is falling. There's restart and dimensional tests too.

adcroft:
The tc4 test seems broken in many ways. Even layout is failing trying to read the grid. Is everything committed?
```
FATAL from PE     0: fms_io(read_data_3d_new), field x in file ./ocean_hgrid.nc: field size mismatch 1
```

adcroft:
That last was because the size has changed (which is good) - cleaning the work dir sorted that out. The actual fails I get interactively are
```
FAIL: Solutions tc4.restart have changed.
FAIL: Diagnostics tc4.dim.t.diag have changed.
FAIL: Diagnostics tc4.dim.l.diag have changed.
FAIL: Diagnostics tc4.dim.h.diag have changed.
FAIL: Solutions tc4.rotate have changed.
```
which I think is consistent with what Travis reported.

adcroft:
After a restart `fluxes%sens` and other member fields are zero where there aren't before the restart?

MJHarrison-GFDL:
From a clean build I get the same. The restart and diagnosric issues I haven't looked at but the rotate fail is weird . The results only differ at timesteps 54 and 63 but are otherwise identical through 144.

MJHarrison-GFDL:
buoy_config="zero".

MJHarrison-GFDL:
Energetics differ temporarily at the last significant digit .Timestats too precise?

marshallward:
I see on my desktop that `SB_max` is undefined, raising a FPE:
```
Thread 1 "MOM6" received signal SIGFPE, Arithmetic exception.
0x00005555559356b0 in mom_ice_shelf::shelf_calc_flux (sfc_state_in=..., fluxes_in=0x555558970190, time=..., time_step=1200, cs=0x555558e676d0)
    at ../../../ac/../src/ice_shelf/MOM_ice_shelf.F90:592
592	                if (Sb_max_set .and. (Sbdry(i,j) > Sb_max)) &
(gdb) p Sb_max
$2 = nan(0x4000000000000)
```
but it is worrying that Travis was unable to detect this.

adcroft:
Should
```fortran
if (Sb_max_set .and. (Sbdry(i,j) > Sb_max))
```
be
```fortran
if (Sb_max_set) then
  if (Sbdry(i,j) > Sb_max)) &
```
?

marshallward:
Yes, I agree, no short-circuit logic can be assumed Fortran

MJHarrison-GFDL:
Registering sfc_state%tau[xy]_shelf  restarts fixes that issue.

MJHarrison-GFDL:
tc4 has thermal coupling turned on so buoyancy forcing is non-zero, but not required for restarts.

src/initialization/MOM_state_initialization.F90
```diff
@@ -2214,24 +2226,25 @@ subroutine MOM_temp_salt_initialize_from_Z(h, tv, G, GV, US, PF, just_read_param
...
-    ! Compute fractional ice shelf coverage of h
-    do j=jsd,jed ; do i=isd,ied
-      if (G%areaT(i,j) > 0.0) &
-        frac_shelf_h(i,j) = area_shelf_h(i,j) / G%areaT(i,j)
-    enddo ; enddo
-    ! Pass to the pointer for use as an argument to regridding_main
-    shelf_area => frac_shelf_h
-
-  endif
+  ! ! This is needed for building an ALE grid under ice shelves
```
adcroft, line 2229: Ditto

src/parameterizations/vertical/MOM_ALE_sponge.F90
```diff
@@ -847,11 +847,11 @@ subroutine apply_ALE_sponge(h, dt, G, GV, US, CS, Time)
           ! Build the source grid
           zTopOfCell = 0. ; zBottomOfCell = 0. ; nPoints = 0; hsrc(:) = 0.0; tmpT1d(:) = -99.9
           do k=1,nz_data
-             if (mask_z(CS%col_i(c),CS%col_j(c),k) == 1.0) then
```
adcroft, line 850: What does this change have to do with ice shelves? It's not inside a `frac_shelf` block so seems unrelated to the mods for tc4.

src/initialization/MOM_state_initialization.F90
```diff
@@ -2125,16 +2137,16 @@ subroutine MOM_temp_salt_initialize_from_Z(h, tv, G, GV, US, PF, just_read_param
...
-    call get_param(PF, mdl, "ICE_THICKNESS_FILE", ice_shelf_file, &
-                 "The file from which the ice bathymetry and area are read.", &
-                 fail_if_missing=.not.just_read, do_not_log=just_read)
-    shelf_file = trim(inputdir)//trim(ice_shelf_file)
-    if (.not.just_read) call log_param(PF, mdl, "INPUTDIR/THICKNESS_FILE", shelf_file)
-    call get_param(PF, mdl, "ICE_AREA_VARNAME", area_varname, &
-                 "The name of the area variable in ICE_THICKNESS_FILE.", &
-                 fail_if_missing=.not.just_read, do_not_log=just_read)
-  endif
+  ! if (use_ice_shelf) then
```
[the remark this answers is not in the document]
MJHarrison-GFDL: Resolved. I refactored the ice shelf module and it is now being initialized before MOM. The way the code was written, we were initializing frac_shelf  independently in the ice shelf and in MOM.

src/initialization/MOM_state_initialization.F90
MJHarrison-GFDL: Resolved.

src/parameterizations/vertical/MOM_ALE_sponge.F90
MJHarrison-GFDL: Nothing to do with ice shelves, but triggered in tc4 when the sponges are turned on.

src/ice_shelf/MOM_ice_shelf.F90
```diff
@@ -772,26 +805,43 @@ end subroutine change_thickness_using_melt
...
+
+  G=>CS%Grid
+
   is = G%isc ; ie = G%iec ; js = G%jsc ; je = G%jec
   isd = G%isd ; jsd = G%jsd ; ied = G%ied ; jed = G%jed
 
-  if ((CS%grid%isc /= G%isc) .or. (CS%grid%iec /= G%iec) .or. &
-      (CS%grid%jsc /= G%jsc) .or. (CS%grid%jec /= G%jec)) &
-    call MOM_error(FATAL,"add_shelf_forces: Incompatible ocean and ice shelf grids.")
+!  if ((CS%grid_in%isc /= G%isc) .or. (CS%grid_in%iec /= G%iec) .or. &
```
adcroft, line 842: Delete or uncomment? Why would this test not be valid now?

src/initialization/MOM_state_initialization.F90
```diff
@@ -1993,7 +2005,7 @@ subroutine MOM_temp_salt_initialize_from_Z(h, tv, G, GV, US, PF, just_read_param
   real    :: eps_rho  ! A negligibly small density difference [R ~> kg m-3].
   real    :: PI_180   ! for conversion from degrees to radians
 
-  real, dimension(:,:), pointer :: shelf_area => NULL()
+!  real, dimension(:,:), pointer :: shelf_area => NULL()
```
adcroft, line 2008: delete

src/initialization/MOM_state_initialization.F90
```diff
@@ -2021,8 +2033,8 @@ subroutine MOM_temp_salt_initialize_from_Z(h, tv, G, GV, US, PF, just_read_param
 
   ! Local variables for ALE remapping
   real, dimension(:), allocatable :: hTarget ! Target thicknesses [Z ~> m].
-  real, dimension(:,:), allocatable :: area_shelf_h ! Shelf-covered area per grid cell [L2 ~> m2]
-  real, dimension(:,:), allocatable, target :: frac_shelf_h ! Fractional shelf area per grid cell [nondim]
+!  real, dimension(:,:), allocatable :: area_shelf_h ! Shelf-covered area per grid cell [L2 ~> m2]
```
adcroft, line 2036: delete

src/initialization/MOM_state_initialization.F90
```diff
@@ -2197,8 +2201,8 @@ subroutine MOM_temp_salt_initialize_from_Z(h, tv, G, GV, US, PF, just_read_param
   do k=1,size(Z_edges_in,1) ; Z_edges_in(k) = -Z_edges_in(k) ; enddo
 
   allocate(rho_z(isd:ied,jsd:jed,kd))
-  allocate(area_shelf_h(isd:ied,jsd:jed))
-  allocate(frac_shelf_h(isd:ied,jsd:jed))
+  !allocate(area_shelf_h(isd:ied,jsd:jed))
```
adcroft, line 2204: delete

src/ice_shelf/MOM_ice_shelf.F90
MJHarrison-GFDL: resolved. The grid checks are not necessary here any longer. If this is being called from the driver, the fluxes need rotation, otherwise if it is being called from within the ice shelf module, they do not require rotation.

src/initialization/MOM_state_initialization.F90
MJHarrison-GFDL: resolved.

src/ice_shelf/MOM_ice_shelf_diag_mediator.F90
```diff
@@ -0,0 +1,812 @@
...
+  !!@}
+  type(axesType) :: axesZi !< A 1-D z-space axis at interfaces
+  type(axesType) :: axesZL !< A 1-D z-space axis at layer centers
+
+  ! Mask arrays for diagnostics
+  real, dimension(:,:),   pointer :: mask2dT   => null() !< 2D mask array for cell-center points
+  real, dimension(:,:),   pointer :: mask2dBu  => null() !< 2D mask array for cell-corners
+  real, dimension(:,:),   pointer :: mask2dCu  => null() !< 2D mask array for east-faces
+  real, dimension(:,:),   pointer :: mask2dCv  => null() !< 2D mask array for north-faces
+  real, dimension(:,:,:), pointer :: mask3dTL  => null() !< 3D mask array for layer cell-centers
```
Hallberg-NOAA, line 94: There are no vertical axes used or defined for the ice-shelf diag mediator, so why are all of these 3d mask variables needed?  It seems to me that the whole point here is to have a 2-d diag mediator type for the use of the ice-shelf code.

Also, it seems like a recipe for confusion to use the same type and subroutine names for the ice-shelf diag mediator as are used by the broader MOM6 diag mediator. Would it make sense to give these types and subroutines unique names to avoid confusion?

src/core/MOM.F90
```diff
@@ -3051,16 +3083,27 @@ subroutine extract_surface_state(CS, sfc_state_in)
   if (CS%rotate_index) &
     turns = G%HI%turns
 
-  if (.not.sfc_state_in%arrays_allocated) &
+  if (.not.sfc_state_in%arrays_allocated)  then
     !  Consider using a run-time flag to determine whether to do the vertical
     ! integrals, since the 3-d sums are not negligible in cost.
-    call allocate_surface_state(sfc_state_in, G_in, use_temperature, &
-        do_integrals=.true., omit_frazil=.not.associated(CS%tv%frazil))
+    if (associated(CS%frac_shelf_h)) then
```
Hallberg-NOAA, line 3089: This could be coded more parsimoniously here and at line 3101 by introducing a logical variable `use_iceshelves = associated(CS%frac_shelf_h)` that is passed to what would become 2 calls (instead of 4) to `allocate_surface_state()`.

src/core/MOM.F90
```diff
@@ -2412,12 +2444,29 @@ subroutine initialize_MOM(Time, Time_init, param_file, dirs, CS, restart_CSp, &
...
+    else
+      call MOM_initialize_state(CS%u, CS%v, CS%h, CS%tv, Time, G, GV, US, &
+          param_file, dirs, restart_CSp, CS%ALE_CSp, CS%tracer_Reg, &
+          CS%sponge_CSp, CS%ALE_sponge_CSp, CS%OBC, Time_in)
+    endif
   endif
 
+  if (use_ice_shelf) &
+    call hchksum(CS%frac_shelf_h, "MOM:frac_shelf_h", G%HI, &
+        haloshift=0, scale=GV%H_to_m)
```
Hallberg-NOAA, line 2468: `GV%H_to_m` is not the right scaling factor for what I believe to be a nondimensional variable. Either the scale argument to this `hchksum()` call is wrong or the documentation describing the units of CS%frac_shelf_h is wrong. Regardless of which is the case, they need to be made consistent and correct.

src/ALE/coord_rho.F90
```diff
@@ -112,10 +116,25 @@ subroutine build_rho_column(CS, nz, depth, h, T, S, eqn_of_state, z_interface, &
   real, dimension(nz+1) :: xTmp   ! Temporary positions [H ~> m or kg m-2]
   real, dimension(CS%nk) :: h_new ! New thicknesses [H ~> m or kg m-2]
   real, dimension(CS%nk+1) :: x1  ! Interface heights [H ~> m or kg m-2]
+  real :: z0_top, eta ! Thicknesses or heights [Z ~> m] or [H ~> m or kg m-2]
+  logical :: new_rho_def
```
Hallberg-NOAA, line 120: The variable `new_rho_def` appears to be set but never used, so perhaps it should be deleted.  If it is retained, a more descriptive name might more appropriate. (Everything new becomes old.)

src/ALE/MOM_regridding.F90
```diff
@@ -1371,6 +1376,8 @@ subroutine build_rho_grid( G, GV, US, h, tv, dzInterface, remapCS, CS )
   endif
 
   nz = GV%ke
+  ice_shelf = .false.
+  if (present(frac_shelf_h)) ice_shelf = .true.
```
Hallberg-NOAA, line 1380: These 2 lines could be written more simply as `ice_shelf = present(frac_shelf_h)`.

src/ALE/MOM_regridding.F90
```diff
@@ -871,8 +871,12 @@ subroutine regridding_main( remapCS, CS, G, GV, h, tv, h_new, dzInterface, frac_
       call calc_h_new_by_dz(CS, G, GV, h, dzInterface, h_new)
 
     case ( REGRIDDING_RHO )
-      if (do_convective_adjustment) call convective_adjustment(G, GV, h, tv)
-      call build_rho_grid( G, GV, G%US, h, tv, dzInterface, remapCS, CS )
+       if (do_convective_adjustment) call convective_adjustment(G, GV, h, tv)
+       if (use_ice_shelf) then
```
Hallberg-NOAA, line 875: Because frac_shelf_h is an optional argument both coming in and for `build_rho_grid()`, this 5 line block could be written as just the single line `call build_rho_grid( G, GV, G%US, h, tv, dzInterface, remapCS, CS, frac_shelf_h )`.

src/ice_shelf/MOM_ice_shelf.F90
```diff
@@ -772,39 +805,54 @@ end subroutine change_thickness_using_melt
 
 !> This subroutine adds the mechanical forcing fields and perhaps shelf areas, based on
 !! the ice state in ice_shelf_CS.
-subroutine add_shelf_forces(G, US, CS, forces, do_shelf_area)
```
Hallberg-NOAA, line 775: The forces and fluxes types use the ocean's grid for their halo sizes and indexing conventions, which may at some point be different from what is used in the ice-shelf code. The G argument with the ocean's grid structure should not have been removed here or in the subsequent interface to `add_shelf_pressure()`.  Adding a grid to the ice_shelf_CS is fine, but if there is a separate ice shelf grid, `add_shelf_forces()` and `add_shelf_pressure()` need to be written in a way that takes into account the fact that these two grids may be different!

src/initialization/MOM_state_initialization.F90
```diff
@@ -234,8 +239,12 @@ subroutine MOM_initialize_state(u, v, h, tv, Time, G, GV, US, PF, dirs, &
     if (.NOT.use_temperature) call MOM_error(FATAL,"MOM_initialize_state : "//&
        "use_temperature must be true if INIT_LAYERS_FROM_Z_FILE is true")
 
-    call MOM_temp_salt_initialize_from_Z(h, tv, G, GV, US, PF, just_read_params=just_read)
-
+    if (use_ice_shelf) then
```
Hallberg-NOAA, line 242: Because frac_shelf_h is an optional argument that is just being passed through, these 5 lines could be replaced by just the call on lines 243-244.

src/ice_shelf/MOM_ice_shelf.F90
```diff
@@ -1497,16 +1681,20 @@ subroutine initialize_ice_shelf(param_file, ocn_grid, Time, CS, diag, forces, fl
...
+     call initialize_ice_thickness(ISS%h_shelf, ISS%area_shelf_h, ISS%hmask, CS%Grid, CS%Grid_in, US, param_file,&
+          CS%rotate_index, CS%turns)
     ! next make sure mass is consistent with thickness
     do j=G%jsd,G%jed ; do i=G%isd,G%ied
       if ((ISS%hmask(i,j) == 1) .or. (ISS%hmask(i,j) == 2)) then
         ISS%mass_shelf(i,j) = ISS%h_shelf(i,j)*CS%density_ice
       endif
     enddo ; enddo
+    if (CS%debug) then
+       call hchksum(ISS%mass_shelf, "IS init: mass_shelf", G%HI, haloshift=0)
```
Hallberg-NOAA, line 1694: Please check the units of mass_shelf and area_shelf_h, and add scaling arguments to these `hchksum()` calls as necessary.

src/ice_shelf/MOM_ice_shelf.F90
```diff
@@ -1539,34 +1727,44 @@ subroutine initialize_ice_shelf(param_file, ocn_grid, Time, CS, diag, forces, fl
...
   enddo ; enddo
-  if (present(fluxes)) then ; do j=jsd,jed ; do i=isd,ied
-    if (G%areaT(i,j) > 0.0) fluxes%frac_shelf_h(i,j) = ISS%area_shelf_h(i,j) / G%areaT(i,j)
+  if (present(fluxes_in)) then ; do j=jsd,jed ; do i=isd,ied
+    if (G%areaT(i,j)>0.) fluxes%frac_shelf_h(i,j) = ISS%area_shelf_h(i,j) / G%areaT(i,j)
   enddo ; enddo ; endif
 
   if (CS%debug) then
     call hchksum(fluxes%frac_shelf_h, "IS init: frac_shelf_h", G%HI, haloshift=0)
+    call hchksum(ISS%area_shelf_h, "IS init: area_shelf_h", G%HI, haloshift=0)
```
Hallberg-NOAA, line 1760: I believe that the two `hchksum()` calls for `ISS%area_shelf_h` and `G%areaT` need dimensional rescaling arguments.

src/ALE/MOM_ALE.F90
```diff
@@ -633,7 +631,7 @@ subroutine ALE_build_grid( G, GV, regridCS, remapCS, h, tv, debug, frac_shelf_h
   if (show_call_tree) call callTree_enter("ALE_build_grid(), MOM_ALE.F90")
   use_ice_shelf = .false.
```
Hallberg-NOAA, line 632: These 4 lines could be written more simply as ice_shelf = present(frac_shelf_h). Similar comments apply elsewhere now that frac_shelf_h is not a pointer but a simple optional array.

src/ice_shelf/MOM_ice_shelf_diag_mediator.F90
MJHarrison-GFDL: Removed the vertical axis and mask lines.

I don't see a problem with namespace conflict, i.e. we are not simultaneously dealing with MOM6 and IceShelf diagnostics within a single module. It would perhaps make sense to *not* have separate diag mediators for  MOM6, SIS2 and the IceShelf. They could be combined in a single module which is overloaded to work with all 3 data structures.

src/core/MOM.F90
MJHarrison-GFDL: The scale factor is wrong. I am not sure why this is passing the rescaling tests. @marshallward ?

adcroft: I assume `use_ice_shelf=.true.` ?

src/ice_shelf/MOM_ice_shelf.F90
MJHarrison-GFDL: The fluxes and forcing types use the unrotated grid  which is pointed to by the shelf structure, so I do not believe we need to retain it in the interface.   The ice shelf is creating a grid using the same parameter file as MOM6. If we were to use a separate grid and PF for the shelf, then we could use the ocean grid pointer which is still an argument to initialize_ice_shelf.

src/core/MOM.F90
MJHarrison-GFDL: Yes

marshallward: This will be fixed after #1220 is merged.  It was requested that I submit those fixes in a separate PR.

src/ice_shelf/MOM_ice_shelf.F90
MJHarrison-GFDL: I returned the ocean grid to the interface to add_shelf_forces and included a grid compatibility check

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11417

marshallward:
There appears to be some kind of error in `coupled_driver` which is causing most of the builds to fail.  This would not have been detected by Travis.

This is what popped up but it's buried amongst too many errors in a parallel build to make more sense:
```
build/gnu/debug/dynamic/mom6/libmom6.a: Error: Actual argument for 'forces_in' must be a pointer at (1)
build/gnu/debug/dynamic/mom6/libmom6.a: ../../../../../..//config_src/coupled_driver/ocean_model_MOM.F90:529:43:
```

@MJHarrison-GFDL can you have a look?

MJHarrison-GFDL:
PR on the way.

MJHarrison-GFDL:
The current regression failure in tc4 is due to a [bugfix](https://github.com/NOAA-GFDL/MOM6/pull/1210/commits/14a81e41152998ab024267f1d5269b6579009627) for the sponge code

marshallward:
Do you mind cherry-picking that one and submitting separately?  Might be better for the commit log as well.

PR #1213, opened 2020-09-23
Ac263 support

marshallward:
Additional changes were made to support older Autoconf versions (at
least as far back as 2.63).

* AC_FC_LINE_LENGTH, added in 2.67, was backported
* AC_OPENMP, bugged for Fortran until 2.69, tests against the C flag

Version restrictions are now explicit (2.63).

This may work with 2.62, when the m4_version_change macro was
introduced, but I have not yet tested it.

The Cray pointer macro's cache variable was also renamed.

### DIFF
--- a/ac/configure.ac
+++ b/ac/configure.ac
@@ -8,6 +8,8 @@
 # - We would probably run this inside of a script to avoid the explicit
 #   dependency on git.
 
+AC_PREREQ([2.63])
+
 AC_INIT(
   [MOM6],
   [ ],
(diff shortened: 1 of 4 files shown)


marshallward:
This did not resolve all of @kshedstrom's issues using 2.63 on her machine, but at the least I would say it is "necessary but not sufficient".

As the window to investigate further may have closed, I think this may as well be merged and we hope it resolves the issue.

adcroft (approved):

PR #1215, opened 2020-09-24
User/bgr/pe mld diag

breichl:
Adding new mixed layer depth diagnostics.

- MLD_EN1, MLD_EN2, MLD_EN3 are output containers for MLD diagnostics.
- The values are computed based on energy based mixed layer depth calculations.
- The energy values used can be set at runtime or be left as default.  The actual value used is written within the description of the variable in the NetCDF output, though I debated whether this feature was desirable or not (can be a bit messy with spacing from converting real to string, perhaps this can be done better).  The variable names are fixed.


src/parameterizations/vertical/MOM_diabatic_aux.F90
```diff
@@ -724,6 +724,208 @@ subroutine diagnoseMLDbyDensityDifference(id_MLD, h, tv, densityDiff, G, GV, US,
...
+
+              H_MixedLayer = H_MixedLayer + dz_mixed_N
+            else
+              Rho_MixedLayer = Rho_MixedLayer_0
+              H_MixedLayer   = H_MixedLayer_0
+            endif
+            mld(i,j,iM) = H_MixedLayer
+          enddo
+        else
+          mld(:,:,iM) = -1.e8 ! This shouldn't be seen, but if it is set it to something obviously wrong.
```
marshallward, line 900: Does this need dimensions?

marshallward: Guess not, since it's not meant to be processed, sorry about that.

breichl:
I'm working on the iteration, potentially will offer an improvement soon, and I also notice that Travis is failing so will need to figure out why.  But I am wondering, is the empty line at the beginning of MOM_diabatic_aux.F90 intentional?  My whitespace macro keeps removing it so I have to manually reinsert it.  It is missing again in my last commit so wanted to check before I reinsert it again.

adcroft:
Yes delete the blank line.

marshallward:
The new diagnostics should no longer signal an error.  In fact, this is passing the regression tests. 

Most of the reported fails are dimensional.  My guess is many of those hard-coded constants need dimensions applied to them.  Could also be that the diagnostics were improperly scaled?

breichl:
Thanks, good to know.  Will double check if that -1.e8 is triggering this or if it is in the scaling of the diagnostics (my hunch).

breichl:
The latest commit is a complete rewrite of the original commit, where the iterator used to calculate the fraction of a layer that could be mixed with a given energy has been replaced.  This should be faster and more accurate.

Hallberg-NOAA (approved):
I am very happy with this PR, especially now that it is using Newton's method and is finding the diagnostic mixed layer depth very quickly. It is a valuable diagnostic addition.

There are couple of places with implied array copies from non-stride-1 array syntax arguments that might be addressed for efficiency (e.g., line 792 of MOM_diabatic_aux.F90), and there is a curious empty logical test at lines 3451-3452 of MOM_diabatic_driver.F90 that probably should be deleted, but these are minor things that could be cleaned up later.

breichl:
Thanks for the review.  I can delete the lines at 3451, it is leftover code that should have been removed.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11308

PR #1216, opened 2020-09-25
MOM_sum_output: Reproducibility fix to energy

marshallward:
The total kinetic energy per unit mass, used heavily for model
reproducibility tests, was being computed in a non-reproducibile form:
```
   KE_mass = 0.5 * (u(i-1)**2 + u(i)**2 + v(j-1)**2 + v**2)
```
which uses an ambiguous order of summation.

This was causing reproducibility errors in rotational tests where u and
v become swapped.

This patch resolves the problem by explicitly summing the `u**2` and `v**2`
interpolations before adding together.

There is an extremely high change that this will change answers
somewhere and cause regressions.

### DIFF
--- a/src/diagnostics/MOM_sum_output.F90
+++ b/src/diagnostics/MOM_sum_output.F90
@@ -703,8 +703,9 @@ subroutine write_energy(u, v, h, tv, day, n, G, GV, US, CS, tracer_CSp, OBC, dt_
   tmp1(:,:,:) = 0.0
   do k=1,nz ; do j=js,je ; do i=is,ie
     tmp1(i,j,k) = (0.25 * KE_scale_factor * (areaTm(i,j) * h(i,j,k))) * &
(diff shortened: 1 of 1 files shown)


marshallward:
To clarify the final comment, numerical solutions will not change but the `ocean.stats` output used to detect regressions probably will change.

marshallward:
This will need to be merged before #1210 can be resolved.

adcroft (approved):
This does indeed change recorded numbers even though the solutions are obviously unchanged. I'll merge this on the terminal in order to update the regression files.

Tested at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11257

Interestingly , PGI did not change so the compiler happened to have chosen the same grouping of terms.
image

marshallward:
I wondered if this will help with some of the discrepancies between the compilers.  Seems like it is at least one less factor to worry about.

PR #1217, opened 2020-09-29
Diagnostics for depth-averaged acceleration from diapycnal mixing

hmkhatri:
Adding to 2D fractional-thickness weighted momentum budget diagnostics, diagnostics for zonal and meridional acceleration terms from diapycnal mixing have been added. 

Plus a small correction in read_data call for `Idamp` in sponge initialization.

### DIFF
--- a/src/initialization/MOM_state_initialization.F90
+++ b/src/initialization/MOM_state_initialization.F90
@@ -1795,7 +1795,7 @@ subroutine initialize_sponges_file(G, GV, US, use_temperature, tv, param_file, C
   if (new_sponges .and. .not. use_ALE) &
     call MOM_error(FATAL, " initialize_sponges: Newer sponges are currently unavailable in layered mode ")
 
-  call MOM_read_data(filename, "Idamp", Idamp(:,:), G%Domain, scale=US%T_to_s)
+  call MOM_read_data(filename, Idamp_var, Idamp(:,:), G%Domain, scale=US%T_to_s)
 
   ! Now register all of the fields which are damped in the sponge.
   ! By default, momentum is advected vertically within the sponge, but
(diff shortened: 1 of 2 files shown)


src/initialization/MOM_state_initialization.F90
```diff
@@ -1795,7 +1795,7 @@ subroutine initialize_sponges_file(G, GV, US, use_temperature, tv, param_file, C
   if (new_sponges .and. .not. use_ALE) &
     call MOM_error(FATAL, " initialize_sponges: Newer sponges are currently unavailable in layered mode ")
 
-  call MOM_read_data(filename, "Idamp", Idamp(:,:), G%Domain, scale=US%T_to_s)
+  call MOM_read_data(filename, Idamp_var, Idamp(:,:), G%Domain, scale=US%T_to_s)
```
Hallberg-NOAA, line 1798: On the one hand, I agree that this line should be using Idamp_var, but because the default value for SPONGE_IDAMP_VAR is "IDAMP", not "Idamp", and variable names are case-sensitive, this change will break some experiments.  This change is a good idea, but only if the default for SPONGE_IDAMP_VAR is simultaneously changed to "Idamp".  I am going to try to unilaterally make that change to the default for SPONGE_IDAMP_VAR.

Hallberg-NOAA: I have added the required corresponding change to the default for SPONGE_IDAMP_VAR.

Hallberg-NOAA:
The pipeline testing for this PR can be found at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11387.

marshallward:
Passes up to a parameter update.

marshallward (approved):

PR #1219, opened 2020-09-30
User/bgr/jhl threshold option

breichl:
Add an option to use the more restrictive tolerance check in place of the less restrictive tolerance check in MOM_kappa_shear and fix an iteration max criteria check.

- The new option (USE_RESTRICTIVE_TOLERANCE_CHECK) defaults to False to maintain the original JHL tolerance check.  Setting it True uses the more restrictive option, which appears to improve the scheme stability.
- Fix for itt number criteria check that should have been max_KS_it but was max_RiNo_it

The first will not change answers by default.  The second could potentially change answers, but only if the iteration count ever hits the cap max_KS_it.  It does not hit the cap default of 13 with USE_RESTRICTIVE_TOLERANCE_CHECK=False in the global_ALE test cases, and therefore it does not change answers in any of these cases.

### DIFF
--- a/src/parameterizations/vertical/MOM_kappa_shear.F90
+++ b/src/parameterizations/vertical/MOM_kappa_shear.F90
@@ -87,6 +87,10 @@ type, public :: Kappa_shear_CS ; private
                              !! time average TKE when there is mass in all layers.  Otherwise always
                              !! report the time-averaged TKE, as is currently done when there
                              !! are some massless layers.
+  logical :: restrictive_tolerance_check !< If false, uses the less restrictive tolerance check to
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA (approved):
I agree that these changes include a genuine bug fix in one case that could change answers in some unlikely instances (but note Murphy's Law), and a logical test that is helpful while preserving answers.

Because this changes the contents of the MOM_parameter_doc files and could change answers, this will have to be handled manually. These changes should have been noted in the commit descriptions, and the commit headers should have had the notation (*) and + at the start of the commit descriptions.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11323

PR #1220, opened 2020-10-05
Dimension: Diagnostic fixes

marshallward:
This patch fixes the dimensional scaling of three diagnostics:

- grid_Re_Ah
- grid_Re_Kh
- boundary_forcing_h_tendency

The fix to `grid_Re_Ah` and `grid_Re_Kh` removes the `Kh_min` and
`Ah_min` parameters, which could not be dimensionally scaled and
replaces them with variables in the control structure.

The minimum Laplacian and biharmonic viscosites are chosed to estimate
the smallest viscosity which is capable to altering the velocity.  We
use `spacing(1.) * L^2N / T` where `spacing()` is the smallest
difference in floating point.  We use the minimum harmonic grid spacing,
precomputed during initialization, rather than the local value, in order
to avoid additional computation and memory access during timestepping.

This will changes the value of grid_Re_[AK]h in certain isolated
examples of negligible viscosity, but will otherwise avoid the
possibility of divide-by-zero while also producing a realizable large
value at near-inviscid upper limit.

The second fix to `boundary_focing_h_tendency` had its T units set but
not its H units.  This patch fixes the dimensionality.

These are required to fix the dimensional verification errors in PR
1210.

### DIFF
--- a/src/parameterizations/lateral/MOM_hor_visc.F90
+++ b/src/parameterizations/lateral/MOM_hor_visc.F90
@@ -1722,11 +1722,12 @@ subroutine hor_visc_init(Time, G, US, param_file, diag, CS, MEKE, ADp)
                  "The absolute maximum value the GME coefficient is allowed to take.", &
                  units="m2 s-1", scale=US%m_to_L**2*US%T_to_s, default=1.0e7)
   endif
-  if (CS%Laplacian .or. CS%biharmonic) &
+  if (CS%Laplacian .or. CS%biharmonic) then
     call get_param(param_file, mdl, "DT", dt, &
                  "The (baroclinic) dynamics time step.", units="s", scale=US%s_to_T, &
                  fail_if_missing=.true.)
     Idt = 1.0 / dt
+  endif
   if (CS%no_slip .and. CS%biharmonic) &
     call MOM_error(FATAL,"ERROR: NOSLIP and BIHARMONIC cannot be defined "// &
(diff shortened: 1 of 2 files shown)


marshallward:
Sorry I just found a bug in this, missed by the testing.  I will update shortly.

marshallward:
This is required for #1210

gustavo-marques (approved):

Hallberg-NOAA (changes requested):
I may be wrong, but it looks to me like `min_grid_sp_h2` and `min_grid_sp_h4` in `hor_visc_init()` will vary between processors (unless a Cartesian grid is used) and that they will not be invariant to domain decomposition. If so, `CS%min_grid_Ah` and `CS%min_grid_Kh` will also not be invariant to domain decomposition. Please address this concern before these changes can be approved.

marshallward:
Thank you @Hallberg-NOAA , I have added `min_across_PEs` calls to account for multiple PEs and layouts.

Hallberg-NOAA (approved):
With the updates, this change should now reproduce properly across PE layouts, and it does address the issues with the hard-coded dimensional constants.

Hallberg-NOAA:
This PR is being tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11386.

Hallberg-NOAA:
The pipeline tests and Travis tests are all passing with this PR. It can be merged in.

Issue #850, opened 2018-09-05
Non-reproducibility across restarts for VERTEX_SHEAR=True


wfcooke:
With the SPEAR setup (1 degree ocean, symmetric) I am seeing differences in answers across restarts. 1x8 day vs 2x4 day.
For VERTEX_SHEAR=False I reproduce old answers.

Hallberg-NOAA:
Do the results reproduce across restarts when run in non-symmetric memory mode?

wfcooke:
@Hallberg-NOAA : Sorry Bob, I should have looked at the code rather then relying on memory. I am using the dynamic path (not dynamic_symmetric) of the code.

wfcooke:
@Hallberg-NOAA @adcroft 
I ran the symmetric code with VERTEX_SHEAR=True and DEBUG and found that u and v have different checksums immediately after restart.

There is an issue with the restart file.
@MJHarrison-GFDL had me take a look at MOM.res.nc and there is a difference in the redundant points.

yes? list/k=1/precision=6 u[j=35:50,i=361],u[j=35:50,i=1]
             DATA SET: ./MOM.res.nc
             LATITUDE: 68.2S to 61.5S
             Z (meter): 1
             T (days): 0.25
 Column  1: U[X=60E] is Zonal velocity (m s-1)
 Column  2: U[X=60E(-300)] is Zonal velocity (m s-1)
                     U       U
68.06S / 35:  0.0000000  0.0000000
67.68S / 36:  0.0000000  0.0000000
67.3S  / 37: -0.0026159 -0.0026159
66.91S / 38: -0.0475604 -0.0475604
66.51S / 39: -0.0033971 -0.0033971
66.11S / 40:  0.0259861  0.0259861
65.7S  / 41:  0.0299010  0.0299010
65.29S / 42:  0.0253395  0.0253395
64.87S / 43:  0.0261799  0.0261799
64.44S / 44:  0.0154302  0.0154302
64S    / 45:  0.0051715  0.0051713
63.56S / 46: -0.0034555 -0.0034556
63.11S / 47:  0.0030681  0.0030681
62.66S / 48: -0.0033926 -0.0033925
62.2S  / 49: -0.0070073 -0.0070025
61.73S / 50: -0.0098893 -0.0098852

(j=0:34 were all zeroes)

Will

wfcooke:
[...]

PR #1221, opened 2020-10-06
(*)Add a halo update to fix VERTEX_SHEAR solutions

Hallberg-NOAA:
Added halo update for fluxes%p_surf_full when VERTEX_SHEAR=true.  This field
is not used directly with the vertex shear code, but its halo regions are used
in the frazil code when the diabatic solver is using slightly wider halos to
work with the VERTEX_SHEAR option for shear mixing.  Without this change, some
cases using VERTEX_SHEAR=True were not reproducing across PE layout, whereas now
they do.  There is no flag to recover the old answers because the old answers
were ill-defined.  All answers in the MOM6-examples test suite are bitwise
identical but there will be answer changes in some ice-ocean or coupled cases
with VERTEX_SHEAR=True; these cases should now reproduce across PE layouts.

  This PR should help address MOM6 Issue #850.

### DIFF
--- a/src/core/MOM.F90
+++ b/src/core/MOM.F90
@@ -1250,10 +1250,13 @@ subroutine step_MOM_thermo(CS, G, GV, US, u, v, h, tv, fluxes, dtdia, &
     call apply_oda_tracer_increments(US%T_to_s*dtdia,G,tv,h,CS%odaCS)
   endif
 
-  if (associated(fluxes%p_surf)) then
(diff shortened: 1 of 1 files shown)


adcroft (approved):

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11388

Note: parameters out of date due to pr/1217; This PR does not affect parameters so will be handled later.

Issue #1218, opened 2020-09-29
Sponge setup not working


hmkhatri:
I am trying to set up an idealized channel configuration with a sponge in a fixed latitudinal band near the northern boundary. I am using `SPONGE_CONFIG = "file"` and supplying tracer profiles and `idamp` values from input files. MOM6 reads the data fine but the sponge does not seem to work at all. Temperature and salinity profiles drift from the specified tracer profiles in the sponge region (see picture below). I have even tried with a damping time scale of 1 day in a 10 deg wide sponge but the solution does not change.

Also, checked that MOM6 correctly sets a nonzero damping rate in the sponge region. Here is the related entry from `MOM_parameter_doc.all`
`!Total sponge columns at h points = 15680 ! The total number of columns where sponges are applied at h points.`

Any idea how to resolve this issue? 

![sponge](https://user-images.githubusercontent.com/17830036/94615577-fd5b9980-0275-11eb-9c40-68f1b481854f.png)

gustavo-marques:
@hmkhatri, I believe @sdbachman is successfully using SPONGE_CONFIG = "file". Are your runs on Gaea? If so, please send me the path via Slack and I will compare it against what we have.

MJHarrison-GFDL:
@hmkhatri . Please paste your MOM_parameter.short file. The sponge code has not been well tested and there are a myriad of parameter choices. I fixed a bug in #1210 but it depends on your  settings.

gustavo-marques:
@hmkhatri, I could not find anything wrong with your parameter choices. I noticed you are restoring T and S to the initial state, using a previously generated MOM_IC file. In this file, you only have data for Time = 0. Perhaps you need to add a second-time entry (e.g., Time = 3650) and repeat the T and S fields.  Something like this:

netcdf sponge_TS {
dimensions:
	lont = 320;
	latt = 272 ;
	depth = 75 ;
	time = UNLIMITED ; // (2 currently)
variables:
	double lont(lont) ;
	double latt(latt) ;
	double depth(depth) ;
	double time(time) ;
		time:calendar = "noleap" ;
		time:units = "hours since 0001-01-01 00:00:00" ;
	double PTMP(time, depth, latt, lont) ;
		PTMP:_FillValue = -1.e+20 ;
	double SALT(time, depth, latt, lont) ;
		SALT:_FillValue = -1.e+20 ;
}

Hope that fixes the problem.

hmkhatri:
[...]

PR #1222, opened 2020-10-08
Sponge file init patch

MJHarrison-GFDL:
This addresses #1218.

### DIFF
--- a/src/initialization/MOM_state_initialization.F90
+++ b/src/initialization/MOM_state_initialization.F90
@@ -1731,6 +1731,7 @@ subroutine initialize_sponges_file(G, GV, US, use_temperature, tv, param_file, L
     tmp, tmp2 ! A temporary array for tracers.
   real, dimension (SZI_(G),SZJ_(G)) :: &
     tmp_2d ! A temporary array for tracers.
+  real, allocatable, dimension(:,:,:) :: tmp_tr ! A temporary array for reading sponge fields
 
   real :: Idamp(SZI_(G),SZJ_(G))    ! The inverse damping rate [T-1 ~> s-1].
   real :: pres(SZI_(G))     ! An array of the reference pressure [R L2 T-2 ~> Pa].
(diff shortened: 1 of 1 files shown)


src/initialization/MOM_state_initialization.F90
```diff
@@ -1821,73 +1826,82 @@ subroutine initialize_sponges_file(G, GV, US, use_temperature, tv, param_file, C
...
+      deallocate(h)
+    else
+      ! Initialize sponges without supplying sponge grid
+      call initialize_ALE_sponge(Idamp, G, param_file, ALE_CSp)
+    endif
+
+
+    ! The remaining calls to set_up_sponge_field can be in any order.
+    if ( use_temperature) then
+      call set_up_ALE_sponge_field(filename, potemp_var, Time, G, GV, US, tv%T, ALE_CSp)
```
Hallberg-NOAA, line 1898: set_up_ALE_sponge_field is an overloaded interface to two different routines. I would have thought that the two different interfaces should have been used for the two settings of time_space_interp_sponge, but both options are using the variant that should be used with time_space_interp_sponge=.true.

  Given that this code was not working before, in a way that our testing was not detecting, I think that we need to be particularly careful in implementing these corrections. As a confidence-building measure, I would prefer that this be implemented in the simpler way, using the static target value interface when time_space_interp_sponge=.false. Once this has been demonstrated to work (e.g., the MOM6-issue that triggered all this has been closed), we could then try using the same interface for both variants and verifying that they give identical answers in test cases that exercise that option.

MJHarrison-GFDL:
@Hallberg-NOAA That sounds like a good plan. I will modify the tc4 test accordingly.

src/initialization/MOM_state_initialization.F90
MJHarrison-GFDL: @Hallberg-NOAA . Updated as suggested.

Hallberg-NOAA: I am now satisfied that with these latest modifications this PR is very likely to be exactly what we need to address and close Issue #1218.

Hallberg-NOAA (approved):
I am now satisfied that these are the right changes to make to address Issue #1218, and I am approving this PR.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11391

PR #1223, opened 2020-10-15
Change bounding of Kv_BBL and bbl_thick in set_viscous_BBL()

adcroft:
The majority of this PR is new documentation, in the form of both code comments and equations in doxygen-generated material, with a small but optionally answer changing fix in the last commit.

- The BBL viscosity used to be simply bounded to be no less than a minimum, KV_BBL_MIN. In some limits where the BBL thickness was smaller than the bottom layer thickness, this viscosity led to too much drag on the bottom layer.
- If CORRECT_BBL_BOUNDS=True we now also adjust bbl_thick when bounding Kv_bbl so that the implied stress, viscosity and thickness are all consistent.
- By default, CORRECT_BBL_BOUNDS=False so that existing answers are unchanged.
- This introduces a new run-time parameter so changes the doc strings in MOM-examples.

First commit:
- Added lots of comments explaining method/equations
- Small refactor of code to reduce duplicated code a smidgen but really in preparation for a solution-changing fix coming later
- Also removed unused diagnostic ids in MOM_vert_friction.F90

Second commit:
- Wrote out equations being solve with reference to model parameters. Includes references.

Third commit:
- Switched to non-English spelling of bottom-most

Fourth commit:
- The actual code correction.

### DIFF
--- a/src/parameterizations/vertical/MOM_set_viscosity.F90
+++ b/src/parameterizations/vertical/MOM_set_viscosity.F90
@@ -60,7 +60,7 @@ type, public :: set_visc_CS ; private
   logical :: bottomdraglaw  !< If true, the  bottom stress is calculated with a
                             !! drag law c_drag*|u|*u. The velocity magnitude
                             !! may be an assumed value or it may be based on the
-                            !! actual velocity in the bottommost HBBL, depending
+                            !! actual velocity in the bottom most HBBL, depending
                             !! on whether linear_drag is true.
   logical :: BBL_use_EOS    !< If true, use the equation of state in determining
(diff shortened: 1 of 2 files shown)


Hallberg-NOAA:
The pipeline testing for this PR has been triggered at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11392.

Hallberg-NOAA (approved):
Now that the Queen's English spelling of bottommost has been changed back to the proper American English spelling, and the pipeline tests have passed (while noting the new entries in the MOM_parameter_doc files), this PR is acceptable.

PR #1225, opened 2020-10-19
Rob's documentation framework

kshedstrom:
See the details/Details.pm file.


kshedstrom:
Travis is failing on the doxygen tests. They are just warnings and we can tell it to ignore warnings. Is that what you want? Some warnings are from latex equation bits in figure captions which are getting fixed up in a post-processing step. Some are warnings from generic tracers and the DA function arguments not being sufficiently documented. I guess doxygen doesn't like just "Unknown".

marshallward:
I'm unsure how to handle the warnings, does doxygen return a nonzero error code on a warning?

This may be a genuine error that needs to be looks into:
```
error: Problems running bibtex. Verify that the command 'perl --version' works from the command line. Exit code: 2
```

kshedstrom:
Yes, Rob is working on another update to fix that and the missing TeX requirement. Stay tuned...

jr3cermak:
We will need to add the pre-built binary so we are using 1.8.19 version of doxygen.   The pre-built binary (bin/doxygen) also helps RTD to succeed.  There are not enough compiler bits on RTD to allow building of doxygen on their VM.  Providing a pre-built binary works for travis and RTD.   Once that is included with the PR, that should leave just a handful of warning to fix in the documentation and code.

The handful of warnings to review are:
```
/home/travis/build/jr3cermak/MOM6/src/core/_Barotropic_Baroclinic_Coupling.dox:77: warning: \cite command has no label
/home/travis/build/jr3cermak/MOM6/src/core/_Discrete_PG.dox:19: warning: \cite command has no label
/home/travis/build/jr3cermak/MOM6/src/parameterizations/vertical/MOM_CVMix_KPP.F90:1523: warning: end of file with unbalanced grouping commands
Tracer_Fluxes:-1: warning: multiple use of section label 'Tracer_Fluxes', (first occurrence: /home/travis/build/jr3cermak/MOM6/src/tracer/_Tracer_fluxes.dox, line 5)
/home/travis/build/jr3cermak/MOM6/src/core/_Discrete_PG.dox:20: warning: unable to resolve reference to 'hallberg1997' for \cite command
/home/travis/build/jr3cermak/MOM6/src/tracer/_Tracer_Transport.dox:40: warning: unable to resolve reference to 'easter1993' for \ref command
/home/travis/build/jr3cermak/MOM6/config_src/external/GFDL_ocean_BGC/generic_tracer_utils.F90:41: warning: Member jsd (variable) of type g_tracer_utils::g_tracer_common is not documented.
/home/travis/build/jr3cermak/MOM6/config_src/external/GFDL_ocean_BGC/generic_tracer_utils.F90:25: warning: Member src_var_name (variable) of type g_tracer_utils::g_tracer_type is not documented.
/home/travis/build/jr3cermak/MOM6/config_src/external/GFDL_ocean_BGC/generic_tracer_utils.F90:25: warning: Member src_var_unit (variable) of type g_tracer_utils::g_tracer_type is not documented.
/home/travis/build/jr3cermak/MOM6/config_src/external/GFDL_ocean_BGC/generic_tracer_utils.F90:25: warning: Member src_var_gridspec (variable) of type g_tracer_utils::g_tracer_type is not documented.
/home/travis/build/jr3cermak/MOM6/config_src/external/ODA_hooks/ocean_da_types.F90:59: warning: Member i_index (variable) of type ocean_da_types_mod::ocean_profile_type is not documented.
```

jr3cermak:
The latest travis run is what is currently expected for answers (2541.1).    There is an additional grep expression to filter out warnings associated with math expressions in figure captions.   Math expressions in figure captions are not supported by doxygen.   However, the sphinx infrastructure has been taught to process it and I post process it using postProcessEquations.py.    Expressions render just fine in the html/MathJax.   

We should be able to produce 6 versions of the documentation: APIs (html/pdf); Sphinx (html/pdf) and RTD (html/pdf).  Examples: https://recon.lccllc.info/MOM6/

There are some latex errors that prevent a clean build.  The Makefile is also slightly modified to place latex in batchmode to ignore up to 100 errors.   A proposal would be to add this to travis to also pickup additional formatting errors.    Here is a section I added for "nortd" type latex run:

```
nortd_latex: $(DOXYGENBIN) $(BUILDDIR)
        $< Doxyfile_nortd_latex
        make -C $(BUILDDIR)/doxygen_latex LATEX_CMD="pdflatex -interaction=batchmode"
```

We can scan refman.log for `!` (errors) and have travis fail upon detection of these items as well.   MathJax is more forgiving than latex.  If we do not check for errors, what might be working in the html rendering might be causing the pdf rendering to fail silently.

```
Runaway argument?
{P(s, \theta ) + p \] where $A, \lambda $ and $P$ are functions only \ETC.
! File ended while scanning use of \frac .
<inserted text> 
                \par 
l.259 \input{Equation_of_State}
                               
I suspect you have forgotten a `}', causing me
to read past where you wanted me to stop.
I'll try to recover; but if the error is serious,
you'd better type `E' or `X' now and fix your file.

! Missing $ inserted.
<inserted text> 
                $
```

adcroft:
> Travis is failing on the doxygen tests. They are just warnings and we can tell it to ignore warnings. Is that what you want? Some warnings are from latex equation bits in figure captions which are getting fixed up in a post-processing step. Some are warnings from generic tracers and the DA function arguments not being sufficiently documented. I guess doxygen doesn't like just "Unknown".

It took forever to get those generic routines to pass doxygen and I didn't know what any of the args were for. Most are guesses and then I decided to use caution and "Unknown" since clearly wrong is better than a wrong guess. We need @nikizadehgfdl to document the real OBGC interfaces and these can be updated when there are. It is interesting that missing documentation has been revealed with a newer version of doxygen and yet only one line was in the main `src/` body of code. I suspect that's because @Hallberg-NOAA regularly rereads the code. Hopefully not for leisure. :)

adcroft:
> ... They are just warnings and we can tell it to ignore warnings. Is that what you want? ...

We intentionally fail on warnings. The point of the Travis-CI test is stop undocumented code from getting onto the main branch. The reason for a filter is that some code does not belong to us to document. Even that will change eventually.

adcroft:
> We will need to add the pre-built binary so we are using 1.8.19 version of doxygen. The pre-built binary (bin/doxygen) also helps RTD to succeed. There are not enough compiler bits on RTD to allow building of doxygen on their VM. Providing a pre-built binary works for travis and RTD. Once that is included with the PR, that should leave just a handful of warning to fix in the documentation and code.

Unfortunately, this is a no no. I see why this solved some problems but this is in violation of NOAA software/security policies. That's one reason I was compiling in place. Ideally we should be able to build the documentation with the versions available which does currently work. I vaguely recall something about sphinx versions so maybe that's what this is about? If it is really impossible to use the available versions then we should talk. Another objection: adding binaries of this size dramatically increased the size of the repo (~23%).

adcroft (changes requested):
There is huge amount of work here and it represents a substantial improvement over the current documentation. Although the overall documentation still has gaps I think we should get this PR to the point of passing, and live at the mom6.readthedocs.io .

Comments:
- Although the RTD pages look sensible the `make nortd` pages have a lot of random high level pages in the left side bar. For example, there are pages for "mdl" and "ntr" at the level of "Diagnostics". While we can update RTD I don't think we should update gh-pages until we've tidied up some of this.
- I was able to build both the vanilla doxygen documentation and sphinx based documentation in a clean linux VM with relative ease by following the README. Kudos.
- Unfortunately I have not been able get things to work quite so easily on any of GFDL's systems. This seems to be a doxygen issue that was introduced in 1.8.17 . I was able to get the `make nortd` to work using 1.8.16. The `autodoc_doxygen` extension dies when processing `src/user/adjustment_initialization.F90` and I haven't figured out why. Side issue: The fragility of our documentation build process is a concern - we need users (including those at GFDL) to be able to develop the documentation without resorting to pushing their code to GitHub. Containers etc solve many problems but not for the developers we're targeting. This is something we'll have to address but probably not in this PR.
- I could not find a doxygen log file for `make html`. Anyone know where it is?

Changes needed:
- We obviously need the Travis-CI test to be passing otherwise we'll have a broken main branch. Most of the warnings are real (see comment elsewhere). It is unfortunate that they are only obtained with a recent `doxygen`. As mentioned in the conversation, the point of the Travis-CI test is to stop adding undocumented code. I have fixed the majority of those warnings in #1231 which you can cherry-pick.
- Please resize the images in `images/` to be at most 800px across. Smaller is fine. Some are unnecessarily large (>1Mb) and many don't fit on a standard width web page.
- As mentioned in my last comment, we simply can not ship the binary executable `bin/doxygen`. It violates NOAA policies and also significantly increased the size of the repo. For the latter reason it would be best to rewrite the history when removing the binary.

kshedstrom:
I don't know how to get that binary out of the history on this PR. Is that possible? If I reset my branch at home, I can't push to github because github says I'm behind. We've got some work to do to pass Travis without the binary, so is it best to close this PR and this branch and start over?

jr3cermak:
For travis, we can utilize the compile in-place option to use the latest doxygen.  For RTD, I recall it being the default doxygen binary on their VM producing insufficient XML output for sphinx.  I will check again and lean on the `flint` module as needed since we at least have control over python code on RTD.

I have redirected all the log files to the `_build` directory and renamed them to `*_log.txt`.  More websites are tuned now to show `*.log` files as text.  Unfortunately, if you point a web browser at a `*.log` file via `file:///` on your local drive, it tends to trigger a download.  Having them named `*.txt` allows in-browser viewing.  `make html` will place doxygen warnings in `_build/doxygen_warn_rtd_log.txt`.    

Will continue to refine the README.md.  Verbose information will be linked out of the README.md and put into the `details/*` to try and keep the README.md lean.   I will be sure to add more information about where log files end up.

Will look into the "high level pages".  That is weird.  Do you have a screen grab?  My copy does not show this for 1.8.19.   See: https://recon.lccllc.info/MOM6/esmg/APIs/Diagnostics.html

Dox1819

I would recommend trying the latest doxygen or at least 1.8.20? There was mention of: [issue #7954: The Doxygen uses too much memory (or has probably a memory leak)](https://www.doxygen.nl/manual/changelog.html).  I would try `make nortd` on at least 1.8.20 to see if it gets you past the failure of processing `src/user/adjustment_initialization.F90`.

adcroft:
> I don't know how to get that binary out of the history on this PR. Is that possible? If I reset my branch at home, I can't push to github because github says I'm behind. We've got some work to do to pass Travis without the binary, so is it best to close this PR and this branch and start over?

Let's fix the other stuff first. I can walk you through how to remove the binary. It basically involves creating a new branch and cherry-picking everything except the binary. Let's not worry about that now until we get the PR passing and then once it's ready we can re-submit a PR without the binary.

adcroft:
> For travis, we can utilize the compile in-place option to use the latest doxygen. For RTD, I recall it being the default doxygen binary on their VM producing insufficient XML output for sphinx. I will check again and lean on the `flint` module as needed since we at least have control over python code on RTD.

I think we should try to use the same version on Travis as what ever we end up using on RTD since the test is whether the documentation works.

> I have redirected all the log files to the `_build` directory and renamed them to `*_log.txt`. More websites are tuned now to show `*.log` files as text. Unfortunately, if you point a web browser at a `*.log` file via `file:///` on your local drive, it tends to trigger a download. Having them named `*.txt` allows in-browser viewing. `make html` will place doxygen warnings in `_build/doxygen_warn_rtd_log.txt`.

+1

> Will look into the "high level pages". That is weird. Do you have a screen grab? My copy does not show this for 1.8.19. See: https://recon.lccllc.info/MOM6/esmg/APIs/Diagnostics.html
>
> I see the mdl and ntr at the "same level" now for a git compiled (1.9.0) version, will check others.
> 
> v1.8.19: the first level menu is clean of mdl and ntr, etc. Yes, there is fragility between versions!
> 
> For v1.8.16, I get a clean menu as I do with v1.8.19.

I've found in the past that when doxygen changes output it's often because of ambiguous or incorrect formatting. The fixes I put in #1231 are an example where the documentation was wrong but warnings only appeared with the recent versions. I'll first suspect there's something odd about the formatting of documentation in those two modules before suggesting we need to work with specific versions of doxygen.

jr3cermak:
> Let's fix the other stuff first. I can walk you through how to remove the binary. It basically involves creating a new branch and cherry-picking everything except the binary. Let's not worry about that now until we get the PR passing and then once it's ready we can re-submit a PR without the binary.

That is essentially what we have done, we rewound the branch to before the insertion of the binary and created new branch and pulling items in above that.  Good.  We can submit a new PR with the fixed branch when ready.   We can continue testing on the current PR.

> I think we should try to use the same version on Travis as what ever we end up using on RTD since the test is whether the documentation works.

This makes sense to me.   Essentially, we will support 1.8.13 forward.  I can systematically test versions going forward and see if I can shore up the fragility between versions.   This will also help guard against a surprise RTD infrastructure update.

I expanded the compile in place option in the Makefile to allow specification of the tag/branch.   I think the native doxygen binary on travis is different than 1.8.13.  We can just specify Release_1_8_13 now and should less likely to be bitten by a travis infrastructure update.

Makefile:
```
# You can set these variables from the command line.
SPHINXOPTS       ?=
SPHINXBUILD       = sphinx-build
PAPER            ?=
BUILDDIR         ?= _build
DOXYGEN_BIN      ?= doxygen
DOXYGEN_CONF     ?=
FINDDOXYGENBIN    = $(DOXYGEN_BIN)
DOXYGEN_RELEASE  ?= Release_1_8_19

doxygen:
	git clone -b $(DOXYGEN_RELEASE) --depth 1 https://github.com/doxygen/doxygen
```

Just down to two things before the new PR: Re-test on version 1.8.13 and moving post processing of equation renumbering into sphinx.  On RTD, they run directly using sphinx-build and do not use the Makefile.  It will not change the command line.  Renumbering will be triggered in sphinx rather than the Makefile.

kshedstrom:
Replaced by #1235 .

PR #1211, opened 2020-09-21
Candidate for main from GFDL 2020-09-18

adcroft:
This is a PR onto our collective main branch. Note we now use the terminology "main" where we once used "dev/master". Instructions on renaming your branches can be found at https://github.com/NOAA-GFDL/MOM6/wiki/Renaming-the-main-branch-in-MOM6.

This PR has some potentially major-impact changes which are detailed below. Each in isolation made sense but unfortunately we let them stack up so may take some work to sort through - our apologies! 

- A change in defaults for parameters potentially changes your answers if you implicitly rely on a parameter by not setting it explicitly at run-time. If your answers have changed with this PR-to-main then follow the instructions outlined in NOAA-GFDL#1127 to recover your answers.
  - NOAA-GFDL#1127  MOM6: +(*)Changed the defaults for 36 parameters
- Reorganization of some source code. The use of git submodules for optional code in `pkg/` has been replaced by the use of null modules in `config_src/external`. This means the search path for code at compile time must be updated.
  - NOAA-GFDL#1133  Added null version of ocean_BGC
  - NOAA-GFDL#1139  Moved hooks from pkg to external
  - NOAA-GFDL#1160  Move MOM_oda_driver to framework directory
- New version of FMS. It's not a major change but you must use tag 2019.01.03 from here on. There is an new argument in one API that is needed to be forward compatible. We will soon be updating GFDL to using a 2020 tag but that will not require source code change and should be optional.
  - NOAA-GFDL#1141  FMS2020.01 "runtime required" but optional argument to diag_axis_init
  - NOAA-GFDL#1143  This is the minimum required update to use 2020.01 FMS & FMScoupler
  - NOAA-GFDL#1158  Test patch to FMS 2019.01.02 from Rusty
  - NOAA-GFDL#1164  Move FMS tag to 2019.01.03
- Change in diagnostic timing. This will change the values you diagnose for temperatire and salinity. Due to a mistake introduced when closing the remapped diagnostic budgets the call to posts for T/S was moved too early in the algorithm.
  - NOAA-GFDL#1207  Post tracer concentration when synced

I've classified the remaining PRs below which we expect should not affect you adversely.

As usual, please can @jiandewang, @awallcraft, @kshedstrom, and @gustavo-marques review the PR.

Bug fixes:
- NOAA-GFDL#1114  Rotate: fixed-time ALE sponge bugfixes
- NOAA-GFDL#1122  Solution for #1121               (OBC)
[...]

PR #1227, opened 2020-10-19
Change default for USE_NET_FW_ADJUSTMENT_SIGN_BUG in drivers

adcroft:
This commit has already been added to main-candidate-2020-09-18 #1211 but needs merging into dev/gfdl

- The default for USE_NET_FW_ADJUSTMENT_SIGN_BUG had been changed to
  False but during the merge of main onto dev/gfdl the default was
  inadvertently flipped. @gustavo-marques pointed this out when reviewing
  NOAA-GFDL/MOM6#1211.
- This affects EMC and NCAR versions of drivers and is not tested at GFDL.


Hallberg-NOAA (approved):
I agree that this fixes a problem that I inadvertently introduced, and that this should be pulled into dev/gfdl at the earliest opportunity.  There is no need to run the MOM6-examples regression tests on this case because it will not be exercised.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11421

PR #1229, opened 2020-10-22
Nullify pointer in type neutral_diffusion_CS

adcroft:
- The was causing an error at copmile time with gcc/7.3.0 using
  FCFLAGS_INIT=-finit-real=snan -finit-integer=2147483647 -finit-derived

### DIFF
--- a/src/tracer/MOM_neutral_diffusion.F90
+++ b/src/tracer/MOM_neutral_diffusion.F90
@@ -97,7 +97,7 @@ type, public :: neutral_diffusion_CS ; private
   integer :: id_uhEff_2d = -1 !< Diagnostic IDs
   integer :: id_vhEff_2d = -1 !< Diagnostic IDs
 
-  type(EOS_type), pointer :: EOS   !< Equation of state parameters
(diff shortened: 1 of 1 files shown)


marshallward:
The error itself makes no sense to me
```
../../../ac/../src/tracer/MOM_neutral_diffusion.F90:9:4:

 use MOM_diag_mediator,         only : diag_ctrl, time_type
    1
Error: The element in the structure constructor at (1), for pointer component 'eos' should be a POINTER or a TARGET
../../../ac/../src/tracer/MOM_neutral_diffusion.F90:9:4:

 use MOM_diag_mediator,         only : diag_ctrl, time_type
    1
Error: Pointer initialization target at (1) must have the SAVE attribute
```
and the fix makes even less sense, but I can confirm that it all works.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11428

Hallberg-NOAA (approved):
This seems like a reasonable change.

PR #1230, opened 2020-10-22
Adds TIME macro to .testing/Makefile

adcroft:
This simply provides a means to avoid out-of-order output when running `make test -j` in parallel. By default we can get output like this:
```
PASS: Diagnostics tc4.dim.q.diag agree.
PASS: Diagnostics tc4.openmp.diag agree.
PASS: Diagnostics tc4.dim.r.diag agree.

real	0m16.012s
user	0m0.129s
sys	0m0.268s

real	0m15.761s
user	0m0.116s
sys	0m0.318s

real	0m15.940s
user	0m0.112s
sys	0m0.302s

real	0m15.705s
user	0m0.177s
sys	0m0.251s
DONE: tc2.a.dim.l; no runtime errors.
DONE: tc2.a.nan; no runtime errors.
DONE: tc2.dim.z; no runtime errors.
```

but by adding `TIME =` to config.mk or invoking in bash with `TIME="" make test -j` the times are omitted and output cleaner.

### DIFF
--- a/.testing/Makefile
+++ b/.testing/Makefile
@@ -27,6 +27,8 @@ FCFLAGS_COVERAGE ?=
 # Set to `true` to require identical results from DEBUG and REPRO builds
 DO_REPRO_TESTS ?=
 
+TIME ?= time
+
 # Many compilers (Intel, GCC on ARM64) do not yet produce identical results
 # across DEBUG and REPRO builds (as defined below), so we disable on default.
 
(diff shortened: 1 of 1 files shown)


marshallward:
I almost wonder if `TIME` should be unset on default, partly for the reasons that you outlined (ambiguity of results when run in parallel).

But this is an incrementally positive step and we can disable it later if there is any feedback to do so.

(NOTE: This does not touch code so does not require a regression test.)

marshallward (approved):

PR #1231, opened 2020-10-24
Fixes doxygen warnings mostly from config_src/external

adcroft:
- #1225 is showing lots of errors due to using a newer version of doxygen
- These doxygen mistakes exist in the current dev/gfdl branch and apparently are missed by the older versions we are using on mom6.readthedocs.io and locally.

Todo:
[ ] - Have the OBGC routines properly documented so we can replace the "Unknown" documentation in the generic_tracer stubs.


Hallberg-NOAA:
The pipeline testing for this PR is at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11453 .

Hallberg-NOAA (approved):
I agree that these corrections to the dOxygen comments will avoid syntax errors, although there are a distressingly large number of uninformative"Unknown" descriptions that need to be corrected.

PR #1233, opened 2020-10-27
Testing: Fixes; Improved compiler configuration

marshallward:
This patch contains several minor fixes to the build and testing suites.

- Missing netcdf module and library errors were incorrectly reported as
  missing FMS libraries.  Fixed.

- Individual tc tests now remove output in `results` in order to improve
  the accuracy of `make test.summary` reporting.

The handling of compiler configuration via FC, FCFLAGS, etc has also
been heavily reimplemented to respect Makefile environment variable
passing conventions.

Several hard-coded variables such as FMS_ENV, PATH_ENV, FC_ENV, etc
which acted as means to transmit such information across Makefile and
configure scripts have now been removed, since Makefile is already
capable of passing such information via $(MAKE) invokation.

This has simplified much of the preamble code in the Makefiles, and should
provide more predictable and consistent behavior when using the .testing
and ac/deps Makefiles.

Implicit variable names have also been disabled, which required that we
also handle the $(AR) library archival program in Makefile.in.  This is
currently hard-coded to ar, since autoconf is not capable of validating
this function and It would traditionally be handled by automake or
libtool, but this is equivalent to relying on Make's assumed value of
`ar` and is sufficient for our platforms.

Documentation has been updated to reflect these changes, and also to
provide more expicit instructions for git submodule support and
directory navigation.

Finally, the `head` output from .testing has been extended to -20 lines
in order to provide more meaningful output for some comparisons.


marshallward:
`PATH` is not being correctly configured for the MOM6 jobs, although it is correctly working for FMS.  I am looking into it right now.

marshallward:
The test suite never runs the `ac/deps/Makefile` and so `ac/deps/bin` never exists.  Instead, the test suite uses its own `.testing/deps/bin` mkmf tools.

The previous PR removed the paths to `.testing/deps/bin`, incorrectly assuming they would be in `ac/deps/bin`.  This update restores the `PATH_FMS` variable which directs to the local `.testing/deps/bin` mkmf tools.

PR #1234, opened 2020-10-28
(+)Added the option to avoid a bug when using KDML

Hallberg-NOAA:
Added the new runtime option KD_BACKGROUND_VIA_KDML_BUG that can be set to
false to correct a bug in an obscure mode of setting the background diffusivity
using KDML if 6 other higher precedence options are not in use.  This case is
not common, and this new parameter is not being logged if it does not do
anything.  This bug was introduced on March 26, 2018 in MOM6 commit 7a818716,
which was included in MOM6 PR#750.  This fix could change the MOM_parameter_doc
file entries in some cases, but by default all answers are bitwise identical.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11457

marshallward (approved):

PR #1236, opened 2020-10-30
Refactor of diagnoseMLDbyEnergy computation

marshallward:
This patch includes some minor modifications to the MLD calculation in
`diagnoseMLDbyEnergy`, in order to retain consisent numerical results
after optimization.

Discrepancies were observed in the MLD_EN1 diagnostic with -O0 and -O2
with gfortran 10.2.  This patch restores agreement of these results.

Most of the changes are explicitly replacing certain integer
multiplcations as floating point (3 * x => 3. * x), additional
parentheses, and replacement of a cubic operation with explicit
multiplication (x**3 => x*x*x).  Equations were also reformatted for
closer consistency to the style guide.

This may changes answers for the diagnostic for some runs which rely on
the optimized result.


marshallward:
@breichl I can't seem to select you as a reviewer, but you should certainly review this.

breichl (approved):
Looks fine to me.  Thanks for fixing the integer multiplication and improving the style.

Is it a general rule that cubes need be expanded, but squares are left as integer exponents?

Hallberg-NOAA:
Integer cubes are fine, because x*(x*x) = (x*x)*x, but higher powers are a problem because (x*x)*(x*x) /= x*(x*(x*x)). We never want to use a real power in place of an integer power because that could resolve to a completely different and more expensive intrinsic function.

marshallward:
Expansion of exponential is compiler dependent, and an aggressive one could even turn `x*x*x` to `x**3` == `exp(3*log(x))` if it saw an advantage, so I'm not sure we could ever define a rule here..  But generally the higher powers will be converted at higher optimization.

In this case, I did not actually see a change with x**3 and x*x*x (it was the first thing I tried), so it was probably unnecessary, but I left it in.  My guess is the `3*x` -> `3.*x` was the more significant change.

breichl:
Okay, got it.  Thanks for the explanations!

PR #1237, opened 2020-10-30
More docs

kshedstrom:
This is bringing more material over from the esmg-docs branch. It moves the docs/equations files to (mostly) src/core, putting it into .dox files instead of .rst files. It also adds the Coriolis and pressure gradient write-ups, plus MOM6 testing from @marshallward 's talk.


.travis.yml
```diff
@@ -14,7 +14,7 @@ addons:
     - mpich libmpich-dev
     - graphviz flex bison cmake
     - python-numpy python-netcdf4
-    - python3 python3-dev python3-venv python3-pip python3-sphinx
+    - python3 python3-dev python3-venv python3-pip python3-sphinx python3-lxml
```
adcroft, line 17: I'm not sure why this is showing up as a diff since it's already on the dev/gfdl branch. https://github.com/NOAA-GFDL/MOM6/blob/dev/gfdl/.travis.yml#L17 . No matter - it's correct.

adcroft: Clicking "Update branch" solved that.

adcroft:
Any chance you can provide a link to the rendered site? I'm assuming you have it already but if not I can push it to a branch on RTD.

There are three \ref errors that are blocking a merge and so need to be fixed.

kshedstrom:
I believe it is passing now. Sorry, it is only rendering on my laptop, not out on the internets.

kshedstrom:
I've got the rest of the documentation in a clean branch. Do you want it all at once?

adcroft:
The rendered view of this PR can be seen at https://mom6-doctest.readthedocs.io/en/doc-test/.

jr3cermak:
> The rendered view of this PR can be seen at https://mom6-doctest.readthedocs.io/en/doc-test/.

I am slowly working through fixing what broke going back to 1.8.13.   I will send a doc software update PR soon and future renderings should improve.   I am about 60% through testing.

jr3cermak:
Here is the latest private rendering on RTD: https://mom6-devesmg.readthedocs.io/en/latest/.   This is what this PR should produce.

For grins, I ran the current software through the various doxygen versions from 13 to 20 so I can get a sense of what needs to be fixed next.  https://recon.lccllc.info/MOM6/mom6.php

There is some breakage in the html renderings going forward.  The PDF renderings seem to be ok.   There is a bit more work for 18+.    Whitespace cleanup will be problematic as RST likes whitespace between links, targets, etc.

I noticed the description for the API modules wraps nicely on RTD.  The column does not wrap when viewed from outside RTD.  There is a scrollbar at the very bottom of the page.   Not quite useful on a large page.    Could be a quick CSS styling fix.

RTD:
RTD

Outside RTD:
nnonRTD

The descriptions do not wrap nicely in the PDF.   I would be happy to take any latex, css or html recommendations and I can try to fold them into the code somehow.

Module list in PDF:
pdfModules

adcroft (approved):

PR #1238, opened 2020-10-30
+Revised interface to calculate_CVMix_conv

Hallberg-NOAA:
Revised the interface to calculate_CVMix_conv to pass in the diffusivities
and viscosities that are to be augmented.  This allows the contents of
CVMIX_conv_cs to be made private to the MOM_CVMix_conv module.  Also eliminated
permanently allocated arrays used for diagnostics in this module, and use GV
instead of G to set the vertical array sizes.  Also modified calls in
MOM_diabatic_driver to use these new interfaces.  All answers are bitwise
identical, but there are changes to a public interface and a type.


adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11479

PR #1240, opened 2020-10-30
Proposal for interface/submodule framework layer

marshallward:
This is a potential solution for implementing a more flexible layer
between MOM6 and its underlying framework.

Currently such a framework is strongly hard-coded to FMS in the
`src/framework` directory, but this can create challenges when using
MOM6 in other coupled models which do not have the same operational
requirements as GFDL.

This solution replaces FMS-based implementations of framework
subroutines with interfaces, which define the API of the framework layer
but do not explicit define it.

The implementations are provided in a separate directory, such as
`config_src/fms_layer`, which is configured at build time.  These
implementations are provided as submodules to the parent module
containing the interface.

This should not be considered a final solution, but rather a starting
point for discussion.

Advantages of separation:

- MOM6 code becomes framework-agnostic, and `src/framework` can
  predominantly contain interfaces to framework-specific
  implementations.

- This is not constrained by any particular API.  The MOM framework API
  is fixed, but is implementation is free to call any combination of
  framework functions in any way needed to achieve the same result.

- This is extendible to an arbitrary number of frameworks, and their
  behavior and testing would be completely separate from MOM6
  development.

Advantages of submodules:

- Framework interfaces are explicit.  Any implementation would need to
  follow the interface as specified in the MOM6 source code.

- A `libMOM6` could be constructed without the framework implementation,
  which could be contained in its own

- Stronger logical separation of framework modules and the other modules
  which reference them.  This may potentially reduce the build times and
  the potential for build errors due to outdated module files.

### DIFF
--- a/ac/configure.ac
+++ b/ac/configure.ac
@@ -168,11 +168,13 @@ AS_IF([test -z "$MKMF"], [
 
 
 # NOTE: MEM_LAYOUT unneeded if we shift to MOM_memory.h.in template
+# TODO: fms_layer should be configurable (and renamed)
 AC_CONFIG_COMMANDS([path_names],
   [list_paths -l \
     ${srcdir}/src \
     ${srcdir}/config_src/solo_driver \
     ${srcdir}/config_src/ext* \
+    ${srcdir}/config_src/fms_layer \
     ${MEM_LAYOUT}
 ], [MEM_LAYOUT=$MEM_LAYOUT])
 
--- a/config_src/fms_layer/MOM_error_handler_fms.F90
+++ b/config_src/fms_layer/MOM_error_handler_fms.F90
@@ -1,0 +1,30 @@
+!> FMS-based implementation of MOM_error_handler subprograms
+submodule (MOM_error_handler) MOM_error_handler_fms
+
+  ! NOTE: These are currently imported from the MOM_error_handler module.
+  !   If uncommented, this line would cause a namespace conflict.
+  !   Once all references to mpp_error and NOTE are removed from
+  !   MOM_error_module, this line should be re-enabled.
+  !
+  !use mpp_mod, only : mpp_error, NOTE
+
+  implicit none
+contains
+  !> This provides a convenient interface for writing an informative comment.
+  module subroutine MOM_mesg(message, verb, all_print)
+  !module subroutine MOM_mesg(message, verb, all_print)
+    character(len=*), intent(in)  :: message !< A message to write out
+    integer, optional, intent(in) :: verb !< A level of verbosity for this message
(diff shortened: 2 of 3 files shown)


marshallward:
Further thoughts on this:

It appears that a "thin wrapper" - as an alias to the FMS function - cannot be done in the submodule.  For example, if we were to define something like `MOM_write_error` as an alias to `mpp_error`, then this is not possible.
```
submodule (MOM_error_handler) MOM_error_handler_fms
  use mpp_mod, only : MOM_write_error => mpp_error
end submodule MOM_error_handler_fms
```
An explicit wrapper is still possible.
```
submodule (MOM_error_handler) MOM_error_handler_fms
  use mpp_mod, only : mpp_error
contains
  module subroutine MOM_write_error(code, mesg)
    integer, intent(in) :: code             !< A level of verbosity for this message
    character(len=*), intent(in)  :: mesg   !< A message to write out
  
    call mpp_error(code, mesg)
  end subroutine MOM_write_error
end submodule MOM_error_handler_fms
```
Obviously not ideal, but also not necessarily an issue.

My own impression is that if we find ourselves doing this, then we may be defining the operation too narrowly.  We ought to define what we want from a lower-level framework (I/O, diagnostics, restart/pickups, process communication) at the MOM-level.  Then `config_src` should be where we make a best effort to align our proposed behavior to the framework.

For example, `MOM_error_handler` defines a verbosity `level` from 0 to 9, with 0, 1, and 2 largely corresponding to FATAL, WARNING, and NOTE or combinations of them.  These levels should be defined in `src/framework` but translated in `config_src/fms_layer`.  So something like `MOM_error` feels like it belongs in `config_src/fms_layer`.
```
  select case (level)
    case (NOTE)
      if (write_msg.and.verbosity>=2) call mpp_error(NOTE, message)
    case (WARNING)
      if (write_msg.and.verbosity>=1) call mpp_error(WARNING, message)
    case (FATAL)
      if (verbosity>=0) call mpp_error(FATAL, message)
    case default
      call mpp_error(level, message)
  end select
```
though perhaps this preamble, which does not reference FMS, belongs in `src/framework`:
```
  write_msg = is_root_pe()
  if (present(all_print)) write_msg = write_msg .or. all_print
```
But a first pass, I'm not sure it matters where that last part goes.  This would eventually be dictated by the needs of future frameworks.

Extremely MOM-specific content, such as the call tree generation, certainly belongs in `src/framework` and would not be moved to `config_src`.

marshallward:
Closing this, given that submodule support is still rather new and would cause problems on older compilers.  I think we should revisit this in a few years after a tender cycle or so.

PR #1241, opened 2020-11-03
Testing: results cleanup fix

marshallward:
Fixing another minor issue related to wiping of empty test directories
in `results`.

1. std.{out,err} logs are now wiped prior to running the test, rather
   than wiping them prior to the

   This might be logically equivalent, but its better to wipe them in
   the rule which creates the new ones, and seems to avoid additional
   edge cases.

2. $(shell ls -A ${dir}) needs to be quoted in order for the shell to
   test the results, which are now added.


adcroft (approved):
Erroneous warnings are cleared up as a bonus.

PR #1242, opened 2020-11-03
Adds .testing jobs to the gitlab pipeline

adcroft:
- This adds the .testing tests, that until now have only run on Travis-CI,
  to the gitlab pipeline.
- Travis-CI only uses the gnu compiler (7.5.0). Adding these tests to gaea
  uses both gcc/7.3.0 and intel/18.0.6.288 which are the versions used to
  define the regression results for the MOM6-examples suite.
- The gnu test (compile and run) completes in under 5 minutes and the intel
  test in 10 minutes. I've added these in parallel to the "run" stage which
  takes >20 minutes so in all this has no impact of total turn around time.
- We can consider adding other available compilers and versions since these
  are relatively quick and cheap.

image

Example of full test is at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11484

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -122,6 +122,30 @@ run:
     - test -f restart_results_gnu.tar.gz
     - time tar zvcf $CACHE_DIR/results-$CI_PIPELINE_ID.tgz *.tar.gz
 
+gnu.testing:
+  stage: run
+  tags:
+    - ncrc4
+  script:
+    - cd .testing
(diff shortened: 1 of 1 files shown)


marshallward:
Once more for good measure: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11485

I had thought that `.testing` did not yet pass under Intel compilers, so this is good news.

marshallward (approved):

PR #1244, opened 2020-11-09
*Correct indexing bug in ALE_sponge

MJHarrison-GFDL:
- Answer change for tc4 or any configurations
   which use the sponge option (tracer nudging to data)
   in ALE mode with USE_TEMPERATURE = True.


marshallward (approved):
Correctly replaces index bound with the index.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11511

marshallward:
This has passed Gaea regression, although it does not pass our .testing regression.  This is most likely because tc4 touches code which the regressions do not.

As this is a bugfix, I am merging this one.

PR #1245, opened 2020-11-09
Testing: Reduce output for diagnostic errors

marshallward:
This patch fixes an issue with excessive output from the new diagnostic
test.

Previously, the `grep` based output used to detect a new diagnostic was
send to stdout, which produced enormous amounts of output.  This patch
moves that output to /dev/null.  It also trims the output from any
diagnostic regressions to the first 20 lines.

This patch also replaces a few unnecessarily recursive deletes (rm -rf)
with non-recursive delets (rm -f).


adcroft (approved):
Brilliant!

PR #1248, opened 2020-11-09
Nullify pointer in type neutral_diffusion_CS

adcroft:
- The was causing an error at copmile time with gcc/7.3.0 using FCFLAGS_INIT=-finit-real=snan -finit-integer=2147483647 -finit-derived
- Note this fix was already submitted in #1229 but inadvertently undone in a later merge.


marshallward:
Undone in #1235 with commit c3d0cc3a72

Two reasons this creeped in:
- The PRs were too large and unreviewable to this detail
- We are not testing GCC 7 anywhere.

Both of these should be resolved at some point.

marshallward (approved):

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11512

Issue #683, opened 2018-01-09
Channels not being implemented and no ERROR


wfcooke:
When using the 
 MOM6-examples/ice_ocean_SIS2/OM4_025/INPUT/MOM_channels_global_025
file to create channels in SPEAR (1 deg coupled ocean) we are finding that the channels are not being implemented.
The problem occurs at 
https://github.com/NOAA-GFDL/MOM6/blob/85c2ea0bf81b12a316fe40bc7713520f14a558fb/src/initialization/MOM_shared_initialization.F90#L907

and L926 where the input values are compared to the model values. If the input limits fall within a gridbox then the code does not create the limiting channel **but does not report that it is not being created**.

e.g.
U_width,  -5.9,  -5.4, 35.8, 36.2,  12000.0  ! Gibraltar
from ocean_static.nc
geolat_u  34.67->35.49->36.30 (in SPEAR)
geolon_u -6 -> -5 -> -4

-5.9, -5.4 falls between -6 and -5 
 35.8, 36.2, falls between 35.49->36.30

so we end up with a full width gridbox for the Gibraltar channel.

All 5 channels are not set for our setup

SUGGESTED FIX : 
Add a FATAL error to bring the model down if the channel is not implemented.
(and then update the MOM_channel  file for your model)

Hallberg-NOAA:
This issue has been addressed by the improved channel list error handling in PR#1251.

PR #1251, opened 2020-11-18
+Better error handling in reset_face_lengths_list()

Hallberg-NOAA:
Added error handling to reset_face_lengths_list() to note (and possibly trigger
a fatal error) for any entries in the CHANNEL_LIST_FILE that do not cause any
open face lengths to change. The runtime parameter FATAL_UNUSED_CHANNEL_WIDTHS
determines whether only warnings are issued or whether there should be a fatal
error; the default is false because otherwise a number of the Baltic test cases
that deliberately share files with global cases would fail due to channels
outside of their domains.  This PR addresses MOM6 issue #683, which should be
closed once this PR is merged in, although by default this new code triggers
warnings and not the suggested fatal error.  All answers are bitwise identical,
but there are new entries in some MOM_parameter_doc files.


src/initialization/MOM_shared_initialization.F90
```diff
@@ -889,16 +898,19 @@ subroutine reset_face_lengths_list(G, param_file, US)
   call broadcast(num_lines, root_PE())
   u_pt = 0 ; v_pt = 0
   if (num_lines > 0) then
-    allocate (lines(num_lines))
-    if (num_lines > 0) then
```
adcroft, line 893: Bizarre!

adcroft (approved):

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11546

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11547

PR #1252, opened 2020-11-19
Fix dimensions in wave structure

raphaeldussin:
routine was failing dimensional testing. Proposed fixes makes it dimensionally consistent.
I also add several dimension checks, available with DEBUG=true

I think the normalization of the vertical structure was wrong and I corrected accordingly.

In spite of the various fixes, tridiagonal solver becomes unstable at step=2 in benchmark :(


raphaeldussin:
@marshallward how do I figure out where the style problem comes from?

adcroft:
> @marshallward how do I figure out where the style problem comes from?

Evidently I messed up the error reporting when merging the two workflows. It was easier when they were separate. I can fix the reporting once the workshop is over...

raphaeldussin:
@adcroft it was a dumb whitespace error, I fixed it. It would be helpful though to have github actions print the styling errors found

adcroft:
> @adcroft it was a dumb whitespace error, I fixed it. It would be helpful though to have github actions print the styling errors found

Yes, and it used to. While developing the workflows I initially had separate workflows for style and docs but we decided we needed fewer workflows. In order to avoid a white space problem not block the testing of docs I had to do a workaround. I got it wrong.

src/diagnostics/MOM_wave_structure.F90
```diff
@@ -195,7 +197,13 @@ subroutine wave_structure(h, tv, G, GV, US, cn, ModeNum, freq, CS, En, full_halo
   Pi = (4.0*atan(1.0))
 
   S => tv%S ; T => tv%T
-  g_Rho0 = US%L_T_to_m_s**2 * GV%g_Earth / GV%Rho0
+  g_Rho0 = GV%g_Earth / GV%Rho0
+
+  if (CS%debug) call chksum0(g_Rho0, "g/rho0 in wave struct", &
+                             scale=US%Z_to_m*(US%s_to_T**2)*US%kg_m3_to_R)
```
Hallberg-NOAA, line 203: I suspect that the scaling here is not quite right, and that it requires an additional factor of US%L_to_Z**2.  See the documentation of the rescaled units for GV%g_Earth in MOM_vertical_grid.F90.  Also, once this is working we may want to comment out this checksum for a value that is the same for each call, or only do the checksum the first time this routine is called.

Hallberg-NOAA (changes requested):
Overall, I think that there is progress with these changes, and suspect that they will all be found to be correct when used with an appropriate test case. However, it appears that there are new definitions for the rescaled units of a number of the variables in this routine, but there have not been any changes to the comments describing these variables to reflect these changes.

  Please add comments describing the (now correct) units of the variables when the variables are defined.  I am aware that some of the variables are not described in comments at all, but this should be the right opportunity to start doing so.

raphaeldussin:
I don't know how to fix style if I can't see the errors in the checks

marshallward:
Looks like one of your lines is exceeding the 120-character limit due to the comment:
```
$ ./.testing/trailer.py -e TEOS10 -l 120 src config_src | tee style_errors
src/diagnostics/MOM_wave_structure.F90, line 166: Line length exceeded "  real, dimension(SZK_(G)+1) :: Uavg_profile !< Vertical profile of the magnitude of horizontal velocity [L T-1 ~> m s-1]."
```

We need to clean this up to give better feedback, I think.

raphaeldussin:
@marshallward this would be very helpful thanks!

raphaeldussin:
I think this is ready but I'd like to give @sonyalegg a chance to chime in

adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11699

Hallberg-NOAA (approved):

PR #1257, opened 2020-11-27
(*)Use hvel_shelf in vertvisc_coef with ice shelves

Hallberg-NOAA:
Explicitly set hvel_shelf in all cases that have shelf points, to avoid using
uninitialized arrays when setting CS%h_u and CS%h_v in some cases, and also to
simplify the code by reducing the number of calls to find_coupling_coef.  All
answers in MOM6-examples test cases are bitwise identical, and it seems probable
that if there are cases where answers would change that these might fail due to
the use of uninitialized variables.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11611

marshallward (approved):

Issue #1224, opened 2020-10-16
The MOM6 available_diags files have become excessively large


Hallberg-NOAA:
The available_diags files document all of the available diagnostics, and they are intended to facilitate the identification of diagnostics that a user might wish to save with a particular configuration, and to prevent the addition of duplicative diagnostics. However, because MOM6 now generates a large number of diagnostics that are derived from the original diagnostic, such as spatial averages or diagnostics on alternate vertical grids, the number of entries in these files is many times larger than the number of root diagnostics.

  For example, the MOM6 available_diags file for the OM4_025 test case is 644Kb, and has 3651 entries, running for 14,953 lines.   Because this case uses two possible horizontal grids, 3 vertical coordinates for diagnostics, and spatially averaged versions of many diagnostics, there 12 entries for many diagnostics.  With plans for even greater flexibility in diagnostics, this inflation factor could grow much larger in the future.

  To remedy this file-size bloat and to make these available_diags files useful once again for selecting diagnostics, I am proposing that we add a run-time option to indicate that only the primary diagnostic should be logged in this file, perhaps with an additional entry in this file indicating the alternative vertical coordinates, horizontal grids, or other automatically derived variants (e.g., spatial averages) are available for each diagnostic.

StephenGriffies:
Sounds like a very good idea @Hallberg-NOAA .

PR #1258, opened 2020-11-29
+Add option to write shorter available diags files

Hallberg-NOAA:
Added the option to write out available diags files that avoid having a huge
number of redundant entries related to alternative diagnostic grids or module
names and derived quantities like spatial averages.  In this case, these
alternative variants are documented in the available diags files with a brief
'variants' line.  A new runtime parameter, FULL_AVAILABLE_DIAGS_LIST, was added
to enable this new behavior; by default the available diags files look like they
did previously, but this might be the exception to the rule where code changes
should not change output by default.  Also cleaned up some non-standard code
style (like white-space) in MOM_diag_mediator.  All answers are bitwise
identical, but there is a new entry in the MOM_parameter_doc.all files.

  Discussion on the right default for FULL_AVAILABLE_DIAGS_LIST would be helpful.

  This PR addresses MOM6 issue #1224, which might be closed after this PR is accepted.

### DIFF
--- a/src/framework/MOM_diag_mediator.F90
+++ b/src/framework/MOM_diag_mediator.F90
@@ -242,7 +242,10 @@ type, public :: diag_ctrl
                                           !! This file is open if available_diag_doc_unit is > 0.
   integer :: chksum_iounit = -1           !< The unit number of a diagnostic documentation file.
                                           !! This file is open if available_diag_doc_unit is > 0.
-  logical :: diag_as_chksum !< If true, log chksums in a text file instead of posting diagnostics
+  logical :: full_diag_list = .true.      !< If true, write an entry for every variant of a
+                                          !! diagnostic to the available_diags file; otherwise only
+                                          !! write an entry for the primary diagnostics.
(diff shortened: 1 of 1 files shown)


adcroft:
- Split the list of modules and vars (currently in variants)
- Expand out list of options rather than introduce a new syntax

adcroft:
I think you can drop the option to keep old format - not sure anyone would want to maintain the old files.

Hallberg-NOAA:
This PR was just updated to always use the new, shorter available_diags format, as suggested in reviews.

  Revised the new available_diags format in response to reviews, including
explicitly listing all variable and module names, and always writing the
available_diags file in the new shortened format.  This includes removing the
recently added runtime parameter FULL_AVAILABLE_DIAGS_LIST.  All answers are
bitwise identical, but there are substantial changes in the available_diags
files, and the MOM_parameter_doc files no longer are changed by this PR.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11615

PR #1264, opened 2020-11-30
Use flush function instead of subroutine/extension

### LINKED ISSUE #1263: Convert from flush subroutine to intrinsic statement
As mentioned in https://github.com/NOAA-GFDL/MOM6/pull/1165#issuecomment-733916634, we (the GEOS developers) found that when using Intel Fortran on macOS that, for some reason, Intel was getting confused by `call FLUSH()` calls. This is probably because the `call FLUSH()` subroutine is a Fortran extension. So we've been internally in GEOS and in our forks, converting all the `call FLUSH()` to use the Fortran Standard `FLUSH()` intrinsic (which I think every good Fortran compiler now should support) and that seems to be helping.

mathomp4:
This fixes #1263. It converts the `call FLUSH()` in MOM6 to be `flush()`

### DIFF
--- a/config_src/solo_driver/MOM_driver.F90
+++ b/config_src/solo_driver/MOM_driver.F90
@@ -257,7 +257,7 @@ program MOM_main
 !$  call omp_set_num_threads(ocean_nthreads)
 !$OMP PARALLEL
 !$  write(6,*) "ocean_solo OMPthreading ", fms_affinity_get(), omp_get_thread_num(), omp_get_num_threads()
-!$  call flush(6)
+!$  flush(6)
 !$OMP END PARALLEL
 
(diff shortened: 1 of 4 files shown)


mathomp4:
Huh. I am not sure how I broke Doxygen. I'm kind of impressed that this could. 😄

marshallward:
@mathomp4 That was some sort of Ubuntu packaging issue (libpoppler security problem?).  Anyway seems fixed now.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11621

marshallward:
Thanks, this does not appear to be causing any problems so I will merge it.

marshallward (approved):

PR #1269, opened 2020-12-06
Avoid warnings to stdout when closing  MOM_param_file for the ice shelf.

MJHarrison-GFDL:
- added an optional argument to MOM_file_parser:close_param_file
     which will override settings for handling unused parameters

### DIFF
--- a/.testing/tc4/MOM_input
+++ b/.testing/tc4/MOM_input
@@ -1,27 +1,27 @@
 ! This file was written by the model and records the non-default parameters used at run-time.
 
 ! === module MOM ===
-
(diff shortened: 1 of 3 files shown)


marshallward:
This is a good solution, but I thought we had privately discussed adding a `MOM_IS_input` input file instead?

If we had such a file, then it ought to not be necessary to introduce a flag to suppress unread parameter lines.

MJHarrison-GFDL:
The other solution  was more disruptive and involved duplicating mom_get_input.f90 with only minor changes. This seemed like the better option.

PR #1272, opened 2020-12-11
Autoconf netCDF flag update

marshallward:
This cleans up some issues related to netCDF flag support which were needed to get FMS 2020.04 built correctly.

Specifically, this PR is more careful with the handling of C and Fortran flags, and tries to isolate the flags a little more correctly.  It also only conditionally checks `nc-config` and `nf-config`, and tries its best to rely on existing environment variables (e.g. `CPATH`) or user-configured flags when possible.

This does not update the default FMS version to 2020.04, but we could do that if desired.  All the tests appear to be passing on various platforms with 2020.04.  This is separate from the MOM6-examples update.


underwoo:
@marshallward please look at the netcdf m4 macro in [FRE-NCtools](https://github.com/NOAA-GFDL/FRE-NCtools/blob/master/m4/ax_lib_netcdf.m4). I made some extensive modifications that may be useful.

marshallward:
Thanks Seth, I'm looking through them now.  It looks like you also encountered a lot of similar problems with many of the flags (`--flibs` mashing `-L` and `-l` together, for example), and had to parse our the relevant bits.

I was doing something a bit similar (though nowhere near as thorough) but eventually abandoned that approach due to problems on some platforms.  (Mostly working around the user's implicit settings at install time, as far as I could tell).  Unlike the other flags, the `--prefix` based seemed to be more reliably defined.

After running this on a few different platforms, the safest option seemed to be to just avoid those other flags and use `--includedir` and `--libdir`, and falling back to `--prefix` if those are unavailable.

If I can get some time, I'll give your macros a try and see how they fare.

----

BTW the impetus here is that we were not linking to the netCDF C library (`-lnetcdf`), only the Fortran netCDF library, since for whatever reason it was not needed.  But it seems FMS 2020.04 does invoke the library now (presumably something in `read_mosaic.c`).

adcroft (approved):

PR #1274, opened 2020-12-14
CPT updates for diagnostics and configurations

adcroft:
This contains the modifications being used by the Ocean eddy CPT:
- Adds diagnostics of normal and shear stress
- Adds a coast feature to the basin builder module

I would send this to "main" but it's small enough that we cause less work by sending to "dev/gfdl".


marshallward:
Given that there are 21 commits, this may need a squash.  Also, do I need write permissions?  I am not able to do an update against `dev/gfdl`.

Hallberg-NOAA (approved):
These code changes appear to me to be correct, and they have passed the automated testing. This should be merged in (along with an update to MOM6-examples because of the updated available diagnostics and perhaps changes to the parameter_doc file due to the introduction of new parameter), once it has passed regression testing.

adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11686

PR #1277, opened 2020-12-21
correct CFC index check in src/tracer/MOM_OCMIP2_CFC.F90

### LINKED ISSUE #1273: typo in cfc index check
In the line of code
https://github.com/NOAA-GFDL/MOM6/blob/446becc0ec83541ce7c469a7d907d6b382433bae/src/tracer/MOM_OCMIP2_CFC.F90#L129
the value of `ind_cfc_11_flux` is being checked twice. It seems likely to me that the one of the conditionals should be checking `ind_cfc_12_flux`.

klindsay28:
fixes #1273


Hallberg-NOAA (approved):
I agree that this change fixes a bug.

adcroft (approved):
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11719

Hallberg-NOAA:
This PR has passed the automated consistency tests and it is so obviously correct that I am merging it in despite the fact that GFDL's pipeline testing is not working at the moment, perhaps because of problems with the Gaea file system, perhaps because of problems that were introduced with a previous commit. There is absolutely no way that there is any problem with this very simple and obviously correct PR.

PR #1278, opened 2020-12-22
Revert drivers

MJHarrison-GFDL:
This PR reverts recent modifications to MOM6 drivers for the ice shelf module.  

* Existing experiments with  ICE_SHELF = True need to modify their diag_table for ice shelf diagnostics. In the first column of the table, "ocean_model" becomes "ice_shelf_model" for ice shelf diagnostics.


MJHarrison-GFDL:
Replace with #1279

PR #1283, opened 2021-01-11
Revise framework

Hallberg-NOAA:
MOM6: +Revise framework code

  Restructured and revised code in the MOM6 framework directory to separate
MOM6-specific code from modules with direct calls to FMS or mpp routines.  All
answers and output files are bitwise identical, except for the ocean_geometry.F90 file
where incorrect dimensional rescaling of areas was corrected.  Although there are
new interfaces and new optional arguments, and this PR includes changes that
exercise these changes, the contents of the new framework directory can be
swapped with previous versions of the code to give identical answers and all
output files.

  The commits in this PR include:

- NOAA-GFDL/MOM6@45d29a99c +Added an explicit interface to open_file
- NOAA-GFDL/MOM6@51720fb33 +Split MOM_io_wrapper.F90 out from MOM_io.F90
- NOAA-GFDL/MOM6@37ef28bc0 +Add get_domain_components to MOM_domains.F90
- NOAA-GFDL/MOM6@81a6ff8ee +Add explicit interface for field_exists to MOM_io
- NOAA-GFDL/MOM6@d8806f48e +Rename rotated_mpp_chksum to rotated_field_chksum
- NOAA-GFDL/MOM6@adb8ec46b +Add deallocate_MOM_domain and get_layout_extents
- NOAA-GFDL/MOM6@502eb301e (*)Call MOM_write_field in write_ocean_geometry_file
- NOAA-GFDL/MOM6@7091a09c1 +Add MOM_write_field
- NOAA-GFDL/MOM6@ee7dd32c8 +Add create_MOM_domain and MOM_domain_init.F90

PR #1288, opened 2021-01-13
Always stream job logs in gitlab pipeline

adcroft:
- When a batch job fails on the gitlab pipeline we used to see the job logs but we suspect this behavior changed a long time ago.
- This commit intercepts an error from the sbatch command and causes the pipeline to fail only after streaming the log, which is the very next command after sbatch.
- A test of the interception working as intended uses the notorious #1283. This pipeline was failing without the logs being produced. With this cherry-picked commit the logs are now visible: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/jobs/59699 (GFDL-only).


adcroft:
Test (that should pass) is at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11847

marshallward (approved):

PR #1291, opened 2021-01-15
Oda use framework

MJHarrison-GFDL:
*This commit removes dependencies within the ocean data assimilation
driver on FMS infrastructure code.  Additionally, the increment to the assimilation
time is made before the DA call instead of after. This will change answers for test
cases which are using the DA interfaces, but aligns answers with SPEAR.


Hallberg-NOAA (approved):
These changes will be acceptable after the conflicts have been resolved.

Hallberg-NOAA:
The updated PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11888.

PR #1296, opened 2021-01-25
Options enabling old bugs to recover SPEAR

### LINKED ISSUE #1271: Recover older answers for compatibility with SPEAR
Since early 2018 @wfcooke has had to revert commits 48e90d0b928457e64c80b065c32f1131cc6b6dfb and bc6c6e65d658f7cdddf0c589ae770feb40287c01 to recover solutions in the SPEAR model. The first commit increased the accuracy with which density anomalies are calculated, and the second was a bug fix changing the value of the model's time in the diabatic phase. In that last few months we've diverged so far from the 2018 code that reverting the commits is no longer practical.

The dev team already agreed that we'd recover answers but @Hallberg-NOAA and I have only just looked at the best way to do that. Proposed fixes are to:
- [x] Add an extra equation of state using the inaccurate evaluation without RHO_REF subtraction (e.g. `EQN_OF_STATE="WRIGHT_LEFT"`);
- [x] Add a flag to optionally use the wrong time at certain places in the algorithm.

adcroft:
Four patches to recover answers for SPEAR. Three add new runtime options to enable bugs and the fourth adds back a runtime parameter that was obsoleted:
- abf6819 Adds back a bug providing the wrong time to diabatic processes
- f499d99 Adds back a less accurate form of the PGF
- e7b7b62 Adds back an older form of neutral diffusion transport accumulation
- a85c918 Adds back an indexing bug for the background viscosity

To enable all four bugs requires:
```
USE_INACCURATE_PGF_RHO_ANOM=True
USE_DIABATIC_TIME_BUG=True
NDIFF_USE_UNMASKED_TRANSPORT_BUG=True
KH_BG_2D_BUG=True
```

In #1271, @wfcooke reports that with these patches the SPEAR answers can be recovered.

Closes #1271


adcroft:
I should add that the new runtime parameters mean this will require an update to MOM6-examples

adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11935

Hallberg-NOAA (approved):
I have inspected these changes, and I agree that they are correct and necessary, and I also note that this PR has passed the TC and pipeline testing. I will handle the updates to MOM6-examples reflecting the changes in the MOM_parameter_doc files from this PR.

PR #1303, opened 2021-01-27
Rename MOM_ensemble_manager to MOM_ensemble_manager_infra

MJHarrison-GFDL:
Pass through interfaces to FMS with documentation.


Hallberg-NOAA:
The documentation of the interfaces looks great - thank you for taking this on.  However there are two issues with this commit that we might want consider.

  First, by renaming the file, there is now an inconsistency between the file and module names.  If the file name is MOM_ensemble_manager_infra.F90, the module should be MOM_ensemble_manager_infra, not MOM_ensemble_manager.

 The second is a structural issue with this commit. The plan is that all of the infrastructure specific modules (like ensemble_manager_infra.F90 ) will move out of the MOM6/src/framework directory into a directory like MOM6/config_src/infra/FMS1/ . However at that point, there should not be any calls to the config_src/infra code from other parts of MOM6 outside of the framework directory. Instead there should be two files, (say MOM_ensemble_manager.F90 and MOM_ensemble_infra.F90), with the calls going via MOM_ensemble_manager to MOM_ensemble_infra and then into the FMS or (or other) infrastructure.  This way hopefully we may be able to buffer any interface changes that we find we need to make to accommodate other infrastructures from requiring changes elsewhere in the MOM6 code (like in the ODA code), or even in routines that we don't know about because other users have not checked them into dev/gfdl or the other main forks.

  As an example of what we are thinking, please see the revised MOM_cpu_clock.F90 and MOM_cpu_clock_infra.F90.  In this case, it looks like all of the routines in MOM_ensemble_manager.F90 might just be pass-throughs to MOM_ensemble_infra.F90.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11950.

Hallberg-NOAA (approved):
This PR does a nice job of documenting the interfaces that MOM6 uses to the ensemble_manager code, it matches the new structure we are moving toward, and it has passed all of our TC and pipeline testing.

PR #1306, opened 2021-01-27
+Add explicit interfaces for MOM_domain_infra

Hallberg-NOAA:
Added explicit interfaces for the remaining routines in MOM_domain_infra,
including global_field, broadcast_domain, and compute_block_extent, and created
a new stand-alone routine MOM_define_layout inside of MOM_domains.F90 to replace
the mpp routine that it had previously wrapped.   Also added comments explicitly
documenting some of the integer parameters that are used to signal staggering or
communication patterns.  All answers are bitwise identical, but some routines
have been recreated within the MOM6 code, in the localized case of
MOM_define_layout with a slightly different interface.

### DIFF
--- a/src/framework/MOM_domain_infra.F90
+++ b/src/framework/MOM_domain_infra.F90
@@ -7,22 +7,17 @@ use MOM_coms_infra,  only : PE_here, root_PE, num_PEs
 use MOM_cpu_clock,   only : cpu_clock_begin, cpu_clock_end
 use MOM_error_infra, only : MOM_error=>MOM_err, NOTE, WARNING, FATAL
 
-use mpp_domains_mod, only : MOM_define_layout => mpp_define_layout, mpp_get_boundary
-use mpp_domains_mod, only : MOM_define_io_domain => mpp_define_io_domain
-use mpp_domains_mod, only : MOM_define_domain => mpp_define_domains
-use mpp_domains_mod, only : domain2D, domain1D, mpp_get_data_domain, mpp_get_domain_components
-use mpp_domains_mod, only : mpp_get_compute_domain, mpp_get_global_domain
-use mpp_domains_mod, only : mpp_get_domain_extents, mpp_deallocate_domain
-use mpp_domains_mod, only : mpp_update_domains, global_field_sum => mpp_global_sum
+use mpp_domains_mod, only : domain2D, domain1D, group_pass_type => mpp_group_update_type
+use mpp_domains_mod, only : mpp_define_io_domain, mpp_define_domains, mpp_deallocate_domain
+use mpp_domains_mod, only : mpp_get_domain_components, mpp_get_domain_extents
+use mpp_domains_mod, only : mpp_get_compute_domain, mpp_get_data_domain, mpp_get_global_domain
+use mpp_domains_mod, only : mpp_get_boundary, mpp_update_domains, global_field_sum => mpp_global_sum
 use mpp_domains_mod, only : mpp_start_update_domains, mpp_complete_update_domains
 use mpp_domains_mod, only : mpp_create_group_update, mpp_do_group_update
-use mpp_domains_mod, only : group_pass_type => mpp_group_update_type
 use mpp_domains_mod, only : mpp_reset_group_update_field, mpp_group_update_initialized
 use mpp_domains_mod, only : mpp_start_group_update, mpp_complete_group_update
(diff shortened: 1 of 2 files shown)


Hallberg-NOAA:
This PR passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11953.

src/framework/MOM_domain_infra.F90
```diff
@@ -7,22 +7,17 @@ module MOM_domain_infra
...
-use mpp_domains_mod, only : MOM_define_domain => mpp_define_domains
-use mpp_domains_mod, only : domain2D, domain1D, mpp_get_data_domain, mpp_get_domain_components
-use mpp_domains_mod, only : mpp_get_compute_domain, mpp_get_global_domain
-use mpp_domains_mod, only : mpp_get_domain_extents, mpp_deallocate_domain
-use mpp_domains_mod, only : mpp_update_domains, global_field_sum => mpp_global_sum
+use mpp_domains_mod, only : domain2D, domain1D, group_pass_type => mpp_group_update_type
+use mpp_domains_mod, only : mpp_define_io_domain, mpp_define_domains, mpp_deallocate_domain
+use mpp_domains_mod, only : mpp_get_domain_components, mpp_get_domain_extents
+use mpp_domains_mod, only : mpp_get_compute_domain, mpp_get_data_domain, mpp_get_global_domain
+use mpp_domains_mod, only : mpp_get_boundary, mpp_update_domains, global_field_sum => mpp_global_sum
```
marshallward, line 14: From our discussions, it seems that `global_field_sum` is being left as a pass-through because it is unused in MOM source, but still could be used by existing drivers or other config_src code.

I'd recommend breaking this out into a separate statement, with a comment explaining why this function is unwrapped, explaining that it is informally part of the MOM API but is unused and undocumented.

src/framework/MOM_domain_infra.F90
```diff
@@ -7,22 +7,17 @@ module MOM_domain_infra
...
 use MOM_error_infra, only : MOM_error=>MOM_err, NOTE, WARNING, FATAL
 
-use mpp_domains_mod, only : MOM_define_layout => mpp_define_layout, mpp_get_boundary
-use mpp_domains_mod, only : MOM_define_io_domain => mpp_define_io_domain
-use mpp_domains_mod, only : MOM_define_domain => mpp_define_domains
-use mpp_domains_mod, only : domain2D, domain1D, mpp_get_data_domain, mpp_get_domain_components
-use mpp_domains_mod, only : mpp_get_compute_domain, mpp_get_global_domain
-use mpp_domains_mod, only : mpp_get_domain_extents, mpp_deallocate_domain
-use mpp_domains_mod, only : mpp_update_domains, global_field_sum => mpp_global_sum
+use mpp_domains_mod, only : domain2D, domain1D, group_pass_type => mpp_group_update_type
```
marshallward, line 10: Since `group_pass_type` is used in the code, does this need to be documented somehow?

Hallberg-NOAA: Because this is an opaque type whose contents are never referenced anywhere that is visible in the MOM6 code, I don't think that we need to document its contents (even if we could figure out how to do this).

src/framework/MOM_domain_infra.F90
Hallberg-NOAA: I don't know that this code is often examined by other people, and while these are in module use statements for the nuopc and mct couplers, they are not actually used there. Rather than putting in some comments here that will just be ignored (or perhaps as well as), I would like to either (1) put in an issue, or (2) actually remove the unused module use statements from the other couplers, now that we have a way to compile them for testing purposes.

src/framework/MOM_domain_infra.F90
marshallward: See response above, PR addresses this change.

src/framework/MOM_domain_infra.F90
marshallward: I still think it's a good idea to document exceptions to the rules, even if it is only referenced by a very small group (perhaps even just ourselves).

In order to avoid creating work for you, I have submitted a PR to your branch which makes this change.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11988

PR #1307, opened 2021-01-28
Added explicit interfaces for MOM_error_infra

Hallberg-NOAA:
Added explicit interfaces with documentation for the routines MOM_err, stdout
and stdlog in MOM_error_infra.F90.  Also clarified some of the comments
describing MOM_error.F90.  All answers are bitwise identical and no interfaces
are changed as seen from outside MOM_error.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11974

marshallward (approved):

PR #1308, opened 2021-01-28
NCAR sponge merge

MJHarrison-GFDL:
This PR contains updates for u,v accelerations in MOM_ALE_sponge provided by NCAR.  

  - A bug was identified in the case of incoming sponge data residing on the model horizontal grid 
    where the returned mask was not initialized properly. 
  - The 2018 answers flags are being used to retain the bug and should be set to False for new experiments.
  - Enable UV sponges by setting SPONGE_UV=True
  - Sponge accelerations use the tracer timescale by default but can be set independently.    
  - New sponge tendency diagnostics added (sp_tendency_{temp,salt} and sp_tendency_{u,v})


src/initialization/MOM_state_initialization.F90
```diff
@@ -1914,8 +1992,38 @@ subroutine initialize_sponges_file(G, GV, US, use_temperature, tv, param_file, L
...
+                       'UV damping to target values only available in ALE mode')
+
+ if (sponge_uv .and. use_ALE)  then
+
+   filename = trim(inputdir)//trim(state_uv_file)
+   call log_param(param_file, mdl, "INPUTDIR/SPONGE_STATE_UV_FILE", filename)
+   if (.not.file_exists(filename, G%Domain)) &
+     call MOM_error(FATAL, " initialize_sponges: Unable to open "//trim(filename))
+
+   call set_up_ALE_sponge_vel_field(filename, u_var, filename, v_var, Time, G, GV, US, ALE_CSp, u, v)
```
angus-g, line 2017: Does this need to handle the case of fixed velocity fields as well? This block initialises the velocity fields as varying, regardless of `time_space_interp_sponge`

src/parameterizations/vertical/MOM_ALE_sponge.F90
```diff
@@ -665,39 +748,41 @@ subroutine set_up_ALE_sponge_vel_field_fixed(u_val, v_val, G, GV, u_ptr, v_ptr,
...
+  real, target, dimension(SZI_(G),SZJB_(G),SZK_(GV)), intent(in) :: v_ptr !< v pointer to the field to be damped
 
   integer :: j, k, col
   character(len=256) :: mesg ! String for error messages
 
   if (.not.associated(CS)) return
 
   ! stores the reference profile
-  allocate(CS%Ref_val_u%p(CS%nz_data,CS%num_col_u))
+  allocate(CS%Ref_val_u%p(CS%Ref_val_u%nz_data,CS%num_col_u))
```
angus-g, line 766: Possible that `nz_data` here is uninitialised?

src/initialization/MOM_state_initialization.F90
MJHarrison-GFDL: Should work provided a time axis exists in the file. To be safe, I added a call to initialize using the fixed interface.

Hallberg-NOAA: I think that we should accept this commit as it currently stands, but if this concern is manifest in practice, it can be revisited by introducing either a new PR to correct it or a new issue. As of now, this PR is passing all of our tests.

Hallberg-NOAA (approved):
I have examined these changes and they seem well thought-out and useful.  The issues that were raised with some of the versions have been addressed, and this PR has passed both the TC testing and the pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12000.  I think that it is ready to be merged into MOM6.

  However, because of the large number of commits (28) within this PR relative to the changes, and the fact that a number of the intermediate commits had problems that would preclude their use in any kind of Git bisection, I am going to do this as a squash-merge.

PR #1310, opened 2021-01-28
interface for MOM_interp_infra

raphaeldussin:


raphaeldussin:
@Hallberg-NOAA this should be working pretty soon, please squash and merge

Hallberg-NOAA (commented):
Thank you for this PR that provides explicit interfaces and documentation for the horizontal interpolation routines.

My once concern here is that we have made the decision that we will only add interfaces for the versions that are actually used by MOM6, not all of the variants that are provided by FMS. I suspect that some of the variants that were introduced to MOM_interp_infra.F90 are not actually being used, and hence can be removed.

raphaeldussin:
@Hallberg-NOAA  the unused interfaces are removed now.

Hallberg-NOAA:
This PR is being tested with https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/11995, and it should be merged (as a squash merge because a number of the intermediate commits encountered errors) as soon as the pipeline testing has passed.

Hallberg-NOAA (approved):
These changes do a very nice job of documenting the interpolation interfaces, and I think that the new names do a much better job of describing what the routines in question actually do.  Now that it has passed all of the TC and pipeline testing, I am happy approve this PR, although the actual merge of this PR should be a squash-merge due to the intermediate commits that had problems exposed by the testing.

PR #1311, opened 2021-01-28
+Partial consolidation of netcdf calls in MOM_io

Hallberg-NOAA:
Took preliminary steps to consolidate the direct calls to netcdf in
MOM_io.F90. Renamed check_grid_def to verify_variable_units and moved it to
MOM_io, where it can be used more widely than just in MOM_regridding.F90.  Also
split get_var_sizes and get_varid out of num_timelevels and made them publicly
visible.  All answers are bitwise identical, but there are new public interfaces
and the renaming of one routine.

### DIFF
--- a/src/ALE/MOM_regridding.F90
+++ b/src/ALE/MOM_regridding.F90
@@ -6,7 +6,7 @@ module MOM_regridding
 use MOM_error_handler, only : MOM_error, FATAL, WARNING
 use MOM_file_parser,   only : param_file_type, get_param, log_param
 use MOM_io,            only : file_exists, field_exists, field_size, MOM_read_data
-use MOM_io,            only : slasher
+use MOM_io,            only : verify_variable_units, slasher
 use MOM_unit_scaling,  only : unit_scale_type
 use MOM_variables,     only : ocean_grid_type, thermo_var_ptrs
 use MOM_verticalGrid,  only : verticalGrid_type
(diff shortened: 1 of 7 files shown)


klindsay28:
@Hallberg-NOAA, note that the netCDF library doesn't check for adequate space when reading in attributes. The netCDF library will happily write beyond the end of your array if it is not long enough. This is a pain to debug when it happens. I've yet to see a compiler that catches this, I suspect because the underlying read is happening in C. I think you are more likely to get bitten with char/string attributes than other types, because non-char attributes tend to be scalars in typical netCDF files, not arrays.

I suggest comparing the attribute length to the size of `att_val` in `read_attribute_*` before calling `NF90_get_att`, particularly for `read_attribute_str`.

Code in the CESM fork of POP that does this for char attributes is at https://github.com/ESCOMP/POP2-CESM/blob/9f2a48328dda6bd98be74f4ccfc13248b1b13868/source/io_netcdf.F90#L232-L258. Note that this code is using the pio API, not the netCDF API. But the issues are the same.

Hallberg-NOAA:
Added more commits to PR#1311 to further concentrate I/O calls in MOM_io.F90, and to respond
to comments on these capabilities. These additional commits are:

- NOAA-GFDL/MOM6@89392e3df +Add ncid_in optional args to MOM_io read routines
- NOAA-GFDL/MOM6@3fed47abc +read_attribute uses allocatable character string
- NOAA-GFDL/MOM6@da14640e5 Added missing Doxygen comments for read_attribute
- NOAA-GFDL/MOM6@44a8bde1c Replace netCDF calls in MOM_tracer_Z_init
- NOAA-GFDL/MOM6@3c3bda753 Replace netCDF calls in MOM_shared_initialization
- NOAA-GFDL/MOM6@953624267 Replace netCDF calls in read_depth_list
- NOAA-GFDL/MOM6@8c01d0e10 +Added read_variable and read_attribute to MOM_io
- NOAA-GFDL/MOM6@199a45312 (*)Corrected bug in field_chksum
- NOAA-GFDL/MOM6@14727e599 +Only do reads from root_PE for get_var_sizes

With these added commits, I consider PR#1311 to be complete, and have no plans for any additions.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12001.

marshallward (approved):

PR #1317, opened 2021-02-05
initialize field-specific vertical extent for target data in fixed sp…

MJHarrison-GFDL:
Fixes issue for fixed sponge cases introduced by #1308


angus-g:
Maybe also for this fixup: it looks like `h_col` is never set when applying the sponge on tracer fields: https://github.com/NOAA-GFDL/MOM6/blob/b5158bb0705d098f6594abbd10250edcba23b316/src/parameterizations/vertical/MOM_ALE_sponge.F90#L948-L968

Hallberg-NOAA (approved):
I agree with these changes, and they have passed both the TC testing and the pipeline testing at
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12015 .

PR #1318, opened 2021-02-06
+(*)Restore ability to use decomposed restart files

Hallberg-NOAA:
Added optional MOM_Domain arguments to the 0-d and 1-d versions of
MOM_read_data, enabling them to read from a domain-decomposed file if the single
file is not present.  The MOM_read_data calls in restore_state were modified to
use these new optional arguments, thereby restoring the ability of MOM6 to read
from domain-decomposed restart files, which had been broken by a known miscreant
on January 3, 2021 with MOM6 commit aea16f73.  All answers and output are
bitwise identical in cases that worked.

### DIFF
--- a/src/framework/MOM_io_infra.F90
+++ b/src/framework/MOM_io_infra.F90
@@ -315,17 +315,23 @@ subroutine get_axis_data( axis, dat )
   call mpp_get_axis_data( axis, dat )
 end subroutine get_axis_data
 
-!> This routine uses the fms_io subroutine read_data to read a scalar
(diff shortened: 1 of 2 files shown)


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12014

marshallward:
Looks good, tests are passing.  I'm expecting we'll discuss on Monday how to modify the test suite to catch these in the future.

marshallward (approved):

PR #1319, opened 2021-02-08
Included array_global_min_man in MOM_generic_tracer.F90

Hallberg-NOAA:
Included a copy of mpp_array_global_min_max from mpp_utilities in
MOM_generic_tracer.F90 as array_global_min_max to document what had been an
interface from mpp that is used in MOM6, and to break a dependence on a part of
the FMS code.  There are a number of serious problems with this routine, both
with it reporting values that are incorrect and with it failing most forms of
self-consistency testing.  But rather than fixing any of these problems, they
are noted in comments, and this is an exact reproduction of its predecessor.


adcroft (approved):
Pending https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12040

PR #1321, opened 2021-02-08
+Cleaned up MOM_inter_infra argument documentation

Hallberg-NOAA:
Modified the MOM_inter_infra argument declarations to follow the usual pattern
for the declaration of optional arguments in MOM6 code.  Also eliminated an
optional argument to horiz_inter_from_weights that was only described as
"unknown" always hard-coded to .true. in all calls from the MOM6 code.
All answers are bitwise identical, but one optional argument was removed.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12043

marshallward (approved):
Au revoir, `new_missing_handle`

PR #1322, opened 2021-02-10
Fixing an indexing error in MOM_barotropic

herrwang0:
There is an indexing error related to the linear wave drag calculation in `MOM_barotropic`. This can lead to a negative column thickness used by the wave drag, and make `bt_rem_v`>1, which can result in extremely large barotropic velocity.

### DIFF
--- a/src/core/MOM_barotropic.F90
+++ b/src/core/MOM_barotropic.F90
@@ -1462,7 +1462,7 @@ subroutine btstep(U_in, V_in, eta_in, dt, bc_accel_u, bc_accel_v, forces, pbce, 
     do J=js-1,je ; do i=is,ie ; if (CS%lin_drag_v(i,J) > 0.0) then
       Htot = 0.5 * (eta(i,j) + eta(i,j+1))
       if (GV%Boussinesq) &
-        Htot = Htot + 0.5*GV%Z_to_H * (CS%bathyT(i,j) + CS%bathyT(i+1,j+1))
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA:
This PR corrects a bug that will change answers with parameter settings that are unusual, and might never have been use in any long simulations.  This bug only applies if BT_LINEAR_WAVE_DRAG is explicitly set to .True. (which is exceedingly unusual), and even then only in Boussinesq mode.  In figuring out whether to fix this but without a runtime flag to reproduce it, we will be asking around to see whether it is in active use. It has never been used in simulations at GFDL, and even in the tidal simulations for which it was implemented in the fist place in November, 2017, He Wang's runs seem to be the first time this option was used. With this option, there needs to be a file with linear wave drag piston velocities, so if a user does not know they are using this option, they almost certainly are not.

  If, however, anyone reads this and discovers that they are using this option, please add a comment to this PR to help guide us on how to handle this bugfix, which will change answers in an exceedingly obscure configuration.

Hallberg-NOAA (approved):
This change is correct, but it will change answers if this particular section of code is used.  It has passed the TC testing and the pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12068.

Given that there have been no objections raised in the related discussion at (https://github.com/NOAA-GFDL/MOM6/discussions/1325), and the fact that it appears that this code has not yet been exercised, I am going to approve this correction to an obvious bug without the usual approach of adding a run-time flag to reproduce the old, incorrect answers.

PR #1332, opened 2021-02-23
*Bugfix for ALE sponges (pass non-zero target thickness to remapping)

MJHarrison-GFDL:
- this addresses a bug in the apply_ALE_sponge routine which
  	had been passing zero values to the remapping routine for the
	target grid thicknesses. This bug was identified by @abocez


adcroft:
Can you also send this to the dev-gfdl-main-candidate-2021-02-19 branch? We're patching that first.

marshallward:
1. Do you want to remove the print statements?
2. Let's update `h_col` instead of passing `h(i,j,:)`, if only to follow the style guide (and avoid the implicit copy)
3. This should go to the dev candidate rather than `dev/gfdl`.

MJHarrison-GFDL:
>     1. Do you want to remove the print statements?
> 
>     2. Let's update `h_col` instead of passing `h(i,j,:)`, if only to follow the style guide (and avoid the implicit copy)
> 
>     3. This should go to the dev candidate rather than `dev/gfdl`.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12106

marshallward:
Handled in #1333, closing.

PR #1339, opened 2021-02-25
+New config_src directory structure

Hallberg-NOAA:
Created the new config_src directory tree structure, as agreed upon at
https://github.com/NOAA-GFDL/MOM6/discussions/1286, to eventually accommodate
the selection of different infrastructures.  No .F90 files are changed but there
are small changes to ac/configure.ac to accommodate the new structure while also
allowing for the old target to use the old structure.  All answers are bitwise
identical, and all MOM6-examples and TC tests are passing.

  This change will require modifications to various compile scripts.


ac/configure.ac
```diff
@@ -42,18 +42,20 @@ srcdir=$srcdir/..
 # Default to symmetric grid
 # NOTE: --enable is more properly used to add a feature, rather than to select
 #   a compile-time mode, so this is not exactly being used as intended.
-MEM_LAYOUT=${srcdir}/config_src/dynamic_symmetric
+MEM_LAYOUT=${srcdir}/config_src/memory/dynamic_symmetric
+AC_CHECK_FILE($MEM_LAYOUT, [MEM_LAYOUT=$MEM_LAYOUT], [MEM_LAYOUT=${srcdir}/config_src/dynamic_symmetric])
```
marshallward, line 46: I'm guessing this directory also needs to be updated.

Hallberg-NOAA: No, it is correct as written.  The point here is that if the MEM_LAYOUT file does not exist, it is because this is the target directory that is still using the old directory structure.  Because I know essentially nothing about autoconf, I spent much more time this morning figuring out how to make the TC make commands work with both the old and new directory structures than I did with actually moving the files around!

marshallward: But it's looking in a directory which ought to no longer exist.  Is this being done to accommodate the regression test?

Hallberg-NOAA: Without this line the TC regression tests were failing to compile; with it they do.  Once we are no longer interested in comparing runs with potentially different directory structures, we can get rid of the ACC_CHECK_FILE line.

marshallward: Then at the least there needs to be a comment explaining why this is here.  It's a nonsense macro in the context of the repository.  While we do provide a regression test, we should not leave around musty old content to accommodate it.

marshallward: As a point of reference, the very last PR to main also did not pass regression.

PR #1343, opened 2021-02-28
Depth-integrated momentum budget diagnostics

hmkhatri:
Building on depth-averaged momentum budget diagnostics, I have added diagnostics for depth-integrated momentum budget terms. These new diagnostics are required for analyzing the depth-integrated vorticity budget. Some of the terms in the depth-integrated vorticity budget are sensitive and offline calculation of these diagnostics can result in significant errors. Hence, online diagnostic calculations are required for accurate analysis.

Also, typos in `register_diag_field` for some old diagnostics have been corrected.

### DIFF
--- a/src/core/MOM_dynamics_split_RK2.F90
+++ b/src/core/MOM_dynamics_split_RK2.F90
@@ -1453,12 +1453,12 @@ subroutine initialize_dyn_split_RK2(u, v, h, uh, vh, eta, Time, G, GV, US, param
 
   CS%id_intz_PFu_2d = register_diag_field('ocean_model', 'intz_PFu_2d', diag%axesCu1, Time, &
       'Depth-integral of Zonal Pressure Force Acceleration', 'm2 s-2', &
-      conversion=US%L_T_to_m_s**2)
+      conversion=US%Z_to_m*US%L_T2_to_m_s2)
   if(CS%id_intz_PFu_2d > 0) call safe_alloc_ptr(CS%ADp%diag_hu,IsdB,IedB,jsd,jed,nz)
 
   CS%id_intz_PFv_2d = register_diag_field('ocean_model', 'intz_PFv_2d', diag%axesCv1, Time, &
       'Depth-integral of Meridional Pressure Force Acceleration', 'm2 s-2', &
-      conversion=US%L_T_to_m_s**2)
+      conversion=US%Z_to_m*US%L_T2_to_m_s2)
   if(CS%id_intz_PFv_2d > 0) call safe_alloc_ptr(CS%ADp%diag_hv,isd,ied,Jsd,JedB,nz)
 
   CS%id_hf_CAu_2d = register_diag_field('ocean_model', 'hf_CAu_2d', diag%axesCu1, Time, &
@@ -1473,12 +1473,12 @@ subroutine initialize_dyn_split_RK2(u, v, h, uh, vh, eta, Time, G, GV, US, param
 
   CS%id_intz_CAu_2d = register_diag_field('ocean_model', 'intz_CAu_2d', diag%axesCu1, Time, &
       'Depth-integral of Zonal Coriolis and Advective Acceleration', 'm2 s-2', &
-      conversion=US%L_T_to_m_s**2)
+      conversion=US%Z_to_m*US%L_T2_to_m_s2)
   if(CS%id_intz_CAu_2d > 0) call safe_alloc_ptr(CS%ADp%diag_hu,IsdB,IedB,jsd,jed,nz)
 
   CS%id_intz_CAv_2d = register_diag_field('ocean_model', 'intz_CAv_2d', diag%axesCv1, Time, &
       'Depth-integral of Meridional Coriolis and Advective Acceleration', 'm2 s-2', &
-      conversion=US%L_T_to_m_s**2)
+      conversion=US%Z_to_m*US%L_T2_to_m_s2)
   if(CS%id_intz_CAv_2d > 0) call safe_alloc_ptr(CS%ADp%diag_hv,isd,ied,Jsd,JedB,nz)
 
   CS%id_uav = register_diag_field('ocean_model', 'uav', diag%axesCuL, Time, &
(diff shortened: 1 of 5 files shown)


hmkhatri:
So, ``Expression verification / test-repro-and-dims`` did not pass, but the compilation and test runs worked fine. I am not sure how to correct this issue. Any help would be appreciated.

marshallward:
The H and L dimensionality tests are failing (e.g. `tc2.dim.l`, `tc2.dim.h`), so you need to work out why the dimensions of these diagnostics are not correct.

Just quickly eyeing the code, the depth-integrated diagnostics probably need a factor of H.  But it's probably better if you work through it more carefully.

I'd recommend trying to run the tests in `.testing` before pushing any new updates.

marshallward:
It looks like it might just be an H vs L issue.  Dimensionality is in a Pi-theorem sense (scaling invariance), rather than as an MKS unit.

hmkhatri:
@marshallward Based on your comments, I tried the following to resolve the H vs L dimensionality test issue,

In `register_diag_field`,  I am now using `conversion=US%Z_to_m*US%L_T2_to_m_s2` as the diagnostics have layer thicknesses multiplied (earlier I used  `conversion=US%L_T_to_m_s**2`). However, this did not resolve the issue and dimensionality tests are still failing ([see here](https://github.com/hmkhatri/MOM6/commit/2165ebb6fb71d7e455e4cf254fb8d89fe746e3ee) )

So, I do not know how to rectify it. 
Could someone please have a quick look and let me know if there is anything unusual?

ashao:
I agree with @marshallward , it's failing because your missing a thickness scaling converstion. I think you need something like:
```
L_T2_to_m_s2*Z_to_m
```

marshallward:
H and Z are separate dimensions here, with H corresponding to thickness and Z being more positional (e.g. vertical gradients).

I will test myself, but you may have luck using H in place of Z.

ashao:
I think you're right, but would only be detectable in non-Boussinesq mode.

ashao:
So thinking something like?:
```
US%(L_T2_to_m_s2)*(GV%H_to_Z)*(US%Z_to_m)
```

hmkhatri:
Thanks, @marshallward and @ashao 

I have made corrections as per your suggestions and all checks passed.

@marshallward also suggested not to use allocatable arrays and just declare the arrays as a stack, e.g.
`real, dimension(SZIB_(G),SZJ_(G)) :: intz_PFu`

I remember there was a discussion on a similar point in #1163.  With regards to performance and memory load, is there a preference to use allocatable vs stack arrays (for both 2D and 3D diagnostics)?

@ashao @Hallberg-NOAA Your inputs would be very helpful.

src/core/MOM_dynamics_split_RK2.F90
```diff
@@ -963,6 +1007,25 @@ subroutine step_MOM_dyn_split_RK2(u, v, h, tv, visc, Time_local, dt, forces, p_s
   !  enddo ; enddo ; enddo
   !  call post_data(CS%id_hf_v_BT_accel, hf_v_BT_accel, CS%diag)
   !endif
+  if (CS%id_intz_u_BT_accel_2d > 0) then
+    allocate(intz_u_BT_accel_2d(G%IsdB:G%IedB,G%jsd:G%jed))
```
ashao, line 1011: These shouldn't be allocated and deallocated each and every timestep

ashao:
@hmkhatri When we talked about this originally (I believe with @Hallberg-NOAA), the goal was to make sure that we don't get into a position where we're adding a bunch of diagnostic-only arrays to the stack. @marshallward correct me if I'm wrong, but hypothetically an unused heap array doesn't actually add to the memory footprint because it is only allocated if it's actually used. For example if you did something like
```
real, dimension(10,10) :: foo
return
```
the memory for foo is never allocated or used, but if you did something like
```
real, dimension(10,10) :: foo
foo(:,:) = 0.
```
then it would be.

Performance-wise, I would guess that what @marshallward is objecting to is the `allocate` and `deallocate` around the `post_data`. That certainly bog down the model since it does take a non-negligible amount of time to do those operations. As you did earlier, where it's an allocatable array inside of the CS, and it's allocated during model initialization, that's a negligible performance hit.

src/core/MOM_dynamics_split_RK2.F90
hmkhatri: Thanks for pointing this out. I will the necessary changes.

hmkhatri:
@ashao Thanks for the comments. I agree that I should not have use allocation and deallocation within initialization and end subroutines only. I will correct these.

marshallward:
Yes, I agree with what @ashao wrote.  If an array is declared on stack but never used, then it should not use any memory.  ("Should" is the key word here, but that is how it's supposed to work.)  There is also the cost of repeated alloc/dealloc per timestep to consider.  And if there is any benefit to allocation, a good Fortran compiler is just going to do it behind the scenes anyway.

I think our policy here (if I can dare to call it that) is not yet clear though, so perhaps @Hallberg-NOAA should weigh in.

hmkhatri:
@ashao @marshallward  Thanks for the comments. I agree that I should not have used allocation and deallocation in each timestep. This was my fault. I will correct it, and move memory allocation calls to initialization subroutines.

In the earlier pull request, I had initially defined pointers for diagnostic arrays in the CS control structure.  For some reason, we wanted to keep the arrays used for the diagnostics local to the modules where they were being used. To have the combination of allocatable arrays and local variables, I ended up putting allocation and deallocation commands within the timestep loop. I agree that this is not ideal. 

For now, I can either define allocatable arrays in the main control structure of the modules, so the arrays are allocated only once, but the arrays will not be local anymore. Or I could just define stack arrays locally. Let me know if there are better ways to do it.

ashao:
IIRC @hmkhatri, there was a couple of arrays that were being updated in one part of the dynamics, but unavailable in the other part of the code where you needed to post the data. In order to avoid making the call trees convoluted and doing unnecessary array copies, we (including with @Hallberg-NOAA ) decided to just assign a pointer array within the lower level CS.

hmkhatri:
@ashao thanks for the clarification. For now, I have just initialized 2D stack arrays to avoid allocate/deallocates. Does the PR require any changes?

StephenGriffies:
We had a meeting with @marshallward today and this pull request follows his recommendation.  In that meeting, we concluded that the diagnostics in MOM6 follow a variety of styles for memory management.  Marshall's suggestion is to define fixed sized arrays as part of the particular subroutine where the diagnostic is needed (on the "stack" I believe is the lingo).  This approach contrasts to the allocate/deallocate done in the previous pull request (and as done for some other MOM6 diagnostics).  

Standardization of MOM6 diagnostic memory management would be useful.  But for this pull, we hope to be doing no harm and not to change anything that was already there. We are not trying to produce the standard.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12153

marshallward (approved):

PR #1344, opened 2021-03-02
Update CVMix submodule

adcroft:
This PR will point the CVMix submodule to the latest tag available, and was implemented in three consecutive commits:
1. Update to v0.93-beta with no API change
2. Update to v0.94b-beta which has an API change and thus a MOM6 code change
3. Update to v0.98-beta with no API change

Each commit was tested on MOM6-examples and GH actions. No answer changes were detected **but** MOM_tidal_mixing.F90 is not exercised in any of our tests. Providing the two intermediate commits provides finer granularity incase anyone has to `git bisect` for why CVMix changed answers.

I suggest @gustavo-marques have a look at this PR since NCAR is most likely to be using these modules.


gustavo-marques:
We use some of the modules touched here and I can confirm that this PR does not change answers for us.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12233

marshallward (approved):
I don't see any obvious answer changes, just some refactoring and maybe some new options.  So given that and no detected regressions, presumably this is safe to merge.

PR #1345, opened 2021-03-05
Typo correction

hmkhatri:
In many array allocations, I had used `Jsd` instead of `JsdB` for indices in many places. It might be irrelevant but I have made the correction anyway.


marshallward (approved):
Changes look correct.

I find it interesting though that we have no way to verify that the bounds of an array are correct if we do not actually use the entire bounds.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12171

Issue #1313, opened 2021-01-30
Date in ocean_solo.res does not override nml as intended


herrwang0:
https://github.com/NOAA-GFDL/MOM6/blob/aee4df21e19ff1b816a40bd55bafb84f12e2b014/config_src/solo_driver/MOM_driver.F90#L254

It looks like the code is trying to locate `ocean_solo.res` in `dirs%restart_input_dir`. But if I interpreted it correctly, `dirs` is not yet assigned here. 

I guess a hotfix could be simply adding something like
```Fortran
call get_MOM_input(dirs=dirs)
```
right before the line in question (in which case obviously `get_MOM_input` needs to be imported from `MOM_get_input` module).

marshallward:
Sorry for only just now looking into this, I think we've all had a lot of distractions in the last two weeks.

I think I see what you are saying.  In the solo driver, `dirs` is an internal variable for `initialize_MOM`, passed into `get_MOM_input` call, and is then an output back to the driver who called the initialization.  Except this call is on L299, ~50 lines after the block mentioned above.

Unsurprisingly, none of our tests ever evaluate this block:

https://codecov.io/gh/NOAA-GFDL/MOM6/src/dev%2Fgfdl/config_src/solo_driver/MOM_driver.F90#L256

So I think you have a good point.  Perhaps there was never an expectation of a calendar change in the middle of a set of runs.

I'll run it past the others and see what the intended behavior is here.  Thanks very much for noticing this.

Hallberg-NOAA:
@herrwang0, I think that your proposed fix of adding

`call get_MOM_input(dirs=dirs)`

right before the line that first uses the value of `dirs%restart_input_dir` is actually a very good solution.  FMS reads all of the namelist files from a single processor and then broadcasts it when it starts up, so this is not an expensive solution.  The one thing extra that I would do is to add a comment before this new line to the effect that
` ! The contents of dirs will be reread in initialize_MOM.`

  There is a similar issue in `ice_solo_driver/ice_shelf_driver.F90`, but the other drivers and caps seem to be fine.  In the case of `ice_solo_driver/ice_shelf_driver.F90`, the solution is to move the line `call get_MOM_input(param_file, dirs)` so that it occurs before the first test in that file that references `dirs%restart_input_dir` .

@herrwang0, if you could, I would appreciate it if you would put in a PR with your solution to this issue, along with the analogous change to ice_solo_driver/ice_shelf_driver.F90.

PR #1348, opened 2021-03-09
Bugfix: empty restart_input_dir for *_solo.res

herrwang0:
The ocean/ice solo drivers look for "ocean_solo.res" and "ice_solo.res" in `dirs%restart_input_dir`. However, `dirs` is only initiated after the call of `file_exist` for the ".res" files.

This PR fixes this issue (https://github.com/NOAA-GFDL/MOM6/issues/1313) by moving `get_MOM_input` in `ice_self_driver.F90` and adding an additional `get_MOM_input` in `MOM_driver.F90`, respectively.

### DIFF
--- a/config_src/ice_solo_driver/ice_shelf_driver.F90
+++ b/config_src/ice_solo_driver/ice_shelf_driver.F90
@@ -185,6 +185,8 @@ program Shelf_main
     endif
   endif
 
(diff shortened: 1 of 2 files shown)


Hallberg-NOAA (approved):
I agree that these changes to correct the problem identified in Issue #1313, and that they are the right solution.

PR #1350, opened 2021-03-09
Explicitly set (1x1) io_domain as a default

Hallberg-NOAA:
Added code to explicitly set a (1x1) io_domain when no other io_layout is
specified, complying with changing requirements for 2020 and later versions of
FMS, and following the default behavior of previous versions.  Also corrected
three module use statements in infra/FMS1 to eliminate any dependencies of code
in config_src/infra/FMS1 on code in src/framework.  These use statements
eventually point to the same place as before, but with less indirection.  This
change should facilitate later steps to compile everything in and under the
config_src/infra directories as libraries.  All answers are bitwise identical.

  This change will allow MOM6 to work with newer versions of FMS than 2019.01.03, and
it will enable efforts to debug why these newer versions are giving unexpected runtime
errors with the PGI 19.10.0 compiler.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6.git

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12220

marshallward (approved):

PR #1355, opened 2021-03-23
Testing: Recurse target submodules, LDFLAGS hook

marshallward:
Two minor changes to the .testing build:

- We now apply `--recurse-submodules` to the target build in the
  regression test.  This is required after an update to the submodules,
  when the target submodule is out of sync with the main branch (e.g.
  dev/gfdl at NOAA-GFDL).

- A LDFLAGS_USER hook was added to the `.testing/Makefile`
  configuration, similar to the FCFLAGS_* hooks.  This is required if
  the library dependencies do not reside in the default directores.

  For example, this may be needed for a custom netCDF or MPI library.

### DIFF
--- a/.testing/Makefile
+++ b/.testing/Makefile
@@ -84,6 +84,9 @@ FCFLAGS_COVERAGE ?=
 # - FMS cannot be built with the same aggressive initialization flags as MOM6,
 #   so FCFLAGS_INIT is used to provide additional MOM6 configuration.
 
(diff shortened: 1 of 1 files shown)


marshallward:
Now that Gaea is running .testing, it probably makes sense to start testing these:

https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12324

adcroft (approved):

PR #1356, opened 2021-03-23
Bugfix in getting surface currents [Additions from GMAO (3)]

sanAkel:
This PR fixes a bug.

In https://github.com/NOAA-GFDL/MOM6/pull/1134 `subroutine ocean_model_get_UV_surf` was introduced, but it lacked the needed call to `pass_vector` to correctly update halo regions. This is now corrected. 

Thanks to @zhaobin74 for bringing it up and @Hallberg-NOAA for confirming my mistake!


Hallberg-NOAA (approved):
I agree that this halo update is necessary.  However, there is some duplication in that this routine extracts either the u or v component of the surface velocity by name, and not both. If both were extracted simultaneously, there would only need to be one halo update for both, rather than one for each of the two velocity components.

Hallberg-NOAA:
@sanAkel, there have been other commits to dev/gfdl since this was added, so this PR is now out-of-date with the base branch.  However, you do not appear to have given the maintainers the ability to modify this PR.  Please click the "Update Branch" button so we can trigger the pipeline tests and merge this PR into dev/gfdl.

sanAkel:
@Hallberg-NOAA, appreciate the [vector pair suggestion](https://github.com/NOAA-GFDL/MOM6/pull/1356#pullrequestreview-622602230). I intend to make some changes to our _coupler_ (making use of new infrastructure), will certainly include it in that, just wanted to get the bug fix out of my way first! Thanks for the suggestion (also updated the branch, not sure how I would let _you all_ do it for me).

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12350 .

PR #1360, opened 2021-03-28
Remove inappropriate timelevel arguments

Hallberg-NOAA:
Remove timelevel arguments from MOM_read_data calls for variables that should
never have multiple time-levels.  All answers are bitwise identical, and this
will eliminate some warning messages in output.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12382

marshallward (approved):

PR #1361, opened 2021-03-29
Coriolis: Improved coradcalc vectorization

marshallward:
This patch restructures the CorAdCalc function so that the loops are
more easily vectorized on a broader range of systems.  The number of
memory access has also been slightly reduced.

We observed a 1.72x speedup on a modern consumer AMD processor (Ryzen 5
2600) and a 1.27x speedup on Gaea's Intel Xeons (E5-2697 v4).

Description
-----------

Several if-blocks were moved outside of the do-loops, enabling
vectorization.

In one particular case, a conditional if-block was replaced with a
merge() statement.  This was able to signal the compiler to use a
vectorized compare operation in place of jmp instructions.

(I don't generally recommend this however; this is the first time I've
seen a merge() statement have a significant impact).

The other major change, which also reduced the number of memory
accesses, was the combination of the two Coriolis-bound loops into a
single loop.  The intermediate (min|max)_f[uv]q arrays no longer need to
be constructed and stored.

General comments
----------------

I can suggest two possible explainations for the AMD/Intel performance
difference:

* Masking instructions have a lower latency on Intel CPUs, which permit
  limited vectorization of if-blocks.  Similar vectorization is not
  possible on AMD CPUs.  So Intel is less likely to benefit from
  if-block re-ordering.

* Our Intel nodes on Gaea have much lower RAM bandwidth, and see a
  smaller benfit from vectorization, which must required greater
  bandwidth.  This speedup may be greater on a more modern Intel
  platform.

Although the code has been vectorized on both Intel and AMD platforms,
there are still a very large number of memory accesses per operation,
which is still hampering performance.  This needs to be addressed if
further improvements are to be considered.

The changes below are not expected to change any answers, and none were
detected.  But since we are changing a core component, I'd suggest
reviewing this carefully.

Runtime measurements
--------------------

AMD Before:
(Ocean Coriolis & mom advection)      1.074557
(Ocean Coriolis & mom advection)      1.074487
(Ocean Coriolis & mom advection)      1.074909
(Ocean Coriolis & mom advection)      1.078152
(Ocean Coriolis & mom advection)      1.074272

AMD After:
(Ocean Coriolis & mom advection)      0.626817
(Ocean Coriolis & mom advection)      0.625959
(Ocean Coriolis & mom advection)      0.629569
(Ocean Coriolis & mom advection)      0.627651
(Ocean Coriolis & mom advection)      0.631355

Intel Before:
(Ocean Coriolis & mom advection)      0.996460
(Ocean Coriolis & mom advection)      0.998821
(Ocean Coriolis & mom advection)      0.994859
(Ocean Coriolis & mom advection)      0.993714
(Ocean Coriolis & mom advection)      1.002142

Intel After:
(Ocean Coriolis & mom advection)      0.781821
(Ocean Coriolis & mom advection)      0.785724
(Ocean Coriolis & mom advection)      0.786662
(Ocean Coriolis & mom advection)      0.789598
(Ocean Coriolis & mom advection)      0.792768


src/core/MOM_CoriolisAdv.F90
```diff
@@ -428,45 +431,48 @@ subroutine CorAdCalc(u, v, h, uh, vh, CAu, CAv, OBC, AD, G, GV, US, CS)
...
+      abs_vort(I,J) = G%CoriolisBu(I,J) + rel_vort(I,J)
+    enddo ; enddo
 
-      if (CS%id_rv > 0) RV(I,J,k) = relative_vorticity
-      if (CS%id_PV > 0) PV(I,J,k) = q(I,J)
-      if (associated(AD%rv_x_v) .or. associated(AD%rv_x_u)) &
-        q2(I,J) = relative_vorticity * Ih
+    do J=Jsq-1,Jeq+1 ; do I=Isq-1,Ieq+1
+      hArea_q = (hArea_u(I,j) + hArea_u(I,j+1)) + (hArea_v(i,J) + hArea_v(i+1,J))
+      Ih_q(I,J) = merge( &
```
Hallberg-NOAA, line 450: Although merge is a standard F90 intrinsic, it seems obscure to me and does not have a direct functional tie to the English language (unlike where), and is not used elsewhere in the MOM6 code. I would suggest that we add a comment here like "! x = merge(a, b, test) is equivalent to 'if (test) then ; x=a ; else ; x=b ; endif`, to help non-proficient F90 programmers like me to understand what this is doing.  Alternately, should we search through the rest of the code for things like this explicit logical assignment and replace them all with merge() calls, since they are so much faster.

It is very surprising to me that replacing the straightforward explicit logic syntax with a merge is faster because it seems like compilers should be able to see through this! What are compilers good for?

Another curious point here - I would have written an explicit 5-command conditional assignment in 1 line with semicolons, whereas this is a 1-command conditional assignment via a merge written in 5 lines.  To each his own, I guess...

marshallward: I would suggest not replacing isolated if-blocks with `merge` statements at this time.  Although it had a positive effect here, often they do much of anything.

As for the etymology, usually it's used to "merge" two arrays based on a third logical array.  Of course, for a scalar that merge basically just reduces to an if-block.  My guess is that somehow the intention to merge is getting interpreted as a vectorized operation.

marshallward: Replacing `h_neglect * Area_q` with a new `vol_neglect` will allow `Area_q` to act as its own mask, and we can avoid the `merge` altogether.  I will update the patch soon.

marshallward:
Resubmitting as a new PR.

PR #1364, opened 2021-04-01
Fixed downsampling for x:sum y:point z:point diags

jkrasting:
- previous code had averaging instead of summation for
  SPP (x:sum,y:point,z:point) diagnostics
- corrects an issue where these diagnostics were incorrect
  by approximately a factor of 2.
- Orginially found when analyzing the depth-integrated
  temperature advection diagnostic (T_ady_2d)

### DIFF
--- a/src/framework/MOM_diag_mediator.F90
+++ b/src/framework/MOM_diag_mediator.F90
@@ -4181,7 +4181,7 @@ subroutine downsample_field_2d(field_in, field_out, dl, method, mask, diag_cs, d
         total_weight = total_weight +weight
         ave = ave+field_in(ii,jj)*weight
       enddo
-      field_out(i,j) = ave/(total_weight+epsilon)  !Avoid zero mask at all aggregating cells where ave=0.0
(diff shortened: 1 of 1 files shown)


raphaeldussin:
do we need a similar fix for https://github.com/NOAA-GFDL/MOM6/blob/6d4a4839070bc599fe03551b2d6fa6f53fe16584/src/framework/MOM_diag_mediator.F90#L4170 ?

jkrasting:
Not sure. Feel free to update my commit though.
--

Hallberg-NOAA:
I agree that we do also need the other corrections on the lines that @raphaeldussin noted.

While we are at it, for clarity when reading the code we might want to replace the variable "ave" with a new variable "sum" in these cases and eliminate the now-unused variable "total_weight" in these cases.

Hallberg-NOAA (approved):
I agree that these changes correct important bugs in the reduced-grid diagnostic output.

Hallberg-NOAA:
This PR passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12421.

PR #1371, opened 2021-04-12
(*)Write full checksums to restarts

Hallberg-NOAA:
Change the types returned from the 5 rotated_field_chksum functions from integer
to integer(kind=int64), so that the full 64-byte checksums are returned.  Without
this change, the checksums that are written to MOM6 restart files or interpreted
from them are truncated to the latter half of their length.  This changes the
checksums that are written to the restart files, but both before and after this
change the values that are written are the same as those that are generated
after reading the restart with the same version of the code.  The code can run
across this change by setting RESTART_CHECKSUMS_REQUIRED = False for the run
segment where the transition occurs.  The solutions themselves are bitwise
identical.


marshallward (approved):

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12464

PR #1372, opened 2021-04-13
Update of Eady growth rate used in MEKE and Visbeck

adcroft:
The Eady growth rate, SN, as used by MEKE and Visbeck is a vertical average (either of full depth or a prescribed depth) of slope magnitude and Brunt-Väisälä frequency. The calculation always had some artifacts associated with topography but it turned out to have significant artifacts in SWE mode too, discovered in the CPT's NeverWorld2 experiments.

The vertical average of SN used to be calculated by vertically accumulating dz*sqrt(S² * N²), where S² and N² are the squares of slope and Brunt-Väisälä frequency returned from a core routine `calc_isoneutral_slopes()`. N² for vanished layers is problematic due to the `h` in the denominator, especially in SWE mode. In diabatic mode, a vertical diffusive smoothing "fixes" N² but S² still has some "sophisticated" discretization in `calc_isoneutral_slopes()`. These quantities are unchanged since I think they are calculated as appropriately as can be.

A new runtime parameter, `USE_NEW_EADY_GROWTH_RATE`, changes the vertical averaging. Since dz sqrt(S² N²) = sqrt(S² N² dz²) we can avoid the division by `h`: dz² N² = (g/ρ₀) δρ dz = dz g' . So we now pass the x- and y- contributions to sqrt(S²dz g') from `calc_isoneutral_slopes()`. We also pass the actual `h` used in the slope and N² calculations so that, when limiting the depth for averaging of SN, the positions are most consistent.

Before (lower) and after (upper) plots of SN based on the initial conditions in OM_1deg:
image

Not easy to see are the removal of hot spots around submerged topographic features and islands like Hawaii. The removal of the large-scale topographic features (mid-Atlantic ridge) is primarily a result of choosing runtime parameter `EADY_GROWTH_RATE_D_SCALE` to 3500m. The reduced growth rate on coastal shelves is mostly due to the change in discretization above.

Note to @gustavo-marques  - the length scale in Visbeck is still a constant.

(edited following comment from @StephenGriffies . Code and notes are correct, just PR text was wrong)

### DIFF
--- a/src/core/MOM_isopycnal_slopes.F90
+++ b/src/core/MOM_isopycnal_slopes.F90
@@ -84,8 +84,8 @@ subroutine calc_isoneutral_slopes(G, GV, US, h, e, tv, dt_kappa_smooth, &
   real :: wtA, wtB, wtL, wtR  ! Unscaled weights, with various units.
   real :: drdx, drdy    ! Zonal and meridional density gradients [R L-1 ~> kg m-4].
   real :: drdz          ! Vertical density gradient [R Z-1 ~> kg m-4].
-  real :: Slope         ! The slope of density surfaces, calculated in a way
-                        ! that is always between -1 and 1.
+  real :: slope         ! The slope of density surfaces, calculated in a way
+                        ! that is always between -1 and 1. [Z L-1 ~> nondim]
   real :: mag_grad2     ! The squared magnitude of the 3-d density gradient [R2 Z-2 ~> kg2 m-8].
   real :: slope2_Ratio  ! The ratio of the slope squared to slope_max squared.
   real :: h_neglect     ! A thickness that is so small it is usually lost
@@ -184,7 +184,7 @@ subroutine calc_isoneutral_slopes(G, GV, US, h, e, tv, dt_kappa_smooth, &
   !$OMP                          private(drdiA,drdiB,drdkL,drdkR,pres_u,T_u,S_u,      &
   !$OMP                                  drho_dT_u,drho_dS_u,hg2A,hg2B,hg2L,hg2R,haA, &
   !$OMP                                  haB,haL,haR,dzaL,dzaR,wtA,wtB,wtL,wtR,drdz,  &
-  !$OMP                                  drdx,mag_grad2,Slope,slope2_Ratio,l_seg)
+  !$OMP                                  drdx,mag_grad2,slope,slope2_Ratio,l_seg)
   do j=js,je ; do K=nz,2,-1
     if (.not.(use_EOS)) then
       drdiA = 0.0 ; drdiB = 0.0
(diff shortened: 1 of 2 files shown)


StephenGriffies:
dz² N² = (g/ρ₀) δρ = g'  is a wrong equation.  Check the dimensions.  This equation should instead read dz N² = (g/ρ₀) δρ = g'.  If the code reflects the original wrong equation, then there is an error in the new code.

adcroft:
> dz² N² = (g/ρ₀) δρ = g' is a wrong equation. Check the dimensions. This equation should instead read dz N² = (g/ρ₀) δρ = g'. If the code reflects the original wrong equation, then there is an error in the new code.

Indeed. Original PR text was incorrect - code (and my notes that led to code) was right.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12471

src/core/MOM_isopycnal_slopes.F90
```diff
@@ -45,6 +46,16 @@ subroutine calc_isoneutral_slopes(G, GV, US, h, e, tv, dt_kappa_smooth, &
...
                                      optional, intent(inout) :: N2_v !< Brunt-Vaisala frequency squared at
                                                                      !! interfaces between v-points [L2 Z-2 T-2 ~> s-2]
+  real, dimension(SZIB_(G),SZJ_(G),SZK_(GV)+1), &
+                                     optional, intent(inout) :: dzu  !< Z-thickness at u-points [Z ~> m]
+  real, dimension(SZI_(G),SZJB_(G),SZK_(GV)+1), &
+                                     optional, intent(inout) :: dzv  !< Z-thickness at v-points [Z ~> m]
+  real, dimension(SZIB_(G),SZJ_(G),SZK_(GV)+1), &
+                                     optional, intent(inout) :: dzSxN !< Z-thickness times zonal slope contribution
+                                                                     !! to Eady growth rate at u-pints. [Z T-1 -> m s-1]
+  real, dimension(SZI_(G),SZJB_(G),SZK_(GV)+1), &
```
Hallberg-NOAA, line 56: I think that someone had been thinking about hitting the pub a bit too much when describing these variables in terms of u-pints and v-pints...

src/parameterizations/lateral/MOM_lateral_mixing_coeffs.F90
```diff
@@ -1089,10 +1294,27 @@ subroutine VarMix_init(Time, G, GV, US, param_file, diag, CS)
        'Inverse eddy time-scale, S*N, at u-points', 's-1', conversion=US%s_to_T)
     CS%id_SN_v = register_diag_field('ocean_model', 'SN_v', diag%axesCv1, Time, &
        'Inverse eddy time-scale, S*N, at v-points', 's-1', conversion=US%s_to_T)
-    call get_param(param_file, mdl, "VARMIX_KTOP", CS%VarMix_Ktop, &
-                 "The layer number at which to start vertical integration "//&
-                 "of S*N for purposes of finding the Eady growth rate.", &
-                 units="nondim", default=2)
+    call get_param(param_file, mdl, "USE_NEW_EADY_GROWTH_RATE", CS%use_new_Eady_growth_rate, &
```
Hallberg-NOAA, line 1297: "New" is an adjective that might be best avoided in a name, since something that is new now won't be new in a few months or years.  What would you think about calling this something like "SIMPLER_EADY_GROWTH_RATE" instead?

Hallberg-NOAA (commented):
I understand what these changes are doing, and I agree with them. However, upon a careful visual inspection of the code I have a couple of minor comments that probably could be addressed before this PR is merged in.

src/parameterizations/lateral/MOM_lateral_mixing_coeffs.F90
adcroft: What if we come up with an even simpler approach? 😁  `SIMPLER_...` it is.

src/core/MOM_isopycnal_slopes.F90
adcroft: Comments fixed in e8e200994f04ce92c93900378d7f8878a3c3339c

src/parameterizations/lateral/MOM_lateral_mixing_coeffs.F90
adcroft: Parameter name changed in d21d21acc78308143836df2ddaf22f99e6485a52

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12506

Hallberg-NOAA (approved):
I am satisfied that this PR is correct and helpful, but because this PR includes new run-time parameters, it will have to be handled manually.

PR #1373, opened 2021-04-15
+Make MOM_read_data work when 4-d arrays exist

Hallberg-NOAA:
Added internal branches of the infra/FMS1 version of MOM_read_data that work
when there are 4-d arrays with a fifth time dimension in the file that is being
read, which was previously failing even if the variable that is being read has
fewer dimensions, due to code limitations within the fms_io read_data routines.
These new branches are selected via the new optional argument file_may_be_4d.
The infra/FMS2 versions also have the same new optional argument, although in
that case it does nothing.  All answers are bitwise identical in the cases that
worked previously, but SIS2 restart files can now be read via the MOM6
infrastructure interfaces.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12518

marshallward:
There was some discussion about whether `file_may_be_4d` was too specific for a generic API and could confuse contributors, but we agreed that it helps to solve a tricky FMS transition problem without much code duplication, and that there will be a later opportunity to revise the infra API after the FMS2 transition is complete.

marshallward (approved):

Issue #1312, opened 2021-01-29
Avoid direct use of NetCDF outside of MOM_io.F90


Hallberg-NOAA:
There are currently a handful of MOM6 files outside of MOM_io.F90 and the framework directory that use direct calls to NetCDF. This is strongly discouraged, as it can lead to significant code duplication, and a loss of efficiency for large processor count jobs if all of the processors are trying to simultaneously read from the same file, rather than having the root PE read data and then broadcast it. Also, as we consider migrating to different modes for doing I/O, such as having a different set of processors doing diagnostics I/O from those that are doing the calculations themselves, having widely distributed I/O calls greatly complicates the testing, development and transition to any such new I/O paradigms.

  Going forward, NetCDF calls should only occur within MOM_io.F90, and the existing MOM6 code will be adapted to change its direct NetCDF calls into calls to routines in MOM_io.F90.

Hallberg-NOAA:
This issue has finally been fully addressed by https://github.com/NOAA-GFDL/MOM6/pull/48.

PR #1374, opened 2021-04-16
User/wfc/hor grid ans2018

wfcooke:
Fix for use of optional argument without checking for presence.
    
The optional argument answers_2018 is used in calls to fill_miss_2d in two places but it may not be present to begin with.
As the default in fill_miss_2d is to have a true value if that argument is not present, I have changed the pass through argument to ans_2018.
    
Also needed to add ans_2018 and related logic to the horiz_interp_and_extrap_tracer_record subroutine.

Additionally.
IMO, the horiz_interp_and_extrap_tracer_record subrotine should probably be removed as there are naked NetCDF calls in there.


Hallberg-NOAA (commented):
I do not understand why this commit is necessary. As I understand it, Fortran allows an optional argument to be used as an optional argument in a call to a second routine, even if it is not present in the call to the first routine.  In practice, passing a non-present argument as an argument is the same as not passing the argument.  So I don't think that this PR does anything at all.

  If there is a functional reason why this is necessary, an explanation would be appreciated, because this construct is used in many places in the MOM6 code.  If this code construct is demonstrably problematic, we would need to revise the code in all the places where this is used.

wfcooke:
@Hallberg-NOAA I agree that it shouldn't make any difference. 
It is the only place in this routine where answers_2018 is used, or ans_2018 would be, so why bother assigning a value to ans_2018 if you are passing through the argument? 
It is just something I spotted and thought should be made consistent with other pieces of code (check an incoming optional, assign a local variable a value, and then use the local)
I don't have any evidence of it causing a problem and it definitely would be a compiler bug if it becomes one.

Would you prefer I take out the ans_2018 logic altogether as it's not being used in that routine?

Hallberg-NOAA:
@wfcooke, we appreciate your diligence in trying to identify and correct odd code constructs. However, in light of the discussion surrounding this PR, I am going to suggest that this PR should be closed, as these changes do not appear to be necessary.  If you agree, @wfcooke, please close this PR. 

  I do agree that the direct calls to NetCDF routines should be encapsulated in the MOM_io routines, as is already noted in MOM6 issue #1312 .

PR #1376, opened 2021-04-19
Display hash of submodules in gitlab pipeline

adcroft:
- This adds a few git commands at the end of the "setup" job in the gitlab pipeline to show the hash of every submodule we care about.
- We recently found it hard to retroactively figure out the exact state of the pipeline because it updates to the latest of MOM6-examples at the time of running, and that might change between the run and when you examine the logs.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12519

marshallward (approved):

PR #1380, opened 2021-04-20
read_attribute_str returns cleanly when found=F

### LINKED ISSUE #1375: read_attribute() doesn't behave quite the same as code it replaced in MOM_tracer_Z_init
I'm finally merging some recent commits to a branch where I'm introducing a new tracer module, and am having problems with this change in `src/tracer/MOM_tracer_Z_init.F90:read_Z_edges()` from [`44a8bde`](https://github.com/NOAA-GFDL/MOM6/commit/44a8bde1c4ddcd1cdfbabcfe3c932a6cc966b791#diff-925d91e0eeb0a776c258e755e51f18de30ca3f6a2a360cc4b9f9dcf3a9eb2e4dR418)

```diff
-  status = NF90_GET_ATT(ncid, layid, "edges", edge_name)
-  if (status /= NF90_NOERR) then
-    call MOM_mesg(mdl//" "//trim(dim_msg)//&
-         " has no readable edges attribute - "//trim(NF90_STRERROR(status)))
-    has_edges = .false.
-  else
-    has_edges = .true.
-    status = NF90_INQ_VARID(ncid, edge_name, intid)
-    if (status /= NF90_NOERR) then
-      call MOM_error(WARNING,mdl//" Difficulties finding edge variable "//&
-          trim(edge_name)//" in "//trim(filename)//" - "//trim(NF90_STRERROR(status)))
-      has_edges = .false.
-    endif
-  endif
+  call read_attribute(filename, "edges", edge_name, varname=dim_names(3), found=has_edges)
```

My initial condition file does not have an `edges` attribute for `DEPTH`, and it is triggering the `Attribute data type is not a char` error from [`src/framework/MOM_io.F90:read_attribute_str()`](https://github.com/NOAA-GFDL/MOM6/blob/dev/gfdl/src/framework/MOM_io.F90#L912) Since I'm passing in the optional `found` argument, I would expect it to return `.false.` rather than abort.

It's late (and Friday, at least in this time zone) so I'm not going to propose a fix right now... I'll probably come back to this mid-week, but won't be offended if somebody beats me to a patch.

mnlevy1981:
Updates `read_attribute_str()` so that if the optional argument found is provided
and the attribute is not defined, then the function returns with `found=.false.`
instead of calling `MOM_error(FATAL)`.

I did a little bit of testing, and it looks like `read_attribute_real` and
`read_attribute_int` (not sure if the default is int32 or int64) already behave
correctly when `found` is present so they do not need a similar update

Fixes #1375

### DIFF
--- a/src/framework/MOM_io.F90
+++ b/src/framework/MOM_io.F90
@@ -906,18 +906,20 @@ subroutine read_attribute_str(filename, attname, att_val, varname, found, all_re
     endif
     if ((varid > 0) .or. (varid == NF90_GLOBAL)) then ! The named variable does exist, and found would be true.
       rc = NF90_inquire_attribute(ncid, varid, attname, xtype=att_type, len=att_len)
-      if ((rc /= NF90_NOERR) .and. (rc /= NF90_ENOTATT)) &
-        call MOM_error(FATAL, trim(hdr) // trim(NF90_STRERROR(rc)) //" Error getting info for "//trim(att_str))
-      if (att_type /= NF90_CHAR) &
-        call MOM_error(FATAL, trim(hdr)//": Attribute data type is not a char for "//trim(att_str))
-!      if (att_len > len(att_val)) &
-!        call MOM_error(FATAL, trim(hdr)//": Insufficiently long string passed in to read "//trim(att_str))
-      allocate(character(att_len) :: att_val)
-
-      if (rc == NF90_NOERR) then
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA (approved):
I have reviewed the logic here, and I am convinced that this is correct and that is should neatly correct the problem noted in issue #1375.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12520

PR #1383, opened 2021-04-26
Adapts Phillips case to work in doubly periodic mode

adcroft:
This does not change the current Phillips test case but by setting `REENTRANT_Y=True` this code mod changes the non-periodic tanh(y) forcing functions to the periodic sin(y) yields a valid GFD configuration for looking at f-plane tubrulence without boundaries. Used to make some eye-candy at https://youtu.be/buI6mKzkbs0.


Hallberg-NOAA (approved):
Brilliant!

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12556

Issue #1384, opened 2021-04-27
Inconsistent units and conversion factors in numerous diagnostics


Hallberg-NOAA:
Prompted by Issue #1017, I decided to survey all of the conversion factors that are used in the calls to register_diag_field, and found that there were at least 17 places where these are inconsistent with the documented units.  With 17 instances out of 715 calls, this is about a 2.5% error rate.

  I will be putting in a pull request to correct these inconsistencies, but I would also like to encourage people to put the units and conversion factor arguments for all register_diag_field calls on the same line to facilitate the detection of inconsistencies.  We do test that the conversion factors are correct with our dimensional consistency testing, but this examination for consistency seems to be the most effective way of scanning through all of our diagnostics to detect these inconsistencies.

Hallberg-NOAA:
This issue has been corrected for now, which can be verified with some relatively simple greps, but it will have to be reexamined periodically in the future, and it should be something that is examined with all new commits that add diagnostics.

marshallward:
I think we could integrate the cited grep command into `trailer.py`-like script without too much trouble if there's a demand.

PR #1385, opened 2021-04-27
(*)Correct inconsistent units in 17 diagnostics

Hallberg-NOAA:
Corrected inconsistent units or unit conversion factors for 17 diagnostics
plus all of the diagnostics for extra tracer fluxes where the scaling factor is
not explicitly supplied (temperature and salinity are not impacted).  In most
cases the corrections are in obscure diagnostics that are probably not used much
or only apply to non-Boussinesq cases.  In addition, dimensional rescaling in
time was applied to the diapycnal velocity diagnostic in layered mode.  This
commit addresses the specific problems noted in MOM6 issue #1384, but it does
not uniformly change to the suggested code construct for register_diag_field
arguments that would make these problems easier to detect in the future.  All
solutions are bitwise identical, but there are changes to the metadata of some
diagnostics and there are changes to the available_diags files.


marshallward:
https://github.com/NOAA-GFDL/MOM6/pull/1385

marshallward (approved):

PR #1386, opened 2021-04-28
.testing: Create python venv at compile time

marshallward:
This patch shifts the creation of the python virtual environment to a
compile-time operation.

The .testing suite creates a Python virtual environment (venv) if the
current environment does not include numpy and the Python netCDF4
modules.

These are detected and set up at runtime rather than compile time, which
can cause issues when run on compile nodes which do not have internet
access since the packages need to be downloaded via Pip.

This patch moves the detection and creation to compile time.  Since
internet access is expected for FMS and mkmf setup, it is consistent to
assume internet access for pip installations.

### DIFF
--- a/.testing/Makefile
+++ b/.testing/Makefile
@@ -193,7 +193,7 @@ endif
 # Rules
 
 .PHONY: all build.regressions
-all: $(foreach b,$(BUILDS),build/$(b)/MOM6)
+all: $(foreach b,$(BUILDS),build/$(b)/MOM6) $(VENV_PATH)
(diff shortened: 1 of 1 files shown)


marshallward:
Not that this does not address the cross-platform issue, where the compile-platform (e.g. login node) has a fully functional Python environment but the runtime-platform (e.g. compute node) does not.

Hallberg-NOAA (approved):
I have tested this out manually, and I am now able to run all of the TC test cases, including tc4, successfully, for the first time ever!

adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12573

PR #1388, opened 2021-04-30
Use MOM_io in Surface_Bands_by_data_override

Hallberg-NOAA:
Revised Surface_Bands_by_data_override in MOM_wave_interface to use the
interfaces from MOM_io instead of direct calls to NF90_ routines to read the
coordinate frequency or wavenumber.  This change will be more efficient by
having only the root processor read the files and then broadcast the information
to the other PEs, and it will be more robust by reusing common code used by
other MOM6 routines.  This commit partially addresses MOM6 issue #1312.  All
answers are bitwise identical.


breichl:
Thanks @Hallberg-NOAA .  At a glance this looks like a good rewriting of the NetCDF interfaces.  I'm going to try to piece together a short test case to verify all the behavior (there unfortunately isn't any way to test in the current MOM6-examples repo).  This may take a few days, but hopefully by early next week.  Might be worth merging that test back into MOM6-examples, probably a light-weight single column example.

breichl (approved):
These changes now work in my test case.  I approve.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12660

marshallward (approved):

PR #1389, opened 2021-04-30
Use MOM_io interfaces in write_depth_list

Hallberg-NOAA:
Revised write_depth_list to use the interfaces from MOM_io instead of direct
calls to NF90_ routines.  The depth list files that are generated this way are
interchangeable with those from previous versions, but they are larger because
FMS insists on including an unnecessary axis variable with a simple list of
integers for the list axis.  This commit partially addresses MOM6 issue #1312.
All answers are bitwise identical.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12599

marshallward (approved):

PR #1390, opened 2021-05-03
Collocate units and conversion for register_diag_field calls

Hallberg-NOAA:
Rearranged arguments in register_diag_field calls to facilitate the detection
of inconsistent units and scaling factors using simple methods, by placing the
unit and conversion arguments on the same line.  At some point, we will develop
more sophisticated methods for detecting such inconsistencies, at which these
changes will lose their value but become benign.  Also, made use of the scaling
factor GV%H_to_MKS, which handles conversion of thickness units to m or kg m-2,
depending on whether the code is being run in Boussinesq or non-Boussinesq mode,
instead of having logic scattered in several places throughout the code to
select GV%H_to_m or GV%H_to_kg_m2.  This commit closes MOM6 issue #1384.  All
answers and output are bitwise identical, and these are mostly essentially
whitespace changes.

### DIFF
--- a/src/ALE/MOM_ALE.F90
+++ b/src/ALE/MOM_ALE.F90
@@ -263,8 +263,8 @@ subroutine ALE_register_diags(Time, G, GV, US, diag, CS)
   CS%id_v_preale = register_diag_field('ocean_model', 'v_preale', diag%axesCvL, Time, &
       'Meridional velocity before remapping', 'm s-1', conversion=US%L_T_to_m_s)
   CS%id_h_preale = register_diag_field('ocean_model', 'h_preale', diag%axesTL, Time, &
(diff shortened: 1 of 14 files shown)


marshallward:
The following grep seems capable of printing `register_diag_field` calls which contain a `conversion` argument:

    grep -rPazo 'register_diag_field *\((.|(?<!\))\n)*conversion *=(.*?)\)\n' src/

It assumes lines end with `)\n`.  It can probably be extended to handle comments.

marshallward (approved):
While generally one should avoid coding to the tool (in this case `grep`), I agree that this is a benign change that will make it easier to detect unit consistency.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12708

PR #1392, opened 2021-05-05
infra, FMS_cap: stdout_if_root()

marshallward:
This patch adds `get_root_stdout` to the MOM_io API, and points to the
FMS `stdout()` function in the FMS infra implementations.

This change was required because calls to `coupler_type_write_chksums`
handles both its own checksums across ranks and its own IO to `outunit`.
Typically only the root PE will write the result.

The FMS `stdout()` function would return the designated stdout unit for
the root PE, and the internal `etc_unit` for other PEs, usually set to
`/dev/null`.

When MOM_io switched from using the FMS `stdout()` function to the
`stdout` unit as defined in `iso_fortran_env`, this functionality was
lost and every PE would write the same result to stdout.

Normally this was controlled with `if (root)`-like tests, but this
cannot be used in functions like `coupler_type_write_checksums`, which
require participation of all PEs.

We decided the only resolution here was to introduce a new function,
`get_root_stdout` which replicates the original behavior of `stdout()`.

Additional comments:

* The `if (root)` checks were retained, since it is presumably still a
  good idea to avoid the `write()` calls when possible, even to
  `/dev/null`.

* This patch only updates the FMS coulper driver, but I would recommend
  that the other driver maintainers review their own calls to this
  function.

### DIFF
--- a/config_src/drivers/FMS_cap/MOM_surface_forcing_gfdl.F90
+++ b/config_src/drivers/FMS_cap/MOM_surface_forcing_gfdl.F90
@@ -27,7 +27,8 @@ use MOM_get_input,        only : Get_MOM_Input, directories
 use MOM_grid,             only : ocean_grid_type
 use MOM_interpolate,      only : init_external_field, time_interp_external
 use MOM_interpolate,      only : time_interp_external_init
-use MOM_io,               only : slasher, write_version_number, MOM_read_data, stdout
+use MOM_io,               only : slasher, write_version_number, MOM_read_data
+use MOM_io,               only : stdout_if_root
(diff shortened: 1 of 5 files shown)


marshallward:
This needs to be reviewed by @nikizadehgfdl before being merged.

marshallward:
PR was updated to rename `get_root_stdout` to `stdout_if_root`.  Also fixed a bug in FMS2 infra (function was not declared public).

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12667.

Hallberg-NOAA (approved):
These changes seem sensible enough and should address the problems of excessive checksum output.

Issue #1391, opened 2021-05-03
Which layer thicknesses to use for thickness-weighted momentum diagnostics and vorticity budgets


hmkhatri:
@StephenGriffies  and I have been using depth-integrated 2D momentum diagnostics for vorticity budget analysis in global MOM6 simulations and we observed that some analytical expressions in vorticity budgets do not agree when compared against these diagnostics. This is probably due to the numerical treatment of individual terms in the model. 

An important issue is - which layer thicknesses to use for thickness-weighted momentum diagnostics such that the discrepency between the online diagnostics and analytical expressions is minimal. This would make it much easier to analyze various budget terms from the model output and compare them against theories.

I added 2D momentum diagnostics last year and, in the current implementation of 2D momentum diagnostics, the layer thicknesses at u/v points (`hu`/`hv`) are returned in the `btstep` call as part of the `ADp` control structure.  

https://github.com/NOAA-GFDL/MOM6/blob/2912058b17b5e8636c8687ed236ac68aba36f609/src/core/MOM_dynamics_split_RK2.F90#L761-L764

https://github.com/NOAA-GFDL/MOM6/blob/2912058b17b5e8636c8687ed236ac68aba36f609/src/core/MOM_barotropic.F90#L2678-L2687

With this choice of `h`, some terms look unrealistically large. For example, consider the following analytical expression.

<img src="https://user-images.githubusercontent.com/17830036/116926070-fc358500-ac27-11eb-9091-5b27a6540b3d.png" height="60">

Using the new 2D momentum diagnotics, one can compute the LHS of the equation above as
`d/dx [ intz_CAv_2d - intz_rvxu_2d - intz_gKEv_2d ] - d/dy [intz_CAu_2d - intz_rvxv_2d - intz_gKEu_2d ]`  

and this should roughly be equal to `- beta x V` (others terms are very small), which one can compute independently using `vmo_2d / vmo` diagnostic. As seen in the figure below, fields from the two calculations differ by more than an order of magnitude. 

![image](https://user-images.githubusercontent.com/17830036/116928334-ed040680-ac2a-11eb-95d7-ffa22ea54d62.png)![image](https://user-images.githubusercontent.com/17830036/116928695-656ac780-ac2b-11eb-8f74-8f1495b41055.png)

Thus, it is critical to use thicknesses consistent with numerics so that thickness-weighted diagnostics result in correct magnitudes of terms in higher-order budgets. 

[...]

PR #1397, opened 2021-05-10
3D Thickness multiplied momentum diagnostics

hmkhatri:
In this PR, 3D thickness multiplied momentum equation diagnostics have been implemented. @NoraLoose @gustavo-marques are using these diagnostics for thickness-weighted analysis in the Ocean-CPT work.

Following suggestions made by @marshallward in PR #1343 for 2D diagnostics, 3D diagnostic arrays are also declared as stack arrays. These arrays should consume memory only if the relevant diagnostics are called. 

Also, it would be useful to discuss which layer thicknesses are appropriate to be used in these diagnostics. In the current implementation, the layer thicknesses at u/v points (`diag_hu` / `diag_hv`) are returned in the `btstep` call as part of the `ADp` control structure. There may be better ways for choosing layer thicknesses consistent with numerics.

https://github.com/NOAA-GFDL/MOM6/blob/2912058b17b5e8636c8687ed236ac68aba36f609/src/core/MOM_dynamics_split_RK2.F90#L761-L764

I have already put a separate issue #1391 on this topic.


gustavo-marques:
@hmkhatri, thank you very much for this PR. I see many commits from Alistair that related to "NW2 tracers updates". However, I do not see the corresponding code changes being added here. In any case, I suggest that you create a new branch with a **single** commit adding **only** the 3D thickness multiplied momentum diagnostics. This would greatly simplify the process of tracking code changes and also of propagating this PR to the other MOM6 forks. Please let me know if you need help with this.

hmkhatri:
@gustavo-marques Not sure why all commits by Alistair are appearing here. Could be that I used `reset` command for files to resolve conflicts from dev/gfdl branch but old commit descriptions are still appearing. In any case, I will create a new branch with these 3D diagnostics and submit a new PR.

marshallward:
@hmkhatri I know this issue is closd, but I would suggest *not* using stack for the 3D arrays.  They are large enough where it could start to cause stack overflows.

I'd suggset we allocate any 3D arrays until we have a better strategy for handling scratch arrays for diagnostics.

PR #1398, opened 2021-05-11
3D thickness x momentum diagnostics

hmkhatri:
In this PR, 3D thickness multiplied momentum equation diagnostics have been implemented. @NoraLoose @gustavo-marques are using these diagnostics for thickness-weighted analysis in the Ocean-CPT work.

Following suggestions made by @gustavo-marques, 3D diagnostic arrays are also declared as allocatable arrays. 

Also, it would be useful to discuss which layer thicknesses are appropriate to be used in these diagnostics. In the current implementation, the layer thicknesses at u/v points (`diag_hu` / `diag_hv`) are returned in the `btstep` call as part of the `ADp` control structure. 

https://github.com/NOAA-GFDL/MOM6/blob/2912058b17b5e8636c8687ed236ac68aba36f609/src/core/MOM_dynamics_split_RK2.F90#L761-L764

There may be better ways for choosing layer thicknesses that are consistent with numerics. It would be nice to discuss this point. I have already put a separate issue #1391 on this topic.

### DIFF
--- a/src/core/MOM_CoriolisAdv.F90
+++ b/src/core/MOM_CoriolisAdv.F90
@@ -233,8 +233,8 @@ subroutine CorAdCalc(u, v, h, uh, vh, CAu, CAv, OBC, AD, G, GV, US, CS)
   ! The code is retained for degugging purposes in the future.
 
 ! Diagnostics for thickness multiplied momentum budget terms
-  real, allocatable, dimension(:,:,:) :: h_gKEu, h_rvxv ! [L2 T-2 ~> m2 s-2].
-  real, allocatable, dimension(:,:,:) :: h_gKEv, h_rvxu ! [L2 T-2 ~> m2 s-2].
+  real, allocatable, dimension(:,:,:) :: h_gKEu, h_rvxv ! [H L T-2 ~> m2 s-2].
+  real, allocatable, dimension(:,:,:) :: h_gKEv, h_rvxu ! [H L T-2 ~> m2 s-2].
 
 ! Diagnostics for depth-integrated momentum budget terms
   real, dimension(SZIB_(G),SZJ_(G)) :: intz_gKEu_2d, intz_rvxv_2d ! [L2 T-2 ~> m2 s-2].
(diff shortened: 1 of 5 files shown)


marshallward:
I did not detect any runtime changes from these new diagnostics, so it should be OK to merge them.

src/core/MOM_CoriolisAdv.F90
```diff
@@ -230,6 +232,10 @@ subroutine CorAdCalc(u, v, h, uh, vh, CAu, CAv, OBC, AD, G, GV, US, CS)
   ! 3D diagnostics hf_gKEu etc. are commented because there is no clarity on proper remapping grid option.
   ! The code is retained for degugging purposes in the future.
 
+! Diagnostics for thickness multiplied momentum budget terms
+  real, allocatable, dimension(:,:,:) :: h_gKEu, h_rvxv ! [L2 T-2 ~> m2 s-2].
```
Hallberg-NOAA, line 236: The documented units of these diagnostics are incorrect.  They should be [Z L T-2 ~> m2 s-2].

src/core/MOM_dynamics_split_RK2.F90
```diff
@@ -339,12 +342,15 @@ subroutine step_MOM_dyn_split_RK2(u, v, h, tv, visc, Time_local, dt, forces, p_s
     hf_CAu_2d, hf_CAv_2d, & ! Depth integeral of hf_CAu, hf_CAv [L T-2 ~> m s-2].
     hf_u_BT_accel_2d, hf_v_BT_accel_2d ! Depth integeral of hf_u_BT_accel, hf_v_BT_accel
 
+  real, allocatable, dimension(:,:,:) :: h_PFu, h_CAu, h_u_BT_accel ! [L2 T-2 ~> m2 s-2].
```
Hallberg-NOAA, line 345: The units in the comments for these diagnostics are incorrect.  They should be [H L T-2 ~> m2 s-2 or kg m-1 s-2]

src/core/MOM_CoriolisAdv.F90
hmkhatri: Thanks. I have made the corrections.

src/core/MOM_dynamics_split_RK2.F90
hmkhatri: Thanks. I have made the corrections.

marshallward (approved):
Approving on behalf of @Hallberg-NOAA .

MOM6-examples will require a parameter update.

marshallward:
Regression with updated parameters: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12705

PR #1399, opened 2021-05-11
Avoids issue with unassociated axes in d-sample code

adcroft:
- Cell methods were being passed an nullified pointer for the axes because the axes variable was re-used where it was meant to be
  pointing to the parent variable axes. Adding a copy, paxes, to save from prior to the d-sample block seems to avoid the problem.

@raphaeldussin Can you try this and see if you can make progress. We think there are other issues in the int-wave code but this fix is likely the one that was stopping you from debugging the rest.


marshallward:
I found an issue where non-grid axes were pointing to input axes residing in stack, which would then point to garbage after the function exit.

I will prepare a modification to this PR which fixes that issue.  It resolved things for me, will hopefully fix your issue as well.

raphaeldussin:
I confirm this PR fixes the issue I encountered

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12659

marshallward (approved):

PR #1400, opened 2021-05-12
Diag_mediator: Fix non-standard axis support

marshallward:
When a diagnostic is registered, its axis group is compared to a
standard set of groups along the model grid.  If it matches one of them,
the diagnostic is associated with this axis group.

When it does not match any of the axes, the intention was to retain the
pointer to the input axis group.  However, there was a bug where the
axis group was pointed to the input stack value within the function,
rather than the original axis group.  This caused the diagnostic to be
associated with an axis group which was largely nonsense.

This patch fixes this bug by allocating the new axis group inside the
function to heap, and then copying the contents of the input axis group
to the new allocated group.  This ensures a permanent reference to the
new axis group.

It also probably creates a minor memory leak, but we can sort that out
as part of the overall memory cleanup PR (currently in preparation).

There was literally only one instance of a new axis, namely the use of
angles axes in the internal tide diagnostics.  This patch fixes any runs
which use these diagnostics.

A minor bug associated with logging and the accidental clobbering of the
axes by the desampled (d2) axes was also fixed.  We now call the logging
functions before constructing the d2 axes.

### DIFF
--- a/src/framework/MOM_diag_mediator.F90
+++ b/src/framework/MOM_diag_mediator.F90
@@ -1051,6 +1051,7 @@ subroutine define_axes_group(diag_cs, handles, axes, nz, vertical_coordinate_num
   else
     axes%v_cell_method = ''
   endif
+
   if (present(nz)) axes%nz = nz
   if (present(vertical_coordinate_number)) axes%vertical_coordinate_number = vertical_coordinate_number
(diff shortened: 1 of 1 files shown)


marshallward:
It was pointed out by @adcroft that this change will mangle the d2 entries of the `available_diags` file.  I will close this and re-submit a PR based on #1399

PR #1402, opened 2021-05-13
Add variable and file names to an error message

Hallberg-NOAA:
Made an error message in horiz_interp_and_extrap_tracer_record more
informative and less cryptic by adding the variable and file names being sought.
Also corrected or clarified other error messages in the same routine.  All
answers are bitwise identical, and no interface change.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12650

marshallward (approved):

PR #1404, opened 2021-05-19
+Move wave module variables into control structure

Hallberg-NOAA:
Moved 15 module variables into wave_parameters_CS, to follow the coding style
and conventions in the rest of the code.  Also added comments describing 8
enumeration parameters in this same module, and added comments suggesting ways
that the accuracy of the calculations could be improved.  Also incorporated a
factor of 2*PI into the definition of Freq_Cen in this module.  In addition, the
initialization procedure for this module was altered in some more commonly used
cases, so that use MOM_wave_interface_init to set wave parameters, even with a
statistical wave model, and eliminated MOM_wave_interface_init_lite.  I believe
that all solutions should be bitwise identical, but this module is not yet well
tested in the MOM6 regression tests.  The commits in this PR include:

- NOAA-GFDL/MOM6@89447f49e Merge branch 'dev/gfdl' into MOM_wave_control
- NOAA-GFDL/MOM6@38a897e05 Merge branch 'MOM_wave_rescale' into MOM_wave_control
- NOAA-GFDL/MOM6@b2af54d8a +Move wave module variables into control structure

### DIFF
--- a/config_src/drivers/FMS_cap/ocean_model_MOM.F90
+++ b/config_src/drivers/FMS_cap/ocean_model_MOM.F90
@@ -55,7 +55,7 @@ use MOM_verticalGrid, only : verticalGrid_type
 use MOM_ice_shelf, only : initialize_ice_shelf, shelf_calc_flux, ice_shelf_CS
 use MOM_ice_shelf, only : add_shelf_forces, ice_shelf_end, ice_shelf_save_restart
 use MOM_wave_interface, only: wave_parameters_CS, MOM_wave_interface_init
-use MOM_wave_interface, only: MOM_wave_interface_init_lite, Update_Surface_Waves
+use MOM_wave_interface, only: Update_Surface_Waves
 use iso_fortran_env, only : int64
 
 #include <MOM_memory.h>
@@ -205,7 +205,7 @@ type, public :: ocean_state_type ; private
     marine_ice_CSp => NULL()  !< A pointer to the control structure for the
                               !! marine ice effects module.
   type(wave_parameters_cs), pointer :: &
-    Waves !< A structure containing pointers to the surface wave fields
+    Waves => NULL()           !< A pointer to the surface wave control structure
   type(surface_forcing_CS), pointer :: &
     forcing_CSp => NULL()     !< A pointer to the MOM forcing control structure
   type(MOM_restart_CS), pointer :: &
@@ -382,11 +382,9 @@ subroutine ocean_model_init(Ocean_sfc, OS, Time_init, Time_in, wind_stagger, gas
 
   call get_param(param_file, mdl, "USE_WAVES", OS%Use_Waves, &
        "If true, enables surface wave modules.", default=.false.)
-  if (OS%use_waves) then
-    call MOM_wave_interface_init(OS%Time, OS%grid, OS%GV, OS%US, param_file, OS%Waves, OS%diag)
-  else
-    call MOM_wave_interface_init_lite(param_file)
-  endif
+  ! MOM_wave_interface_init is called regardless of the value of USE_WAVES because
+  ! it also initCSializes statistical waves.
+  call MOM_wave_interface_init(OS%Time, OS%grid, OS%GV, OS%US, param_file, OS%Waves, OS%diag)
 
   call initialize_ocean_public_type(OS%grid%Domain, Ocean_sfc, OS%diag, &
                                     gas_fields_ocn=gas_fields_ocn)
--- a/config_src/drivers/FMS_cap/ocean_model_MOM.F90
+++ b/config_src/drivers/FMS_cap/ocean_model_MOM.F90
@@ -383,7 +383,7 @@ subroutine ocean_model_init(Ocean_sfc, OS, Time_init, Time_in, wind_stagger, gas
   call get_param(param_file, mdl, "USE_WAVES", OS%Use_Waves, &
        "If true, enables surface wave modules.", default=.false.)
   ! MOM_wave_interface_init is called regardless of the value of USE_WAVES because
-  ! it also initCSializes statistical waves.
+  ! it also initializes statistical waves.
   call MOM_wave_interface_init(OS%Time, OS%grid, OS%GV, OS%US, param_file, OS%Waves, OS%diag)
 
   call initialize_ocean_public_type(OS%grid%Domain, Ocean_sfc, OS%diag, &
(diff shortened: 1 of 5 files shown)


config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90
```diff
@@ -388,13 +388,14 @@ subroutine ocean_model_init(Ocean_sfc, OS, Time_init, Time_in, gas_fields_ocn, i
...
-    call MOM_wave_interface_init(OS%Time, OS%grid, OS%GV, OS%US, param_file, OS%Waves, OS%diag)
-    call get_param(param_file,mdl,"SURFBAND_WAVENUMBERS",OS%Waves%WaveNum_Cen, &
-           "Central wavenumbers for surface Stokes drift bands.",units='rad/m', &
-           default=0.12566)
-  else
-    call MOM_wave_interface_init_lite(param_file)
+    ! I do not know why this is being set here.  It seems out of place. -RWH
+    call get_param(param_file,mdl,"SURFBAND_WAVENUMBERS", OS%Waves%WaveNum_Cen, &
+           "Central wavenumbers for surface Stokes drift bands.", &
+           units='rad/m', default=0.12566, scale=OS%US%Z_to_m)
```
breichl, line 398: In my prototype implementation of the coupling with FMS I have these passed through the coupler, and not read from the input file (to ensure consistency with what is inside of a wave model).  Hence why in the coupler option of MOM_wave_interface (L311 of the code before these modifications) we only read the number of bands and allocate arrays in the "coupler" option, but don't read this field.  Presumably this is a temporary work around for nuopc to avoid breaking my codes until these features are more concretely established on our end, though @JessicaMeixner-NOAA may elaborate further if there are other reasons.

src/user/MOM_wave_interface.F90
```diff
@@ -353,40 +360,40 @@ subroutine MOM_wave_interface_init(time, G, GV, US, param_file, CS, diag )
...
   ! Langmuir number Options
-  call get_param(param_file, mdl, "LA_DEPTH_RATIO", LA_FracHBL,              &
-         "The depth (normalized by BLD) to average Stokes drift over in "//&
-         "Langmuir number calculation, where La = sqrt(ust/Stokes).",       &
-         units="nondim",default=0.04)
-  call get_param(param_file, mdl, "LA_MISALIGNMENT", LA_Misalignment,    &
+!  call get_param(param_file, mdl, "LA_DEPTH_RATIO", CS%LA_FracHBL,              &
+!         "The depth (normalized by BLD) to average Stokes drift over in "//&
+!         "Langmuir number calculation, where La = sqrt(ust/Stokes).",       &
+!         units="nondim",default=0.04)
```
breichl, line 389: So far as I can tell you moved this call up so the parameter gets set even in statistical mode.  In that case do we need to keep this around here as commented code?

src/user/MOM_wave_interface.F90
```diff
@@ -423,37 +430,8 @@ subroutine MOM_wave_interface_init(time, G, GV, US, param_file, CS, diag )
...
-  if (StatisticalWaves) then
-    WaveMethod = LF17
-    PI=4.0*atan(1.0)
-  else
-    WaveMethod = NULL_WaveMethod
-  end if
-
-  return
-end subroutine MOM_wave_interface_init_lite
-
```
breichl, line 456: Thank you for fixing/removing this!

breichl (approved):
Thanks @Hallberg-NOAA, these look like good changes and improvements to the code.  As you point out, the testing is not presently sufficient and this will be addressed soon, but I've checked some cases offline and it looks to not change any functionality.  I approve.  I made just a couple of minor comments on comments, but nothing that actually changes the code.

config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90
breichl: To clarify, "L311 of the code" refers the mom_wave_interface code, not this particular file.

JessicaMeixner-NOAA: I don't know of a current way to be able to pass this information from the coupler with NUOPC like it's done withe FMS.  So the way I was getting these variables to the ocean model was to just set them and read them in from the input file.  It's not ideal in the sense that it does allow for human error for the wave and ocean model to have different values here.

src/user/MOM_wave_interface.F90
Hallberg-NOAA: You are welcome!

src/user/MOM_wave_interface.F90
Hallberg-NOAA: Thanks for pointing this out.  This duplicated code has now been removed.

config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90
Hallberg-NOAA: This is a good explanation for why we need to leave this call here, at least for now. @JessicaMeixner-NOAA, do you think that it would it be OK if the comment is left as a reminder to revisit this some day, or should my commenting questioning why SURFBAND_WAVENUMBERS is set here be removed?

config_src/drivers/FMS_cap/ocean_model_MOM.F90
```diff
@@ -382,11 +382,9 @@ subroutine ocean_model_init(Ocean_sfc, OS, Time_init, Time_in, wind_stagger, gas
 
   call get_param(param_file, mdl, "USE_WAVES", OS%Use_Waves, &
        "If true, enables surface wave modules.", default=.false.)
-  if (OS%use_waves) then
-    call MOM_wave_interface_init(OS%Time, OS%grid, OS%GV, OS%US, param_file, OS%Waves, OS%diag)
-  else
-    call MOM_wave_interface_init_lite(param_file)
-  endif
+  ! MOM_wave_interface_init is called regardless of the value of USE_WAVES because
+  ! it also initCSializes statistical waves.
```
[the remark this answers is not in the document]
Hallberg-NOAA: Thank you for noticing this typo.  It has been corrected in an update to the PR.

config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90
Hallberg-NOAA: For now, we are going to leave this as-is with the comment noting the unusual position for the call to set  SURFBAND_WAVENUMBERS.  @JessicaMeixner-NOAA, if you want to follow this up with a PR replacing this comment with an explanation, or if you want to move this call from the NUOPC coupler to mimic the other couplers, that would be welcomed.

Hallberg-NOAA:
Brandon Reichl has approved these changes.  He will be following this up with further changes to eliminate some of the redundant options.

config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90
JessicaMeixner-NOAA: @Hallberg-NOAA sounds good

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12752

marshallward (approved):
Approving on behalf of @breichl

PR #1405, opened 2021-05-19
Call variable_att_exists before get_variable_att

Hallberg-NOAA:
Added variable_att_exists calls before two calls to get_variable_attribute in
the FMS2 version of get_file_fields so that runs do not crash if there are
variables being read that do not have a long_name or units attribute.  All
answers are bitwise identical in cases that did not crash.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12725

marshallward (approved):

PR #1407, opened 2021-05-21
corrections to beta-plane

raphaeldussin:
Setting up a channel with beta-plane configuration (center lat = 40N) with the following parameters resulted in abnormally large values of `Coriolis` leading to numerical instability. 

```
GRID_CONFIG = "spherical"       
SOUTHLAT = 10.0
LENLAT = 75.0
WESTLON = -200.0
LENLON = 100.0
ROTATION = "betaplane"
F_0 = 9.35E-05
BETA = 1.75E-11
```

Detail of the computation of parameters as follows:

```python
def betaplane(lat0, Rearth=6.378e6):
    """ write f0 and beta for a beta plane centered on lat0 (deg)"""
    import numpy as np
    omega = (2*np.pi/86400.) # earth rotation rate 2 pi / lenght of day
    deg2rad = 2*np.pi/360
    lat0r = deg2rad * lat0
    f0 = 2 * omega * np.sin(lat0r)
    beta = 2 * omega * np.cos(lat0r) / Rearth
    return f0, beta

f0_wide, beta_wide = betaplane(40)
print(f'wide channel: f0 = {f0_wide}, beta = {beta_wide}')
```

```
wide channel: f0 = 9.348966816711922e-05, beta = 1.7468900581468308e-11
```

I believe the y-scaling needs to be corrected according to this PR and an origin for the beta-plane is added for applications other
than the equatorial beta-plane. Experiment with updated code behaves as expected.

### DIFF
--- a/src/initialization/MOM_shared_initialization.F90
+++ b/src/initialization/MOM_shared_initialization.F90
@@ -511,9 +511,6 @@ subroutine set_rotation_beta_plane(f, G, param_file, US)
     case ("m")
       beta_lat_ref_units = "meters"
       y_scl = 1.
-    case ("c")
-      beta_lat_ref_units = "centimeters"
-      y_scl = 1.E-2
     case default ; call MOM_error(FATAL, &
       " set_rotation_beta_plane: unknown AXIS_UNITS = "//trim(axis_units))
   end select
--- a/src/initialization/MOM_shared_initialization.F90
+++ b/src/initialization/MOM_shared_initialization.F90
@@ -478,12 +478,13 @@ subroutine set_rotation_beta_plane(f, G, param_file, US)
   integer :: I, J
   real    :: f_0    ! The reference value of the Coriolis parameter [T-1 ~> s-1]
   real    :: beta   ! The meridional gradient of the Coriolis parameter [T-1 m-1 ~> s-1 m-1]
-  real    :: lat_0  ! The reference latitude for the beta plane [degrees]
+  real    :: beta_lat_ref  ! The reference latitude for the beta plane [degrees/km/m/cm]
   real    :: y_scl, Rad_Earth
   real    :: T_to_s ! A time unit conversion factor
   real    :: PI
   character(len=40)  :: mdl = "set_rotation_beta_plane" ! This subroutine's name.
   character(len=200) :: axis_units
+  character(len=40) :: beta_lat_ref_units
 
   call callTree_enter(trim(mdl)//"(), MOM_shared_initialization.F90")
 
@@ -495,9 +496,6 @@ subroutine set_rotation_beta_plane(f, G, param_file, US)
   call get_param(param_file, mdl, "BETA", beta, &
                  "The northward gradient of the Coriolis parameter with "//&
                  "the betaplane option.", units="m-1 s-1", default=0.0, scale=T_to_s)
-  call get_param(param_file, mdl, "LAT_0", lat_0, &
-                 "The reference latitude (origin) of the beta-plane", &
-                 units="degrees", default=0.0)
   call get_param(param_file, mdl, "AXIS_UNITS", axis_units, default="degrees")
 
   PI = 4.0*atan(1.0)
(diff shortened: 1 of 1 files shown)


marshallward:
Two minor comments/suggestions:
* Could the `LAT_0` parameter be moved inside the `degree` conditional, say after `RAD_EARTH`?
* Do you think renaming it to something like `BETA_LAT_REF` would be better?  We still use the flat parameter list so `LAT_0` may not be descriptive enough.

raphaeldussin:
`LAT_0` (or `BETA_LAT_REF`) could be useful for the other conditions as well, not only the `degree` hence I've put it outside the conditional. The issue is that the units are logged in degrees when they could be any of the ones in the conditional. I can add a `get_param` after the conditional after I define a `beta_lat_ref_units` inside if that's ok with you.

I'll rename to something more self-describing.

marshallward:
Good idea, but I think there could still be a potential mixup in the units of `geoLatBu` and `beta_lat_ref`.

You might want to use `set_grid_metrics_cartesian` as a precedent here.  It uses `AXIS_UNITS` to set the units.

raphaeldussin:
in the `cartesian` case, geoLatBu is already in `m` or `km` since the call the `MOM_initialize_rotation` is after the `set_grid_metrics` (of which `set_grid_metrics_carthesian` is a special case) in `MOM_fixed_initialization`. However, the case for `cm` is not supported in `set_grid_metrics_carthesian` so I think we should take it off the `set_betaplane` routine.

marshallward:
Sorry I was accidentally looking at the diff, not the file.  I missed that this function was already using `AXIS_UNITS`, so the units are probably OK.

For `cm`, I would think it's OK to just rescale `beta_lat_ref` cm to either m or km?  I agree it's a bit redundant, but also generally better to keep a feature.

marshallward:
There was a recommendation from @Hallberg-NOAA that we retain the original answers in case they are needed for legacy experiments.

src/initialization/MOM_shared_initialization.F90
```diff
@@ -501,16 +503,24 @@ subroutine set_rotation_beta_plane(f, G, param_file, US)
...
       " set_rotation_beta_plane: unknown AXIS_UNITS = "//trim(axis_units))
   end select
 
+  call get_param(param_file, mdl, "BETA_LAT_REF", beta_lat_ref, &
+                 "The reference latitude (origin) of the beta-plane", &
+                 units=trim(beta_lat_ref_units), default=0.0)
+
   do I=G%IsdB,G%IedB ; do J=G%JsdB,G%JedB
-    f(I,J) = f_0 + beta * ( G%geoLatBu(I,J) * y_scl )
+    f(I,J) = f_0 + beta * ( (G%geoLatBu(I,J) - beta_lat_ref) * y_scl )
```
adcroft, line 523: The offset seems wrong to me: f = fo + beta * y has two degrees of freedom for a straight line but adding `beta_lat_ref` here adds up to three parameters.

adcroft:
Which experiments are affected by the `yscl` correction?

src/initialization/MOM_shared_initialization.F90
marshallward: Is it not just a unit conversion?  I don't know if I'd consider that a degree of freedom.

marshallward: Do you mean it would be more correct to redefine `f_0` rather than introduce `beta_lat_ref`?

adcroft: Indeed, `f_0` can be adjusted to get the values you want. I see why `beta_lat_ref` was introduced (namely for convenience) but now there are two ways to get the same result. I guess it's just a formula with parameters, so it's not wrong. I'm not sure if too many parameters is a problem or not.

raphaeldussin: Imho most users expect to input parameters as in: 

https://ocw.mit.edu/courses/earth-atmospheric-and-planetary-sciences/12-808-introduction-to-observational-physical-oceanography-fall-2004/lecture-notes/course_notes_3a.pdf

With f0 being the Coriolis parameter at lat = lat0 (beta_lat_ref).

adcroft: I agree the generality provides the convenience to do it either way and that's probably a good thing.

marshallward:
After some discussion, we feel reasonably confident that there is no prior work relying on this error and we can safely replace it with the correct expression.

marshallward (approved):

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12800

PR #1408, opened 2021-05-24
Remove duplicated solvers between regrid modules

angus-g:
In the removal of `regrid_edge_slopes` (dc1bed6bd9bc29fc8cf223d49fad6f2fe46b8b73), a couple of the solvers became duplicated between `regrid_edge_values` and `regrid_solvers`. Here, they're simply made consistent by removing the solvers from `regrid_edge_values`, and bringing over the zero-pivot check for the linear solver that was in one variation but not the other.

### DIFF
--- a/src/ALE/regrid_edge_values.F90
+++ b/src/ALE/regrid_edge_values.F90
@@ -4,7 +4,8 @@ module regrid_edge_values
 ! This file is part of MOM6. See LICENSE.md for the license.
 
 use MOM_error_handler, only : MOM_error, FATAL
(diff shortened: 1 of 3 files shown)


Hallberg-NOAA (approved):
I have examined these changes, and I have carefully compared the routines that are being removed with the alternate versions that are being called.  I agree that this PR is both correct and a useful simplification.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12764.

PR #1409, opened 2021-05-24
Ice shelf restart fix

MJHarrison-GFDL:
This reorders some lines related to setting up the ice shelf restart information, allowing experiments with ice shelves enabled to restart properly.


Hallberg-NOAA (approved):
I agree that this change is correct and necessary.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12765

PR #1412, opened 2021-05-27
Fix to masking of u,v sponge code.

kshedstrom:
This also has Bob's updates to the user code for my OBC tests.

### DIFF
--- a/src/parameterizations/vertical/MOM_ALE_sponge.F90
+++ b/src/parameterizations/vertical/MOM_ALE_sponge.F90
@@ -872,6 +872,8 @@ subroutine apply_ALE_sponge(h, dt, G, GV, US, CS, Time)
   real, allocatable, dimension(:,:,:) :: sp_val_u ! A temporary array for fields
   real, allocatable, dimension(:,:,:) :: sp_val_v ! A temporary array for fields
   real, allocatable, dimension(:,:,:) :: mask_z ! A temporary array for field mask at h pts
+  real, allocatable, dimension(:,:,:) :: mask_u ! A temporary array for field mask at u pts
+  real, allocatable, dimension(:,:,:) :: mask_v ! A temporary array for field mask at v pts
   real, allocatable, dimension(:,:,:) :: tmp    !< A temporary array for thermodynamic sponge tendency diagnostics,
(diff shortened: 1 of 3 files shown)


Hallberg-NOAA (approved):
After careful visual inspection, I agree with these changes.  The one point that I had to mull over was whether the masking of the input data at velocity points should be masked if either tracer column is masked, or whether it should be unmasked if either is unmasked (as in this code).  Upon reflection, I agree with the code that if there is valid data, it can be used.  The model will take care of any land masking of the velocity points based on the coastlines, which is a separate question than the depths over which the input data exists.

  Assuming that this passes the pipeline testing, I am happy to approve this PR.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12811 :heavy_check_mark:

src/parameterizations/vertical/MOM_ALE_sponge.F90
```diff
@@ -1006,7 +1010,8 @@ subroutine apply_ALE_sponge(h, dt, G, GV, US, CS, Time)
 
       call pass_var(sp_val,G%Domain)
       do j=CS%jsc,CS%jec; do I=CS%iscB,CS%iecB
-       sp_val_u(I,j,1:nz_data) = 0.5*(sp_val(i,j,1:nz_data)+sp_val(i+1,j,1:nz_data))
+        sp_val_u(I,j,1:nz_data) = 0.5*(sp_val(i,j,1:nz_data)+sp_val(i+1,j,1:nz_data))
+        mask_u(I,j,1:nz_data) = max(mask_z(i,j,1:nz_data),mask_z(i+1,j,1:nz_data))
```
abozec, line 1014: Taking the average of the sp_val will potentially include a masked valued (huge value if set that way in the input file) and it will not be masked by taking the max of the mask_z. So I would suggest, if we want to keep the max of mask idea to weigh the sp_val by mask_z:
sp_val_u=(sp_val(i)*mask_z(i) + sp_val(i+1)*mask_z(i+1)) / max(mask_z(i)+mask_z(I+1),1e-10)

abozec: My bad, that comment was for the main-GFDL-candidate ... on the same code. I added it there.

PR #1421, opened 2021-06-10
Fluxconst temp salt

wfcooke:
Allows SALT and TEMP constants to restore surface fluxes to the relative surface anomalies.


config_src/drivers/FMS_cap/MOM_surface_forcing_gfdl.F90
```diff
@@ -1376,8 +1372,13 @@ subroutine surface_forcing_init(Time, G, US, param_file, diag, CS, wind_stagger)
                  "The constant that relates the restoring surface fluxes to the relative "//&
                  "surface anomalies (akin to a piston velocity).  Note the non-MKS units.", &
                  default=0.0, units="m day-1", scale=US%m_to_Z*US%T_to_s)
+    call get_param(param_file, mdl, "FLUXCONST_SALT", CS%Flux_const_salt, &
```
Hallberg-NOAA, line 1375: These expressions will have problems with the dimensional rescaling. CS%Flux_const has already been scaled to units of [Z sec T-1 day-1], so to use it as a default here, you would have to write it as `..., default=CS%Flux_const*US%Z_to_m*US%s_to_T, ...` . A similar issue pertains to FLUXCONST_TEMP.  Alternately (and I think preferably), you could use an `unscaled` argument in the call to get_param for FLUXCONST to get that parameter in the units from the file [m day-1] that can then be used as the default for FLUXCONST_SALT and FLUXCONST_TEMP.

  Regardless of which way you go, please verify that setting T_RESCALE_POWER = 37 and Z_RESCALE_POWER = -5 (or some other values of your choice) gives identical answers in whatever test case you are using to exercise this code as if you set T_RESCALE_POWER = 0 and Z_RESCALE_POWER = 0.

PR #1422, opened 2021-06-10
Ice freezing sss

wfcooke:


wfcooke:
@Hallberg-NOAA : I'm going to close this for now while @MJHarrison-GFDL  and I iterate over an update for this.
FYI, the parameters for this equation appear to be the same as that for masking salinity restoring under ice  at
https://github.com/NOAA-GFDL/MOM6/blob/da287e1a799e5d9250245ec3b8e9af00116896d3/config_src/drivers/FMS_cap/MOM_surface_forcing_gfdl.F90#L356

PR #1423, opened 2021-06-11
Additional memory cleanups

marshallward:
This PR contains three patches, mainly related to ice-ocean simulations.

- Three more deallocation cleanups.  Most notable is use of `horiz_interp_del` after `horiz_interp_and_extrap_tracer_record` calls.  The others are more generic final-type calls.
- A few short-circuit if-block fixes
- A linked list was added to the diag mediator to handle non-standard grid axes.  This was introduced to track and deallocate the internal copies of these axis groups.

Details are in the individual commit logs.

### DIFF
--- a/config_src/infra/FMS1/MOM_interp_infra.F90
+++ b/config_src/infra/FMS1/MOM_interp_infra.F90
@@ -5,7 +5,8 @@ module MOM_interp_infra
 
 use MOM_domain_infra,    only : MOM_domain_type, domain2d
 use MOM_time_manager,    only : time_type
-use horiz_interp_mod,    only : horiz_interp_new, horiz_interp, horiz_interp_init, horiz_interp_type
+use horiz_interp_mod,    only : horiz_interp_new, horiz_interp, horiz_interp_type
+use horiz_interp_mod,    only : horiz_interp_init, horiz_interp_del
 use mpp_io_mod,          only : axistype, mpp_get_axis_data
 use time_interp_external_mod, only : time_interp_external
 use time_interp_external_mod, only : init_external_field, time_interp_external_init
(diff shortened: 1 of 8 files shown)


marshallward:
Sorry I just noticed this PR still has a reference to FMS (`horiz_interp_end`).  I will fix that up.

config_src/infra/FMS1/MOM_interp_infra.F90
```diff
@@ -14,7 +15,7 @@ module MOM_interp_infra
 
 implicit none ; private
 
-public :: horiz_interp_type, horiz_interp_init
+public :: horiz_interp_type, horiz_interp_init, horiz_interp_del
```
Hallberg-NOAA, line 18: I think that there is a consensus that we should not have simple pass-throughs of FMS routines, but rather we should have explicit interfaces for all of the routines (like horiz_interp_del) that are being made publicly visible, so that there is at least one place where all subroutine interfaces used in MOM6 are properly documented.  So the FMS routine `horiz_interp_del()` should be wrapped in another routine, perhaps `horiz_interp_delete()` that has comments describing the routine and documenting all of the arguments that we use.  I also note that some of the other interfaces in this file (like `horiz_interp_init()`) also are simple pass-throughs without adequate documentation.

marshallward: Fair point, but these are not isolated examples and there are other examples.

I also suspect that the whole operation could be reconsidered so that the MOM user does not need to explicit call any `_del()` operations.

I would suggest we accept these as following the established pattern, and address it in a future PR.

Hallberg-NOAA: Yes, upon further inspection of the code, I do see that there are a number of instances of undocumented pass-through interfaces from FMS. Especially given that these new interfaces are only used in modules in the src/framework code, within which there are well-documented interfaces, I think that upon further reflection I agree that these are OK.

marshallward:
Something in this PR has broken FMS2 runs, and possibly FMS1 runs as well.

It has something to do with the diag mediator commit, and specifically with the deallocation of some key components which may be needed by the diag manager, which may itself be called after the `*_end` call of the submodels.

If there's no easy fix here then I will withdraw it, and possibly just send the two smaller fixes.

marshallward:
I don't think this PR was the issue, but I will close for now while some more volatile issues are being sorted out.

PR #1424, opened 2021-06-11
fix reflect for internal tides

raphaeldussin:
* En array defined with wrong size leading to energy leakage
* clean initialization for local arrays


Hallberg-NOAA (changes requested):
I agree that the change in the array size declaration for En makes it consistent with the variables being passed in. However, this array size does not make any sense for where these variables are staggered. Looking at the routines that call reflect(), it seems to me that the right answer is to correct the original array size declarations for Fdt_m and Fdt_p at lines 1392 and 1467, and not make any changes to the array size declaration for the En argument at line 1613.

raphaeldussin:
@Hallberg-NOAA I don't understand why these other branch and PR have been merged onto mine

raphaeldussin:
that PR and branch have been compromised and code changes are not giving the expected answers, closing PR, deleting branch and re-opening clean one...

PR #1425, opened 2021-06-14
Change the declaration of stack variable in apply_flux_adjustments

MJHarrison-GFDL:
This fixes an error encountered in SPEAR configurations.


Hallberg-NOAA (approved):
The data override capability for the surface forcing needs to work with an array that is allocated only on the computational grid.  This PR does just that.  This PR has passed TC testing and passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12882.

PR #1426, opened 2021-06-15
(*)Correct multi-name handling by save_restart

Hallberg-NOAA:
Corrected the code that adds a suffix to the restart file names created by
save_restart when there are multiple restart files to satisfy the NetCDF
requirement that all variables in a time record start within the first 4 Gb of
the record in the case where the restart file name was provided with the ".nc"
extension already included.  The code that reads the restarts already handled
this case properly, but save_restart had an extra ".nc" in the names.  MOM6
usually provided the root for the restart file without the .nc extension, but
SIS2 does not and was having trouble reading its own restart files in some very
large (e.g., 1/12-degree global) cases; identical corrections were made
separately to MOM_restart.F90 and SIS_restart.F90.  All answers and filenames
are bitwise identical in any cases that were correctly restarting before.

### DIFF
--- a/src/framework/MOM_restart.F90
+++ b/src/framework/MOM_restart.F90
@@ -847,13 +847,14 @@ subroutine save_restart(directory, time, G, CS, time_stamped, filename, GV, num_
   integer :: start_var, next_var        ! The starting variables of the
                                         ! current and next files.
   type(file_type) :: IO_handle          ! The I/O handle of the open fileset
-  integer :: m, nz, num_files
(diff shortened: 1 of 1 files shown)


Hae-CheolKim-NOAA:
It works now. Thanks!

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12907

marshallward (approved):

PR #1427, opened 2021-06-15
Update flux const

wfcooke:
Updating for addition of FLUX_CONST_SALT and FLUX_CONST_TEMP with clean history.
Unit_scaling has been added, but answers change for different values of T_RESCALE_POWER.

### DIFF
--- a/config_src/drivers/FMS_cap/MOM_surface_forcing_gfdl.F90
+++ b/config_src/drivers/FMS_cap/MOM_surface_forcing_gfdl.F90
@@ -111,6 +111,8 @@ type, public :: surface_forcing_CS ; private
   logical :: restore_temp       !< If true, the coupled MOM driver adds a term to restore sea
                                 !! surface temperature to a specified value.
   real    :: Flux_const                     !< Piston velocity for surface restoring [Z T-1 ~> m s-1]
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA (changes requested):
This PR is failling with "invalid memory references" in the OM4_025 test case with the gnu compiler in the pipeline tests.  Unfortunately the trace-back for these errors does not provide concrete information about where in the code the error is occurring (see
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/jobs/65471 ), but if I had to guess, the conditional allocation of fluxes%salt_flux_added and fluxes%heat_flux_added would seem like the most likely suspect.  Running with the intel compiler or with with an executable compiled for debugging might give more useful information.

  Please correct these errors and verify that the OM4_025 test case runs properly with the updated code.

MJHarrison-GFDL:
Thanks @Hallberg-NOAA . @wfcooke , can you please remore my changes for the conditional allocation of these arrays, Sorry about the mixup.

Hallberg-NOAA:
After the update, this PR is now passing the pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/12946, but because it changes the contents of the MOM_parameter_doc files, it has to be merged manually.

Hallberg-NOAA (approved):

PR #1428, opened 2021-06-15
Bugfix: Minor changes on clipping topography from file

herrwang0:
This PR fixes a bug on clipping topography. 

Topography read from file are clipped and the actual values used by the model are between `MINIMUM_DEPTH` and `MAXIMUM_DEPTH`. If `MASKING_DEPTH` is specified, D<=`MASKING_DEPTH` is masked as land; otherwise D<=`MINIMUM_DEPTH` is masked. 

The original code is not working with topography containing valid negative values, i.e. wetting and drying. With the fix, both `MINIMUM_DEPTH` and `MASKING_DEPTH` can now be set as negative numbers, so that wetting/drying topography can be corrected clipped as desired.

The illustration below shows how `D` is changed after calling the two subroutines involved.
```
If MASKING_DEPTH is specified:

 ---MASKING_DEPTH----MINIMUM_DEPTH----MAXIMUM_DEPTH-----
 ---MASKING_DEPTH[M I N IMUM_DEPTH----MAXIMUM_D E P T H]   # call limit_topography()
 ////////////////[M I N IMUM_DEPTH----MAXIMUM_D E P T H]   # call initialize_masks()

If MASKING_DEPTH is unspecified:

 --------------------MINIMUM_DEPTH----MAXIMUM_DEPTH-----
 M  I  N  I  M  U  M  _  D E P T H----MAXIMUM_D E P T H]   # call limit_topography()
 /////////////////////////////////----MAXIMUM_D E P T H]   # call initialize_masks()
```


herrwang0:
Comments on the regression tests:
The solution and diag changes are related to the change of this part of code:
https://github.com/NOAA-GFDL/MOM6/blob/195575a5d39ee2c5f02a9736c39ff9c554ade84c/src/initialization/MOM_shared_initialization.F90#L417-L421 

This particular line is used when `MASKING_DEPTH` is **not** specified (technically, when it is smaller than -9990 meter  with the default being -9999). The original code intends to reset all points <0.5\*`min_depth`, and then in the subsequent subroutine `initialize_masks` everything <=`min_dpeth` is masked as land point. This is problematic for two reasons:

* The underlying assumption here is 0.5\*`min_depth` < `min_depth` and `MINIMUM_DEPTH` can only be positive. This won't work with wetting/drying cases, in which negative topography points exist and `MINIMUM_DEPTH` must be negative. 

* In addition, the exact value to clip the topography [in case of original code, 0.5\*`min_depth`] should not matter as long as it is <=`min_depth` - it is only resetting would-be masked-out land points. 

Therefore, in the proposed change, `0.5` is dropped. The change does introduce apparent solution and diagnostic changes:

1. `tc1` series have solutions changes: 
* `benchmark` topography has land points whose values are set differently before and after the proposed change. 
* Only `Mean SL` field in `ocean.stats` is changed. This seems to be related to different values in the masked-out regions. It looks like the calculation in subroutine`write_energy` do not exclude land points. Therefore if there is a difference over the land topography, even if it is supposed to be masked out, the "solution" will change. 

2. `tc1` and `tc2` series have diag changes:
* Similarly, halo points have different values, so chksum bitcount with a haloshift will be different.

awallcraft:
I suggest preserving answers by using 0.5*min_depth when min_depth is non-negative and whatever makes sense when min_depth is negative.

Hallberg-NOAA:
I agree with @awallcraft about how the code in this PR could be modified to avoid changing any answers in the cases that are already being tested.

herrwang0:
@awallcraft @Hallberg-NOAA Thanks for the comments. This makes sense. Will update to reflect this.

src/initialization/MOM_grid_initialize.F90
```diff
@@ -1217,11 +1217,17 @@ subroutine initialize_masks(G, PF, US)
...
 
+  if (mask_depth > min_depth) then
+    mask_depth = -9999.0*m_to_Z_scale
+    call MOM_error(WARNING, "MOM_grid_init: initialize_masks "//&
+                  'MASKING_DEPTH is larger than MINIMUM_DEPTH and therefore ignored.')
+  endif
+
   Dmin = min_depth
-  if (mask_depth>=0.) Dmin = mask_depth
+  if (abs(mask_depth - (-9999.*m_to_Z_scale)) > 1.0e-10) Dmin = mask_depth
```
Hallberg-NOAA, line 1230: This expression is not dimensionally consistent.  The right hand side of the inequality should be `... > 1.0e-10*m_to_Z_scale`, with a similar issue on line 423 of MOM_shared_initialization.F90.

src/initialization/MOM_shared_initialization.F90
```diff
@@ -413,17 +413,27 @@ subroutine limit_topography(D, G, param_file, max_depth, US)
...
-      D(i,j) = min( max( D(i,j), 0.5*min_depth ), max_depth )
-    enddo ; enddo
+  if (mask_depth > min_depth) then
+    mask_depth = -9999.0*m_to_Z
+    call MOM_error(WARNING, "MOM_shared_initialization: limit_topography "//&
+                  'MASKING_DEPTH is larger than MINIMUM_DEPTH and therefore ignored.')
+  endif
+
+  ! Make sure that min_depth < D(x,y) < max_depth for ocean points
+  if (abs(mask_depth - (-9999.*m_to_Z)) < 1.0e-10) then
```
Hallberg-NOAA, line 423: The expression on line 423 has problems with the dimensional rescaling.  It should be `if (abs(mask_depth - (-9999.*m_to_Z)) < 1.0e-10*m_to_Z) then`.

src/initialization/MOM_grid_initialize.F90
herrwang0: Thanks for pointing this out!
A related but probably too subtle question: What is the best practice here for comparing two float point numbers here? The purpose of this expression really should be checking if `mask_depth` is taking the default or not. I ended up using a minimum thickness (an Angstrom, which should be scaled here), but not sure if the logic actually checks out: we want to check  if the depth numbers are equal, rather than two thicknesses are considered Indistinguishable

Hallberg-NOAA: In this case, because we are setting a real number explicitly to a hard-coded default value, we are probably OK doing a simple comparison, although in other cases your instincts of doing a comparison between real values to within a small tolerance (but larger than roundoff) would be prudent.  The dimensional scaling would have tripped us up here because 1e-10 is not scaling with the same factor as the other values here. If we had effectively rescaled our vertical units by 2^50 (which does work), we would be using something of order a light-year as our unit for vertical heights, and 1e-10 light years is not a small number compared with the depth of the ocean!

Hallberg-NOAA (approved):
I agree with these changes, and they are now passing all of the TC testing and the pipeline testing (at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13021), although the MOM_parameter_doc files will need to be updated due to a change in the description of MASKING_DEPTH.

favorliao:
I am running the latest MOM6-examples in the Princeton cluster (version on Jul. 28 2021, OM4_05, everything is the default) and have this error message "NaN in input field of reproducing_EFP_sum(_2d)". 
According to my test, the error is related to the fixed code related to mask_depth. If I add the following code in Line 437 in src/initialization/MOM_shared_initialization.F90, it could be resolved. I am not sure if this is the correct way to resolve this issue. Could you help to test if this happens to you?

"      if (D(i,j) > mask_depth) then
        D(i,j) = min( max( D(i,j), min_depth ), max_depth )
      else !add this in the new code
        D(i,j) = 0. !add this in the new code
"

herrwang0:
Hi @favorliao 
I just did a run of OM4_05 on Gaea but wasn't able to reproduce the issue. 
It seems that in your case the model thought there was NaN over the land. Unless I missed something, the topography file used in this example doesn't contain any peculiar number. Could you double check your topography?
 
Otherwise, we can add the else statement as you suggested (but we want to set it to `mask_depth` instead of zero). 

```Fortran
if (D(i,j) > mask_depth) then
  D(i,j) = min( max( D(i,j), min_depth ), max_depth )
else
  D(i,j) = mask_depth
endif
```

My initial reason for not having the else statement intent was to only modify the read-in topography when needed and let land-sea mask take care of the rest. But there is no reason to follow that philosophy if it breaks runs or changes answers.

favorliao:
Hi @herrwang0 
My topography is from ftp://ftp.gfdl.noaa.gov/perm/Alistair.Adcroft/MOM6-testing/
OM4_05/mosaic_ocean.v20180227.unpacked/ocean_topog.nc (I downloaded two weeks ago. Is there any update on the topography file?)
In the default setup I ran, the mask_depth=0; the land point depth in the ocean_topog.nc is also 0. Probably this excludes the condition of d(i,j)=0. Should I update the topography file or use the else statement (D(i,j) = mask_depth)?

herrwang0:
> Hi @herrwang0
> My topography is from ftp://ftp.gfdl.noaa.gov/perm/Alistair.Adcroft/MOM6-testing/
> OM4_05/mosaic_ocean.v20180227.unpacked/ocean_topog.nc (I downloaded two weeks ago. Is there any update on the topography file?)
> In the default setup I ran, the mask_depth=0; the land point depth in the ocean_topog.nc is also 0. Probably this excludes the condition of d(i,j)=0. Should I update the topography file or use the else statement (D(i,j) = mask_depth)?

I checked the topography file you provided and it seemed fine, so I doubt it is anything wrong with the topography file. `D` reads  from the topography file, where the land points are zero, so you shouldn't have to set it to zero again here. 

I am happy to help but it seems that I couldn't reproduce the bug. Apparently you can easily fix this by adding an else-statement, but I guess I'm a bit puzzled (and curious about) what was really wrong here.

favorliao:
It is really strange that only happens to me. I will continue digging into it and would let you know once I have some answers.

adcroft:
I spoke too soon. It appears later changes (maybe #1457 ?) further changed the meaning of MASKING_DEPTH. The patch I described above isn't enough. I'll have to `git bisect` again...

Issue #1430, opened 2021-06-20
Imperfect restart with uv_sponges


kshedstrom:
On restart, the h field on the outside of the domain comes in as 1.e-10. It is used as such to find the h_col field for the uv_sponges right at the edge, differing from the h_col value found later on in the run, in which h has been set to be an image of the interior value. It probably wouldn't matter if there were no open boundaries.

Options:
1. Set h outside before the call to the sponges.
2. Add OBC code to the uv_sponges.

As I recall, h outside had some nonsense value on startup as well, different from the restart nonsense value.

By the way, my Bering domain with sponges gives slightly different answers optimized and not. We don't want this behavior, do we?

MJHarrison-GFDL:
@kshedstrom . Shouldn't we be disabling the uv_sponges at the boundary ?

kshedstrom:
OK, that's a third option, but we should enforce it via MOM6 code, not count on people to build the sponge weights with zeros at the edges. @MJHarrison-GFDL @adcroft @Hallberg-NOAA

MJHarrison-GFDL:
@kshedstrom  After the call to horiz_interp, mask_z is not filled in the halos. Adding a pass_var and replacing the definition of mask_uv with min(ij, ij+1) instead of max(ij,ij+1) would eliminate the sponging at the OBC points.

abozec:
@kshedstrom @MJHarrison-GFDL, just an FYI,  I've just sent a pull request that get reproducibility with sponge_uv (SpongeNotOnGrid) and OBC, but that required a mask at 0. for the u and v sponge variable.

kshedstrom:
In thinking about it, in other places the OBC code makes sure that outside values are never, ever used. I'm not sure why h outside ever gets set because we don't want to be using it for anything. The OBCs everywhere demands that you don't depend on outside values since one point can be the image point for more than one interior point.

kshedstrom:
Closed by #1434

PR #1434, opened 2021-06-24
Solution to UV sponge restarts

kshedstrom:
Solution to #1430, sitting on other PRs, hopefully not a problem.

@MJHarrison-GFDL's tips were helpful, not quite all that was required.

I cleaned up the unused tmpT1d stuff from the uv sponges.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,49 +1,34 @@
 stages:
-  - merge+setup
   - builds
   - run
   - tests
   - cleanup
 
+variables:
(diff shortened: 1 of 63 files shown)


MJHarrison-GFDL:
This commit has #1429 merged.

Hallberg-NOAA (approved):
I agree with the logic of these changes, and they have passed both the TC testing and the pipeline tests (which I ran manually).  In particular, the halo update on the input file mask should avoid improperly using data that inadvertently differs between PEs.

 However, it should be noted that this PR could change answers in some (unidentified) cases that attempt to use the velocity sponges at open boundary points.

PR #1435, opened 2021-06-25
(+)Modified some optional arguments

Hallberg-NOAA:
Modified optional arguments in 4 modules to reflect their actual usage.

1. Eliminated the optional argument full_prec to zonal_flux_adjust and
   meridional_flux_adjust, which were always called with the hard-coded value
   "true", and made the optional arguments monotonic and simple_2nd to
   PPM_reconstruction_[xy] mandatory.

2. Eliminated the optional argument eta_bt to calculate_diagnostic_fields, which
   was never present.

3. Made the two optional arguments to unit_scaling_init mandatory.

4. Eliminated the optional do_i argument to F_to_ent, which was never present
   in calls, and made the parameter just_read_params to entrain_diffusive_init
   mandatory.

All answers are bitwise identical.


adcroft (approved):
Pending https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13047

PR #1436, opened 2021-06-30
Update gitlab pipeline to use scripts from MOM6-examples

adcroft:
This essentially moves commands from .gitlab-ci.yaml to scripts in MOM6-examples.
- This emphasizes the circular dependence we have for testing: MOM6-examples is the one place that "knows" versions and paths to src repositories and so is where we have to define the build process.
- For now, I've put in place the same Makefiles we were using anyway but we will work towards a library-style approach allowing a mode modular definition of the build process
- The .gitlab-ci.yml is much simpler (one line per test)
- I took the opportunity to fold (tidy) much of the output to avoid defaulting to massive raw logs, e.g. https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/jobs/66390
- I also sped up the end-to-end by about 5 minutes although we are subject to waiting in the queue and a slow lustre file system, and thus still ~1 hr.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,49 +1,34 @@
 stages:
-  - merge+setup
   - builds
   - run
   - tests
   - cleanup
 
+variables:
+  CACHE_DIR: "/lustre/f2/scratch/oar.gfdl.ogrp-account/runner/cache/"
+
+
 # Merges MOM6 with dev/gfdl. Changes directory to test directory, if it exists.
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA (approved):
I agree with the script simplification in these updates, and they are passing both the TC testing and the updated pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13051 .

PR #1437, opened 2021-07-02
+Add wind-stress acceleration diagnostics

Hallberg-NOAA:
Added new diagnostics of the acceleration driven by the wind stress
accelerations as redistributed by a timestep of the vertical viscosity and not
lost to bottom drag within a timestep.  This is also in the diagnostics of the
accelerations due to the vertical viscosity, but the redistribution can be found
from the difference of the two.  Also added a diagnostic of the contribution of
the wind stresses to kinetic energy, and applied an underflow limiter on both
the new acceleration diagnostic and the existing viscous acceleration
diagnostic.  All solutions are bitwise identical, but there are new diagnostics.


adcroft (approved):
I particularly appreciate the more concise OMP directives.

Testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13244

Issue #1417, opened 2021-06-04
`sponges` don't work in adiabatic configurations; enhance docstrings?


navidcy:
We've been trying to set up an idealised configuration in isopycnal coordinates with sponges. After lots of trial and error and with the detective skills of @angus-g we (re)discovered that `ADIABATIC=TRUE` won't allow sponges.

Retrospectively, it makes sense. But perhaps enhancing the docstring a bit to explicitly say that sponges only work in adiabatic configurations would be good?

PR #1438, opened 2021-07-02
Flag inconsistent parameters if ADIABATIC = True

Hallberg-NOAA:
Add checks for inconsistent parameter settings in adiabatic_driver_init() when
ADIABATIC = True, and issue instructive error messages if any are found.  This
PR addresses MOM6 issue #1417.  All answers are bitwise identical, although some
cases where the inconsistent parameter settings were previously ignored may now
issue fatal errors and will not run.


adcroft (approved):
These changes make sense. Testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13258

Issue #1214, opened 2020-09-23
Abort if registering a restart too late in init()?


mnlevy1981:
I'm working on the MARBL driver based off of the NCAR fork, and I ran into something unexpected with `register_restart_field()`. Basically, we use an iterative solver to compute PH, with an initial guess of PH from the previous time step - to ensure bit-for-bit comparisons when restarting, we need these initial guesses in the restart file. I naively called `register_restart_field()` from a subroutine called by `tracer_flow_control_init()` and the fields were dutifully being written to the restart file but not being read in upon restarting. I moved the calls to a subroutine  called from `call_tracer_register()` and everything worked as expected.

So it appears to me that the restart file is read in somewhere between the call to `call_tracer_register()` and the call to `tracer_flow_control_init()` -- if that's the case, it would be really nice to get an explicit error message when trying to register a new restart field after the restarts have already been processed. I didn't see anything in the log files to indicate I had done anything wrong, so it was dumb luck that I stumbled on the solution on my own.

marshallward:
Hi Michael,

I discussed with others here and we agree that there should be no reason to call `register_restart_field` after `restore_state` has been called.  So it may be possible to introducing a "lock" on restart registration and raising an error.

Having said that, there are two minor and one major issue with doing this right now.

The first minor issue is that we do actually call `register_restart_field` after `restore_state` as part of the indirect method for writing the initial state.  More on this later.

The second minor issue is that `restore_state` would not be called on an initialization, so we would also need to find a way to introduce this lock without actually calling `restore_state`.

I don't think either of these instances would be difficult to resolve.

But there is a larger issue, namely that we are preparing for a major overhaul of the FMS I/O, which includes a new handling of restart files (#1165).  Any work that we here would likely cause a conflict with this PR.

There is also the possibility that this ought to be handled at the FMS level, which we can also look into.

Due to the ongoing FMS update, I'm going to recommend that we hold off on this and leave this issue open until the FMS PR has been merged.  Does that work for you?

mnlevy1981:
> Due to the ongoing FMS update, I'm going to recommend that we hold off on this and leave this issue open until the FMS PR has been merged. Does that work for you?

That sounds reasonable to me -- I don't think this is an urgent issue, but did want to open a ticket in case others run into similar problems with restart files seemingly not being read.

PR #1441, opened 2021-07-08
+Add a restart registry lock & fix OBC call order

Hallberg-NOAA:
Added code to lock the restart registry once all registration should have
occurred or if the restart has been read, along with a new public interface,
restart_registry_lock, to allow this lock to be set or unset.  All calls to
register restart fields now check the state of this lock and issue a fatal error
if the registry is locked.  This PR addresses MOM6 issue #1214.

  In the process of adding this restart lock, the new error messages revealed
that some of the restart registration calls related to some types of open
boundary conditions were not happening early enough.  To avoid this, a new
interface, register_DOME_OBC, was added to the DOME_initialization module and is
being called from call_OBC_register, and a number of the OBC-related calls
during the initialization were collected in the same (appropriate) place.  Some
OBC error messages were also corrected.  All answers are bitwise identical, but
there are two new public interfaces and the order of some OBC-related entries in
the MOM_parameter_doc calls changed.

### DIFF
--- a/src/core/MOM.F90
+++ b/src/core/MOM.F90
@@ -41,7 +41,7 @@ use MOM_io,                   only : MOM_io_init, vardesc, var_desc
 use MOM_io,                   only : slasher, file_exists, MOM_read_data
 use MOM_obsolete_params,      only : find_obsolete_params
 use MOM_restart,              only : register_restart_field, register_restart_pair
-use MOM_restart,              only : query_initialized, save_restart
+use MOM_restart,              only : query_initialized, save_restart, restart_registry_lock
 use MOM_restart,              only : restart_init, is_new_run, determine_is_new_run, MOM_restart_CS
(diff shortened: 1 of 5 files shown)


adcroft (approved):
Passed at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13260

andrew-c-ross:
+1 @nikizadehgfdl  

@Hallberg-NOAA this PR also appears to have introduced a second bug when running with both OBCs and generic tracers/COBALT. Moving call_OBC_register to after tracer_registry_init results in the generic tracers being registered before temperature and salinity on the boundary, meaning T and S become tracers 34-35 and everything expecting them to be 1-2 breaks. 

I have a tentative fix and can open an issue after some more work to confirm the bug is resolved.

PR #1442, opened 2021-07-09
fix dim_names assignment in MOM_io:read_var_sizes

### LINKED ISSUE #1439: MOM_io:read_var_sizes sets all dim_names to last dimname
In the following loop
https://github.com/NOAA-GFDL/MOM6/blob/52f906cfc458ee77630f75c9d65da52a87fcfa70/src/framework/MOM_io.F90#L712-L719
the assignment on line 717 sets the full array of strings `dim_names` to `dimname`, instead of just setting `dim_names(n)`. The effect is that all `dim_names` values get set to the last value of `dimname`.

I encountered the above in an experiment where I set `AGE_IC_FILE_IS_Z=True`, and `AGE_IC_FILE` to a file containing `age` with a singleton `time` dimension and the corresponding `time` variable didn't have an `edges` attribute. This combination led to `dim_names(3)='time'` in `MOM_tracer_Z_init:read_Z_edges` and `MOM_io:read_attribute_str` aborted at line 916 in the following block
https://github.com/NOAA-GFDL/MOM6/blob/52f906cfc458ee77630f75c9d65da52a87fcfa70/src/framework/MOM_io.F90#L910-L927
So it looks like `read_attribute_str` is aborting with message `Attribute data type is not a char ...` if the `edges` attribute is not found. I find that error message confusing, as it misdirects the user in this scenario.

Also, is aborting the desired behavior if the argument `found` is passed to `read_attribute_str`, which is the case in `read_Z_edges`'s call to `read_attribute`? I would have expected `rc == NF90_ENOTATT` to be a non-fatal error if the optional argument `found` were present.

klindsay28:
fixes #1439

### DIFF
--- a/src/framework/MOM_io.F90
+++ b/src/framework/MOM_io.F90
@@ -714,7 +714,7 @@ subroutine read_var_sizes(filename, varname, ndims, sizes, match_case, caller, d
     if (status /= NF90_NOERR) call MOM_error(WARNING, trim(hdr) // trim(NF90_STRERROR(status)) //&
         " Getting dimension length for "//trim(varname)//" in "//trim(filename))
     if (present(dim_names)) then
-      if (n <= size(dim_names)) dim_names = trim(dimname)
+      if (n <= size(dim_names)) dim_names(n) = trim(dimname)
     endif
   enddo
   deallocate(dimids)


Hallberg-NOAA (approved):
Thank you for detecting and fixing this bug, @klindsay28!

marshallward:
Although our pipeline is down, I was able to replicate it locally, so I can "sign off" on this.  I will merge this now in order get it into the current dev/gfdl -> main review.

When we get the runners back online, I will go back and validate the PR through GitLab.

PR #1443, opened 2021-07-09
+Add query_wave_properties & fix NUOPC wave queries

Hallberg-NOAA:
Added a new routine, query_wave_properties, that can be call to get
information about the wave properties from the waves control structure, and
added a get_param call for SURFBAND_WAVENUMBERS to MOM_wave_interface_init when
it is using the options that are typical with NUOPC coupler.  These changes
should allow the NUOPC coupler to compile and work again, while still keeping
the same level of opacity in the wave_parameters_CS.  All answers should be
bitwise identical, although the order of some entries in the MOM_parameter_doc
files may change when coupled with waves is WAVE_METHOD=SURFACE_BANDS and
SURFBAND_SOURCE=COUPLER.

### DIFF
--- a/src/user/MOM_wave_interface.F90
+++ b/src/user/MOM_wave_interface.F90
@@ -323,10 +323,10 @@ subroutine MOM_wave_interface_init(time, G, GV, US, param_file, CS, diag )
          "STK_BAND_COUPLER is the number of Stokes drift bands in the coupler. "// &
          "This has to be consistent with the number of Stokes drift bands in WW3, "//&
          "or the model will fail.",units='', default=1)
+      allocate( CS%WaveNum_Cen(CS%NumBands) )
       call get_param(param_file, mdl, "SURFBAND_WAVENUMBERS", CS%WaveNum_Cen, &
            "Central wavenumbers for surface Stokes drift bands.", &
            units='rad/m', default=0.12566, scale=US%Z_to_m)
-      allocate( CS%WaveNum_Cen(CS%NumBands) )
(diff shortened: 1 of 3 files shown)


src/user/MOM_wave_interface.F90
```diff
@@ -322,6 +323,9 @@ subroutine MOM_wave_interface_init(time, G, GV, US, param_file, CS, diag )
          "STK_BAND_COUPLER is the number of Stokes drift bands in the coupler. "// &
          "This has to be consistent with the number of Stokes drift bands in WW3, "//&
          "or the model will fail.",units='', default=1)
+      call get_param(param_file, mdl, "SURFBAND_WAVENUMBERS", CS%WaveNum_Cen, &
+           "Central wavenumbers for surface Stokes drift bands.", &
+           units='rad/m', default=0.12566, scale=US%Z_to_m)
       allocate( CS%WaveNum_Cen(CS%NumBands) )
```
Hallberg-NOAA, line 329: The allocate call for CS%WaveNum_Cen needs to occur before it can be read into with the get_param call.  I have an update to this PR that fixes this problem, but I can not push it back to github at the moment because Gaea is reporting network connectivity issues.

Hallberg-NOAA: Gaea can not connect to the internet from interactive nodes, and I was foolishly pushing from the wrong window.  The updated version of this PR addresses this issue.

marshallward:
I have run the Gaea regression suite manually and can confirm that this has passed the regression suite.

We will repeat this test through GitLab when the runners are back online.

marshallward (approved):

PR #1446, opened 2021-07-19
initialize CS%WaveNum_Cen before read in this parameter in MOM_wave_interface.F90

jiandewang:
initialize CS%WaveNum_Cen before read in this parameter in MOM_wave_interface.F90


Hallberg-NOAA (approved):
I agree that these changes correct a problem that was recently introduced for certain code configurations that use an active waves module.

marshallward:
I've confirmed that this passes a manual regression test and will move it.

As for the last few PRs, we will validate this PR in the pipeline.

PR #1448, opened 2021-07-20
add option to prescribe internal wave generation site in grid indices, fixes #1149

### LINKED ISSUE #1149: Nitpicking on internal tides source test
in https://github.com/NOAA-GFDL/MOM6/blob/d44e79b9b16310d0539352ab48bdd6dd941d8b94/src/parameterizations/vertical/MOM_internal_tide_input.F90#L129

should this be:

```diff

     if (time_end <= CS%time_max_source) then
       do j=js,je ; do i=is,ie
         ! Input  an arbitrary energy point source.id_
-        if (((G%geoLonCu(I-1,j)-CS%int_tide_source_x) * (G%geoLonBu(I,j)-CS%int_tide_source_x) <= 0.0) .and. &
-            ((G%geoLatCv(i,J-1)-CS%int_tide_source_y) * (G%geoLatCv(i,j)-CS%int_tide_source_y) <= 0.0)) then
+        if (((G%geoLonCu(I-1,j)-CS%int_tide_source_x) * (G%geoLonCu(I,j)-CS%int_tide_source_x) <= 0.0) .and. &
+            ((G%geoLatCv(i,J-1)-CS%int_tide_source_y) * (G%geoLatCv(i,J)-CS%int_tide_source_y) <= 0.0)) then
           itide%TKE_itidal_input(i,j) = 1.0*US%kg_m3_to_R*US%m_to_Z**3*US%T_to_s**3
         endif
       enddo ; enddo

```

raphaeldussin:
comparing with longitude/latitude does not allow to specify accurately the source of internal wave generation.
the new logical flag allows to pick a point using global indices. The former option is kept for now but could be phased out
later.

### DIFF
--- a/src/parameterizations/vertical/MOM_internal_tide_input.F90
+++ b/src/parameterizations/vertical/MOM_internal_tide_input.F90
@@ -51,8 +51,12 @@ type, public :: int_tide_input_CS ; private
   type(time_type) :: time_max_source !< A time for use in testing internal tides
   real    :: int_tide_source_x       !< X Location of generation site
                                      !! for internal tide for testing (BDM)
+                                     !! for internal tide for testing (BDM)
   real    :: int_tide_source_y       !< Y Location of generation site
                                      !! for internal tide for testing (BDM)
+  integer :: int_tide_source_i       !< I Location of generation site
+  integer :: int_tide_source_j       !< J Location of generation site
+  logical :: int_tide_use_glob_ij    !< Use global indices for generation site
 
 
   !>@{ Diagnostic IDs
@@ -99,6 +103,7 @@ subroutine set_int_tide_input(u, v, h, tv, fluxes, itide, dt, G, GV, US, CS)
   type(time_type) :: time_end        !< For use in testing internal tides (BDM)
 
   integer :: i, j, k, is, ie, js, je, nz, isd, ied, jsd, jed
+  integer :: i_global, j_global
 
   is = G%isc ; ie = G%iec ; js = G%jsc ; je = G%jec ; nz = GV%ke
   isd = G%isd ; ied = G%ied ; jsd = G%jsd ; jed = G%jed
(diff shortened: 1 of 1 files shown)


src/parameterizations/vertical/MOM_internal_tide_input.F90
```diff
@@ -389,10 +404,17 @@ subroutine int_tide_input_init(Time, G, GV, US, param_file, diag, CS, itide)
                  "If true, apply an arbitrary generation site for internal tide testing", &
                  default=.false.)
   if (CS%int_tide_source_test)then
+    call get_param(param_file, mdl, "INTERNAL_TIDE_USE_GLOB_IJ", CS%int_tide_use_glob_ij, &
+                 "Use global IJ for interal tide generation source test", default=.false.)
     call get_param(param_file, mdl, "INTERNAL_TIDE_SOURCE_X", CS%int_tide_source_x, &
```
Hallberg-NOAA, line 409: Depending on the value of INTERNAL_TIDE_USE_GLOB_IJ, only one for the two sets of parameters INTERNAL_TIDE_SOURCE_[XY] or INTERNAL_TIDE_SOURCE_[IJ] will get used. I think that maybe it would make sense to (1) only log the pair that will be used (i.e., set `do_not_log=CS%int_tide_use_glob_ij` on lines 410 and 412 and `do_not_log=.not.CS%int_tide_use_glob_ij` on lines 414 and 416 and (2) add a useful error message if one of the parameters that will not get used is given a non-default value.

Hallberg-NOAA (changes requested):
I think that this is a case where slight code changes to improve the error checking on the use of the new parameters (see the specific suggestions in a separate comment) could save some user a lot of grief in the future.  @raphaeldussin, please feel free to rebut this suggestion if you think it is more effort than it is worth.  Otherwise this PR looks like it will be a useful addition.

src/parameterizations/vertical/MOM_internal_tide_input.F90
raphaeldussin: added

raphaeldussin:
@Hallberg-NOAA the last changes implement the proposed improvements. ready to merge on my end

Hallberg-NOAA:
I think that there should be one more minor addition to this PR.  There is a test to see if the (x,y) test source locations are being set and `CS%int_tide_use_glob_ij == .true.`, and that test looks perfect, but I think that there should also be a test to see if the (i,j) test source locations are being set and `CS%int_tide_use_glob_ij == .false.`.

raphaeldussin:
agreed! and added!

Hallberg-NOAA (approved):

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13332

Issue #1449, opened 2021-07-20
Geothermal fails if enabled with GEOTHERMAL_SCALE = 0


marshallward:
In some tests, turning on `DO_GEOTHERMAL` but setting `GEOTHERMAL_SCALE` to zero will cause segmentation faults:
```
At line 604 of file /lustre/f2/dev/gfdl/Marshall.Ward/CI/Gaea-stats-MOM6-examples/MOM6-examples/src/MOM6/src/parameterizations/vertical/MOM_geothermal.F90
Fortran runtime error: Attempt to DEALLOCATE unallocated 'cs'
```
I suspect this should be an easy fix.  At worst, we can just do an `if(associated(..))` check.

Thanks to @sanAkel for reporting.

PR #1450, opened 2021-07-21
+Fixed rare issue with MOM_geothermal and cleanup

Hallberg-NOAA:
Added a test to avoid attempting to deallocate the geothermal heating field if
it is not allocated, and changed the geo_heat element of geothermal_CS from a
pointer into an allocatable array.  Also clarified the comments describing
several of the elements of geothermal_CS, and added a test to avoid logging the
value of GEOTHERMAL_DRHO_DT_INPLACE when the model is not in layered-mode and
this parameter is unused.  This PR addresses MOM6 issue #1449.  All answers are
bitwise identical in all cases that worked before, but there are fewer entries
in some ALE-mode MOM_parameter_doc files.


marshallward (approved):

marshallward:
Update: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13275

PR #1451, opened 2021-07-22
Horizontal Regridding MOM_IO

MJHarrison-GFDL:
This removes netCDF API calls contained in MOM_horizontal_regridding and replaces with calls to MOM_io routines.

### DIFF
--- a/config_src/infra/FMS2/MOM_interp_infra.F90
+++ b/config_src/infra/FMS2/MOM_interp_infra.F90
@@ -6,7 +6,8 @@ module MOM_interp_infra
 use MOM_domain_infra,    only : MOM_domain_type, domain2d
 use MOM_time_manager,    only : time_type
 use horiz_interp_mod,    only : horiz_interp_new, horiz_interp, horiz_interp_init, horiz_interp_type
-use mpp_io_mod,          only : axistype, mpp_get_axis_data
+use mpp_io_mod,          only : mpp_axistype=>axistype, mpp_get_axis_data
+use MOM_io_infra,        only : axistype, get_axis_data, set_axis_data
 use time_interp_external_mod, only : time_interp_external
 use time_interp_external_mod, only : init_external_field, time_interp_external_init
 use time_interp_external_mod, only : get_external_field_size
@@ -16,8 +17,8 @@ implicit none ; private
 
 public :: horiz_interp_type, horiz_interp_init
 public :: time_interp_extern, init_extern_field, time_interp_external_init
-public :: get_external_field_info, axistype, get_axis_data
-public :: run_horiz_interp, build_horiz_interp_weights
+public :: get_external_field_info
+public :: run_horiz_interp, build_horiz_interp_weights, get_extern_field_axes
 
 !> Read a field based on model time, and rotate to the model domain.
 interface time_interp_extern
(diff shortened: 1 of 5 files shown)


Hallberg-NOAA (changes requested):
At first glance, these changes mostly look good to me. However, any changes to the interfaces in config_src/infra/FMS1 also have to be mirrored in the corresponding interface in config_src/infra/FMS2.

Hallberg-NOAA (approved):

config_src/infra/FMS2/MOM_io_infra.F90
```diff
@@ -666,19 +674,22 @@ subroutine get_axis_data( axis, dat )
...
-    do i=1,size(axis%ax_data) ; dat(i) = axis%ax_data(i) ; enddo
-  elseif (.not.FMS2_writes) then
-    call mpp_get_axis_data( axis%AT, dat )
-  endif
+  call mpp_get_axis_data( axis%AT, dat )
 
 end subroutine get_axis_data
 
+
+!> Extracts and returns the axis data stored in an axistype.
```
Hallberg-NOAA, line 682: This does not seem like the right description for the subroutine `set_axis_data()`.  Rather it seems to have been copied unmodified from `get_axis_data()`.  Please correct this desciption.

config_src/infra/FMS2/MOM_interp_infra.F90
```diff
@@ -16,8 +17,8 @@ module MOM_interp_infra
 
 public :: horiz_interp_type, horiz_interp_init
 public :: time_interp_extern, init_extern_field, time_interp_external_init
-public :: get_external_field_info, axistype, get_axis_data
```
Hallberg-NOAA, line 19: These changes in the publicly declared interfaces from the FMS2/MOM_interp_infra module do not appear to mirror changes in the FMS1/MOM_interp_infra module.  They need to be kept identical.

Hallberg-NOAA (changes requested):
The additional changes appear to be moving in the right direction, but the new comments describing set_axis_data need to be corrected.  Also, there are changes to the interfaces that are declared to be public from config_src/infra/FMS2/MOM_interp_infra that do not appear to be in line with what is done for FMS1/MOM_interp_infra.  Please correct these issues before this PR is merged into dev/gfdl.

marshallward:
I find it confusing that there is a generic `set_axis_data()` function available to both frameworks, yet only one of them can actually use the function.

Since this function appears to only be accessed inside of the framework, I feel it would be better to make this an internal (possibly private) function exclusive to FMS2 and not part of the framework API.

MJHarrison-GFDL:
The reason set_axis_data exists in the fms2 framework is in order to copy the axis data from the mpp_axis type. In the case of FMS1, axistype is identical to mpp_axis, so it is not necessary to copy the data. I personally do not see a problem making this a null operation in one framework.  The point of adding the warning was so that in case anyone tried to call this routine when using FMS1, they would be made aware that the call can be skipped. 

I could make it private by moving to MOM_interp_infra, but it would involve moving the axistype definition and methods and create a dependency in MOM_io_infra on MOM_interp_infra, which seems backwards.

marshallward:
I know that a different trajectory has been decided for this PR, but wanted to follow up and withdraw my objection to introducing `set_axis_data`.  I had incorrectly though that it was raising an ERROR, not a WARNING.

I have no problem with "no-op" functions like this in one framework if it is (for example) indirectly doing the work in another function.

I would have suggested that you remove the WARNING rather than print it, since stdout is already too noisy.  But given that you're going to rework this PR, it's not necessary.

PR #1453, opened 2021-07-26
Spear ecda

MJHarrison-GFDL:
These updates are supporting the SPEAR online ensemble data assimilation system using MOM6. This update includes minor changes to existing interfaces, as well as a complete implementation of the online incremental updates returned from the ensemble filter routine. A background temperature and salinity increment (e.g. an ensemble climatology of previous increments) read from file can be enabled using the APPLY_TRACER_TENDENCY_ADJUSTMENT option .


MJHarrison-GFDL:
Comment from @Hallberg-NOAA : Why is oda_CS%h initialized to "Angstrom_m" instead of zero? This will  cause the SPEAR data assimilation configuration to fail dimensional consistency tests.

MJHarrison-GFDL:
Added dimensional rescaling for DA thickness initialization.

config_src/infra/FMS1/MOM_interp_infra.F90
```diff
@@ -231,7 +231,7 @@ end subroutine time_interp_extern_3d
 
 !> initialize an external field
 integer function init_extern_field(file, fieldname, MOM_domain, domain, verbose, &
-                                   threading, ierr, ignore_axis_atts )
+                                   threading, ierr, ignore_axis_atts, correct_leap_year_inconsistency )
```
adcroft, line 234: The addition of this argument in FMS1 should be replicated in FMS2 - we have identical APIs between src-level code and infra, even if it is not needed and does nothing. Otherwise, it won't compile with FMS2.

src/ocean_data_assim/MOM_oda_driver.F90
```diff
@@ -514,44 +636,152 @@ subroutine set_analysis_time(Time,CS)
...
 
-  !! switch back to ensemble member pelist
-  !call set_PElist(CS%ensemble_pelist(CS%ensemble_id,:))
+!   !! switch back to ensemble member pelist
+!   !call set_PElist(CS%ensemble_pelist(CS%ensemble_id,:))
 
-  return
-end subroutine save_obs_diff
+!   return
+! end subroutine save_obs_diff
```
marshallward, line 662: Did you want to remove this function?  If not, maybe a comment explaining why it is still here.

config_src/infra/FMS1/MOM_interp_infra.F90
marshallward: 31 characters is awfully long for an argument, even if it is optional.  I think "inconsistency" can be dropped since it's not adding any information about the purpose of the flag here.

src/ocean_data_assim/MOM_oda_driver.F90
```diff
@@ -124,7 +161,7 @@ subroutine init_oda(Time, G, GV, CS)
   type(hor_index_type), pointer :: HI=> NULL()
   type(directories) :: dirs
 
-  type(grid_type), pointer :: T_grid !< global tracer grid
+  type(grid_type), pointer :: T_grid => NULL() !< global tracer grid
```
marshallward, line 164: As identified in #1456, we ought to stop initializing our pointers in subroutines like this, since it implicitly tags them with `save`.

I doubt it is an issue here since the function is probably only called once, but it would be safer to initialize this pointer with an explicit statement.  (Same for `HI`, although it's not part of this PR).

src/ocean_data_assim/MOM_oda_driver.F90
```diff
@@ -233,84 +292,89 @@ subroutine init_oda(Time, G, GV, CS)
...
   ! breaking with the MOM6 convention and using global indices
-  call get_domain_extent(G%Domain,is,ie,js,je,isd,ied,jsd,jed,&
-                         isg,ieg,jsg,jeg,idg_offset,jdg_offset,symmetric)
-  isd=isd+idg_offset; ied=ied+idg_offset
-  jsd=jsd+jdg_offset; jed=jed+jdg_offset
-  !call mpp_get_data_domain(G%Domain%mpp_domain,isd,ied,jsd,jed)
+  !call get_domain_extent(G%Domain,is,ie,js,je,isd,ied,jsd,jed,&
+  !                       isg,ieg,jsg,jeg,idg_offset,jdg_offset,symmetric)
+  !isd=isd+idg_offset; ied=ied+idg_offset ! using global indexing within the DA module
+  !jsd=jsd+jdg_offset; jed=jed+jdg_offset ! TODO:  switch to local indexing? (mjh)
```
marshallward, line 311: Zombie code; remove?

src/ocean_data_assim/MOM_oda_driver.F90
```diff
@@ -438,20 +518,62 @@ subroutine oda(Time, CS)
...
+            mask_z,z_in,z_edges_in,missing_value,.true.,.false.,.false.,.true.)
+
+    ! This should be replaced to use mask_z instead of the following lines
+    ! which are intended to zero land values using an arbitrary limit.
+    fld_sz=shape(T_bias)
+    do i=1,fld_sz(1)
+       do j=1,fld_sz(2)
+          do k=1,fld_sz(3)
+             if (T_bias(i,j,k) .gt. 1.0E-3) T_bias(i,j,k) = 0.0
+             if (S_bias(i,j,k) .gt. 1.0E-3) S_bias(i,j,k) = 0.0
```
marshallward, line 562: Do the 1e-3 terms need dimensional scaling?

src/ocean_data_assim/MOM_oda_driver.F90
```diff
@@ -514,44 +636,152 @@ subroutine set_analysis_time(Time,CS)
...
+
+  call enable_averaging(dt, Time_end, CS%diag_CS)
+  if (CS%id_inc_t > 0) call post_data(CS%id_inc_t, T_inc, CS%diag_CS)
+  if (CS%id_inc_s > 0) call post_data(CS%id_inc_s, S_inc, CS%diag_CS)
+  call disable_averaging(CS%diag_CS)
+
+  call diag_update_remap_grids(CS%diag_CS)
+  call cpu_clock_end(id_clock_apply_increments)
+
+  return
```
marshallward, line 729: This can be removed.

src/ocean_data_assim/MOM_oda_driver.F90
MJHarrison-GFDL: I don't think the tracer values are subject to dimensional scaling. There is a comment that this should be replaced to use mask_z instead, which I would concur with. @feiyulu , would switching to use a logical mask impact your answers?

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13511

Hallberg-NOAA (approved):
I have examined this code, and overall it seems to be correct, so I am approving this PR.

 There is, however, one minor point that we may want to address in a subsequent PR.  The description of the new optional arguments `correct_leap_year_inconsistency` to the two `init_extern_field()` functions, which are inherited from the underlying FMS code, as "If present and true, then allow for leap year inconsistency" is confusing to me. Does this option fix the problem or allow it to persist?  We should clarify these two comments at some point to unambiguously describe what this argument actually does.

adcroft:
> There is, however, one minor point that we may want to address in a subsequent PR. The description of the new optional arguments `correct_leap_year_inconsistency` to the two `init_extern_field()` functions, which are inherited from the underlying FMS code, as "If present and true, then allow for leap year inconsistency" is confusing to me. Does this option fix the problem or allow it to persist? We should clarify these two comments at some point to unambiguously describe what this argument actually does.

Just for reference, the relevant bit of FMS code is https://github.com/NOAA-GFDL/FMS/blob/main/time_interp/time_interp_external.F90#L590-L594 .

Hallberg-NOAA:
There is no description of this `correct_leap_year_inconsistency` argument in time_interp_external.F90, but there is a description of this argument (repeated identically 3 times) further into the FMS code in time_interp.F90, for example at https://github.com/NOAA-GFDL/FMS/blob/a7b4e6d5dce7853366834203107977d3f4a1f0cd/time_interp/time_interp.F90#L492-L499.  However, even after reading this description in the FMS code, I still don't have a clear idea what this argument is intended to do.  Is argument this to allow for datasets with a leap year to be used in models that do not use a leap year (by skipping the extra day of input data), the reverse (by duplicating a day of input data), or something even more obscure to do with the difference between the leap years in Gregorian and Julian calendars? If we are adding an optional argument to MOM6 code, its function should be clearly described.

Hallberg-NOAA:
With the updated comments, this finally makes sense to me. Thank you @MJHarrison-GFDL !

Issue #1455, opened 2021-07-27
Registered modules for diagnostics differ between compilers


adcroft:
The PGI complier is producing different output from Gnu and Intel in available_diags for just two variables, not registering the "_d2" module. I initially thought that because two other disagreed with PGI that PGI was wrong, but `temp_layer_ave` is 1D so there should not be a "_d2" module. Needs investigation.
```
diff --git a/ocean_only/global_ALE/z/available_diags.000000 b/ocean_only/global_ALE/z/available_diags.000000
index 93161600..187ba861 100644
--- a/ocean_only/global_ALE/z/available_diags.000000
+++ b/ocean_only/global_ALE/z/available_diags.000000
@@ -1223,12 +1223,12 @@
     ! standard_name: sea_water_salinity_at_sea_floor
     ! cell_methods: xh:mean yh:mean area:mean
 "temp_layer_ave"  [Used]
-    ! modules: {ocean_model,ocean_model_d2}
+    ! modules: ocean_model
     ! long_name: Layer Average Ocean Temperature
     ! units: degC
     ! cell_methods: zl:mean
 "salt_layer_ave"  [Used]
-    ! modules: {ocean_model,ocean_model_d2}
+    ! modules: ocean_model
     ! long_name: Layer Average Ocean Salinity
     ! units: psu
     ! cell_methods: zl:mean
```

PR #1456, opened 2021-07-28
Fix register_diag_field axes_grp pointer init

marshallward:
Several axes_grp pointers in register_diag_field were null()-initialized
at declaration, e.g.
```
  type(axes_grp), pointer :: axes => null()
```
This implicitly declares the pointers with `save`, which preserves its
value across calls.   These axes will not be initialized to null()
(excepting the first call).

In this instance, this was causing d2 axes to be created when they
should not have existed.

This patch eliminates the implicit `save` attributes and applies
explicit initialization when required.

### DIFF
--- a/src/framework/MOM_diag_mediator.F90
+++ b/src/framework/MOM_diag_mediator.F90
@@ -1969,10 +1969,10 @@ integer function register_diag_field(module_name, field_name, axes_in, init_time
                                                          !! integrated). Default/absent for intensive.
   ! Local variables
   real :: MOM_missing_value
-  type(diag_ctrl), pointer :: diag_cs => NULL()
-  type(axes_grp), pointer :: remap_axes => null()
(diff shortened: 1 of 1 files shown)


marshallward:
This should fix #1455.

Hallberg-NOAA (approved):
I understand and agree with these changes. Moreover, I have verified that the three pointers that are no longer being nullified when they are declared and are not being explicitly nullified in the code are being properly set before they have used.

This PR has passed the TC testing, and I will merge it in as soon as the pipeline tests have run successfully.

adcroft:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/jobs/67881 passed with doc string msgs

marshallward:
Sorry, I updated main -> dev/gfdl before this managed to get in.

Since the PR has already been approved and tested, I will do this merge and update the parameters.

Issue #1419, opened 2021-06-09
MOM_shared_initialization.F90: Unintended land mask changes with use of TOPO_EDITS_FILE, positive MASKING_DEPTH value and shallower depth edits


jr3cermak:
This bug triggered through use of the `TOPO_EDITS_FILE` feature.   A positive `MASKING_DEPTH` with shallower depth edits will cause unintended land mask changes.

```
      if (new_depth(n)/=0.) then
        write(stdout,'(a,3i5,f8.2,a,f8.2,2i4)') &
          'Ocean topography edit: ', n, ig(n), jg(n), D(i,j)/m_to_Z, '->', abs(new_depth(n)), i, j
        D(i,j) = abs(m_to_Z*new_depth(n)) ! Allows for height-file edits (i.e. converts negatives)
      else
        if (topo_edits_change_mask) then
          write(stdout,'(a,3i5,f8.2,a,f8.2,2i4)') &
            'Ocean topography edit: ',n,ig(n),jg(n),D(i,j)/m_to_Z,'->',abs(new_depth(n)),i,j
            D(i,j) = abs(m_to_Z*new_depth(n)) ! Allows for height-file edits (i.e. converts negatives)
        else
          call MOM_error(FATAL, ' apply_topography_edits_from_file: '//&
            "A zero depth edit would change the land mask and is not allowed in"//trim(topo_edits_file))
        endif
```

PR #1457, opened 2021-07-31
Bugfix: a couple of mask_depth related issues

herrwang0:
This PR addresses a couple of bugs related to `MASKING_DEPTH`.

1. In subroutine `limit_topography`, value of `D` over the land is re-assigned to `mask_depth` (previously untouched). This is required for cases with masked-out land PEs. When topography is read from a file, `D` is initialized with a large number (-9.e30) https://github.com/NOAA-GFDL/MOM6/blob/0abf9c2b63e61980382457cc1b448904d7f36784/src/initialization/MOM_shared_initialization.F90#L170 , this number stays in the halo regions of the masked PEs. The fix avoids NaN's in global check-sums.  This is first reported by @favorliao in https://github.com/NOAA-GFDL/MOM6/pull/1428 
2. Local variable `Dmin` in subroutine `initialize_masks` is renamed to `Dmask` to clarify the difference between `min_depth` and `mask_depth`
3. A zero `mask_depth` was implicitly used in subroutine `apply_topography_edits_from_file`. The fix here should address issue #1419

### DIFF
--- a/src/initialization/MOM_grid_initialize.F90
+++ b/src/initialization/MOM_grid_initialize.F90
@@ -1187,7 +1187,7 @@ end function Adcroft_reciprocal
 !> Initializes the grid masks and any metrics that come with masks already applied.
 !!
 !!    Initialize_masks sets mask2dT, mask2dCu, mask2dCv, and mask2dBu to mask out
-!! flow over any points which are shallower than Dmin and permit an
+!! flow over any points which are shallower than Dmask and permit an
 !! appropriate treatment of the boundary conditions.  mask2dCu and mask2dCv
 !! are 0.0 at any points adjacent to a land point.  mask2dBu is 0.0 at
 !! any land or boundary point.  For points in the interior, mask2dCu,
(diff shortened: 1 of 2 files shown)


marshallward (approved):
I agree with these changes, and also with restructuring `initialize_masks` around `mask_depth` rather than `min_depth`.

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13297 :heavy_check_mark: :yellow_circle:

adcroft:
@herrwang0 That's a relief! I was worried you were using the "old" pathway with MASKING_DEPTH=-9999. The changed code block with `0.5*min_depth` inside that old path made me wonder if you were using it.

adcroft:
I agree, but you don't reach either half if you set MASKING_DEPTH - the whole of that excerpt is only for when MASKING_DEPTH is not set. I think you're the only one trying so far and I think using MASKING_DEPTH as you are if the better pathway.

herrwang0:
@adcroft Thanks so much for doing this! It is complicated with these two somewhat independent variables here, and extra work is needed to take into consideration of almost all the scenarios, to make all parties happy.

PR #1458, opened 2021-08-03
fix internal wave refraction

raphaeldussin:
current code uses halo (or land) points in the computation of the gradient of Internal Wave group velocity, which results in spurious gradients and unexpected refraction close to boundaries. This PR computes IW group velocities on U and V points, discarding contribution from halo or land, before computing the gradient.


src/parameterizations/lateral/MOM_internal_tides.F90
```diff
@@ -779,9 +781,36 @@ subroutine refract(En, cn, freq, dt, G, US, NAngle, use_PPMang)
...
   integer :: is, ie, js, je, asd, aed, na
   integer :: i, j, a
+  real :: wgt1, wgt2
+  real :: eps
 
   is = G%isc ; ie = G%iec ; js = G%jsc ; je = G%jec ; na = size(En,3)
   asd = 1-stencil ; aed = NAngle+stencil
+  eps=1.0e-20 * US%m_s_to_L_T
+
+  do i=is-1,ie ; do j=js-1,je
```
Hallberg-NOAA, line 791: This loop extent would make sense for variables that are discretized at a corner point, but not at a u-point. I suspect that the correct extent here should be
`do I=is-1,ie ; do j=js,je`
 based on how cn_u is used in line 845 below.

src/parameterizations/lateral/MOM_internal_tides.F90
```diff
@@ -813,16 +842,13 @@ subroutine refract(En, cn, freq, dt, G, US, NAngle, use_PPMang)
                    (G%CoriolisBu(I,J-1) + G%CoriolisBu(I-1,J)))
       df_dx = 0.5*((G%CoriolisBu(I,J) + G%CoriolisBu(I,J-1)) - &
                     (G%CoriolisBu(I-1,J) + G%CoriolisBu(I-1,J-1))) * G%IdxT(i,j)
-      dlnCn_dx = 0.5*( G%IdxCu(I,j) * (cn(i+1,j) - cn(i,j)) / &
-                       (0.5*(cn(i+1,j) + cn(i,j)) + cn_subRO) + &
-                       G%IdxCu(I-1,j) * (cn(i,j) - cn(i-1,j)) / &
-                       (0.5*(cn(i,j) + cn(i-1,j)) + cn_subRO) )
+      dlnCn_dx = G%IdxT(i,j) * (cn_u(I,j) - cn_u(I-1,j)) / (cn(i,j) + cn_subRO)
```
Hallberg-NOAA, line 845: Even with the cn_subR0, using just a single value for cn(i,j) in the denominator could be dangerously large if that value is 0 (or very small) but one of the adjacent values of cn_u in the numerator is non-zero. Perhaps it might make sense to replace the denominator with `(0.5*cn(i,j) + 0.25*(cn(i-1,j) + cn(i+1,j)) + cn_subRO)`, and similarly for the denominator of dlnCn_dy below.

src/parameterizations/lateral/MOM_internal_tides.F90
raphaeldussin: fixed

src/parameterizations/lateral/MOM_internal_tides.F90
raphaeldussin: I think this would re-introduce the problematic values from the halos or land, instead I propose to test for values of cn

Hallberg-NOAA: The updated code addresses this issue very nicely.

marshallward:
After a chat with @raphaeldussin we came up with the following recommended changes:

* Loops at L799 and L811 should swap the `i` and `j` loops, so that `i` is innermost

* The `c / (wgt1 + wgt2)` should probably replaced with a precomputed mask containing 0., 1., or 0.5.  
  Two benefits:
  
  * Eliminate the floating point division (since it's unlikely a compiler will 
  * Remove the if-blocks with some sort of mask

  But at the moment I don't have a good suggestion on how to efficiently construct this mask.
  
We will sit on this one for a bit while Raf looks into some other issues, and will settle on a solution when that other work is finished.

marshallward:
The method by @raphaeldussin for computing `cn_[uv]` looks good to me :+1:

marshallward (approved):

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13373 :heavy_check_mark:

PR #1459, opened 2021-08-03
Fix to commit 893a06e062 which broke the SIGMA_SHELF_ZSTAR option

MJHarrison-GFDL:
- Testing with an ice shelf (e.g. ISOMIP) and using
    the SIGMA_SHELF_ZSTAR coordinate would have caught this bug.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13301 :heavy_check_mark:

marshallward (approved):

PR #1462, opened 2021-08-09
Bug fix for FMS2 issue #761, broadcast from root pe

nikizadehgfdl:
- This addresses the FMS issue $761
https://github.com/NOAA-GFDL/FMS/issues/761

- There is a mpp_broadcast in the FMS2 subroutine
get_unlimited_dimension_name() and this subroutine has to be called by
all pes, so it cannot be inside a if(is_root_pe()) block

### DIFF
--- a/src/framework/MOM_restart.F90
+++ b/src/framework/MOM_restart.F90
@@ -1092,7 +1092,6 @@ subroutine restore_state(filename, directory, day, G, CS)
 
   ! Check the remaining files for different times and issue a warning
   ! if they differ from the first time.
-  if (is_root_pe()) then
     do m = n+1,num_file
       call get_file_times(IO_handles(n), time_vals, ntime)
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA (changes requested):
Although I agree that this PR fixes a problem that was introduced by FMS2, I think that we do not want any processors apart from the root PE issuing any warning messages, so that we are not deluged with a flood of redundant messages.  Please add a second expression to the logical test that is now on line 1102, so that it reads `if ((t1 /= t2) .and. is_root_PE()) then`, to avoid excessive numbers of warnings.

Hallberg-NOAA (approved):

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13336

PR #1467, opened 2021-08-15
MOM6: +Preparations to allow reference heights to change

Hallberg-NOAA:
This PR includes various commits that prepare to allow for the reference
heights for bathymetry and interfaces to be changed, or corrects problems that
were exposed in the process of developing that capability, but which do not
themselves change any answers or algorithms.  This set of changes includes the
addition of several new optional arguments, and it adds comments noting (but
does not fix) some buts that seem likely to change answers.  These new
capabilities have been tested and work, but they will not be exercised until the
subsequent commit that implements the ability to change reference heights.  All
answers and output are bitwise identical with the changes in this PR.

  The commits in this PR include:

- NOAA-GFDL/MOM6@5017bf6c6 Added comments highlighting bugs
- NOAA-GFDL/MOM6@c72d441ea (*)Fix dimensionally inconsistent MEKE beta calcs
- NOAA-GFDL/MOM6@62446ec0f Cleanup of comments in MOM_ALE_sponge.F90
- NOAA-GFDL/MOM6@7cc058291 Prep initialization for reference height changes
- NOAA-GFDL/MOM6@e4ffa7654 Corrects the routine indicated in 2 error messages
- NOAA-GFDL/MOM6@4d8d7afc7 +Add optional Z_0p argument to int_density_dz
- NOAA-GFDL/MOM6@db43b353f Longer error string in get_polynomial_coordinate
- NOAA-GFDL/MOM6@a2df3b75e Code cleanup in MOM_regridding
- NOAA-GFDL/MOM6@6f1a19173 +Add optional dZref argument to find_eta

### DIFF
--- a/src/ALE/MOM_regridding.F90
+++ b/src/ALE/MOM_regridding.F90
@@ -880,21 +880,20 @@ subroutine check_remapping_grid( G, GV, h, dzInterface, msg )
 
   !$OMP parallel do default(shared)
   do j = G%jsc-1,G%jec+1 ; do i = G%isc-1,G%iec+1
-    if (G%mask2dT(i,j)>0.) call check_grid_column( GV%ke, GV%Z_to_H*G%bathyT(i,j), h(i,j,:), dzInterface(i,j,:), msg )
+    if (G%mask2dT(i,j)>0.) call check_grid_column( GV%ke, h(i,j,:), dzInterface(i,j,:), msg )
   enddo ; enddo
 
 end subroutine check_remapping_grid
 
 !> Check that the total thickness of new and old grids are consistent
-subroutine check_grid_column( nk, depth, h, dzInterface, msg )
+subroutine check_grid_column( nk, h, dzInterface, msg )
   integer,               intent(in) :: nk !< Number of cells
-  real,                  intent(in) :: depth !< Depth of bottom [Z ~> m] or arbitrary units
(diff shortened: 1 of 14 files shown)


herrwang0:
There was a parameter called `Z_ref` in `ocean_grid_type` introduced a couple years ago. 
https://github.com/NOAA-GFDL/MOM6/blob/e88cdaa785b9fab4e24286031bdd409caf673034/src/core/MOM_grid.F90#L213

It was only used once in the barotropic solver (calculating `IDat[uv]`). Does it make sense to utilize that for changing reference height in various subroutines (instead of optional input)? or do we want to make the variable more explicit  as subroutine input argument (and maybe allow it to vary across subroutines)?

Hallberg-NOAA:
@herrwang0 , the plan is to use exactly this `G%Z_ref` variable in numerous places in the code in an upcoming PR, but before doing so there are several changes that we will be making to correct bugs or otherwise minimize the locations where this variable will be used.  The current PR is one of these preliminary steps.

marshallward (approved):
Adds some new computational work (e.g. MOM_density_integrals) but the offset is a fixed scalar and does not require array access, so I would not expect this to affect runtime.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13413

Issue #1447, opened 2021-07-19
Trouble with negative depths in land mask


kshedstrom:
In btstep, the surface height is compared to bathyT and a warning is issued if the water depths go negative. This can be triggered in the land where if one asks for a depth of -5 for tidal wetting and drying applications. SSH eta is 0.0 inside the mask which makes for a dry spot with that depth. It also causes trouble for the barotropic wave speed calculation based on depth.

In ROMS, when wetting and drying is turned on, SSH is always at least some D_crit above the bottom, even inside the land mask. The waves speeds use full water depth, including the contribution from SSH.

Hallberg-NOAA:
In examining the code, it looks like btstep no longer assumes that is 0 a special geopotential height, and that the total depth is always the sum of bathyT+eta whereever it used in Boussinesq mode. Would it be sufficient to apply a land-mask before issuing a warning on lines 2358 and 2363 of MOM_barotropic.F90 to correct this issue in your case, or did you have something more extensive in mind, @kshedstrom ?

  Also, there are several different places where external wave speeds are calculated in the barotropic solver. Could you be more specific about which one is becoming problematic?

kshedstrom:
It is this one, this time:
```
#1  0x4df7a8 in __mom_open_boundary_MOD_update_obc_segment_data
	at //import/c1/AKWATERS/kate/ESMG/ESMG-configs/src/MOM6/src/core/MOM_open_boundary.F90:3820
#2  0x1ba4430 in __mom_boundary_update_MOD_update_obc_data
	at //import/c1/AKWATERS/kate/ESMG/ESMG-configs/src/MOM6/src/core/MOM_boundary_update.F90:136
```

kshedstrom:
As for your other question, yes, applying a land mask to the checks would be sufficient.

Hallberg-NOAA:
When I look at what is happening in the relevant code block where this is failing, https://github.com/NOAA-GFDL/MOM6/blob/dev/gfdl/src/core/MOM_open_boundary.F90#L3820-L3825, it strikes me that 
if we were to replace the existing code:

```
        segment%Cg(i,J) = sqrt(GV%g_prime(1)*G%bathyT(i,j+jshift))
        segment%Htot(i,J)=0.0
        do k=1,GV%ke
          segment%h(i,J,k) = h(i,j+jshift,k)
          segment%Htot(i,J)=segment%Htot(i,J)+segment%h(i,J,k)
        enddo
```
with 

[...]

PR #1470, opened 2021-08-18
(*)Use summed thicknesses in OBC wave speed

Hallberg-NOAA:
(*)Use summed thicknesses in OBC wave speed

  Changed the calculation of the external mode gravity wave speed used by the
radiation open boundary conditions and in the Kelvin-wave test case to use the
summed layer thicknesses rather than bathyT.  This change will prevent taking
the square root of negative thicknesses, even if there is wetting and drying so
some of the bathymetry is above the mean sea level, and it is more physically
accurate.  This PR will at least partly address MOM6 issue #1447.  Some
solutions will exhibit changing answers, including the barotropic Kelvin wave
test case, but after a broad discussion it was decided not to reproduce the
previous solutions with a runtime parameter.

  This single-commit PR builds on PR#1469, and should only be dealt with after
that one is merged into dev/gfdl.

### DIFF
--- a/src/ALE/MOM_regridding.F90
+++ b/src/ALE/MOM_regridding.F90
@@ -880,21 +880,20 @@ subroutine check_remapping_grid( G, GV, h, dzInterface, msg )
 
   !$OMP parallel do default(shared)
   do j = G%jsc-1,G%jec+1 ; do i = G%isc-1,G%iec+1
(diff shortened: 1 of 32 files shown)


kshedstrom:
I approve this PR. So many changes!

marshallward (approved):

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13429 :heavy_check_mark:

PR #1474, opened 2021-08-20
Updates from dev/cpt 08/20/21

gustavo-marques:
This PR includes all the developments been done in dev/cpt since Dec 23, 2020. The main updates are summarized below:

* Many updates to the nw2_tracers module;
* implements modified diagnostics for the momentum budget (https://github.com/ocean-eddy-cpt/MOM6/pull/10);
* Remove MEKE_damping from use_drag_rate so we can use linear dissipation without bottom drag. This is needed for the
GEOMETRIC scheme (https://github.com/ocean-eddy-cpt/MOM6/pull/11).


src/tracer/nw2_tracers.F90
```diff
@@ -0,0 +1,314 @@
...
+public nw2_tracer_column_physics
+public nw2_tracers_end
+
+!> The control structure for the nw2_tracers package
+type, public :: nw2_tracers_CS ; private
+  integer :: ntr = 0  !< The number of tracers that are actually used.
+  type(time_type), pointer :: Time => NULL() !< A pointer to the ocean model's clock.
+  type(tracer_registry_type), pointer :: tr_Reg => NULL() !< A pointer to the tracer registry
+  real, pointer :: tr(:,:,:,:) => NULL()   !< The array of tracers used in this package, in g m-3?
+  real, allocatable , dimension(:) :: restore_rate !< The exponential growth rate for restoration value [year-1].
```
Hallberg-NOAA, line 36: From the code, it appears that this damping rate is in [s-1]. not [year-1]. Also, to more fully implement the dimensional consistency testing and help in the detection of inconsistent expressions, I recommend that this be changed to [T-1 ~> s-1] by moving the conversion factor up to the point where restore_rate is first set.

src/parameterizations/lateral/MOM_MEKE.F90
```diff
@@ -175,8 +175,7 @@ subroutine step_forward_MEKE(MEKE, h, SN_u, SN_v, visc, dt, G, GV, US, CS, hu, h
   if (.not.associated(MEKE)) call MOM_error(FATAL, &
          "MOM_MEKE: MEKE must be initialized before it is used.")
 
-  if ((CS%MEKE_damping > 0.0) .or. (CS%MEKE_Cd_scale > 0.0) .or. (CS%MEKE_Cb>0.) &
```
Hallberg-NOAA, line 178: I do not think that the logic change here is correct.   This change will cause lines 517-532 not to be executed, even if CS%MEKE_damping is positive, which I think is wrong.

src/tracer/nw2_tracers.F90
adcroft: Yes, it is in [s-1]. This is a hold over from the other tracer code using [yr-1].

src/parameterizations/lateral/MOM_MEKE.F90
gustavo-marques: Thanks, @Hallberg-NOAA. 
@sdbachman and I have modified the code to address your comment.  Now, drag_rate is always initialized, either to zero or to the bottom drag formula. We also updated the logic to always execute the second Strang splitting calculation. We noticed that drag_rate_J15 is never used, so we deleted this variable. We also deleted some commented code.

gustavo-marques: It seems the last commit is changing answers. I will look into this and report back here.

gustavo-marques: I accidentally deleted some code which led to an answer change (thankfully CI tests caught that). After re-introducing the code, all tests are passing.

Hallberg-NOAA: With the recent additional changes, I am satisfied that this issue has been addressed.

src/tracer/nw2_tracers.F90
Hallberg-NOAA: Let us accept this PR as it currently stands, and then fix this very minor issue in a subsequent PR.

Hallberg-NOAA (approved):
With the corrections to this PR, I am satisfied that it should be accepted, while still noting that there is a minor issue with one of the comments mis-documenting the units of a variable that will be addressed in a subsequent small PR.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13506

PR #1480, opened 2021-08-26
User/bgr/homogenize iob

breichl:
These changes are necessary to run a single-column like configuration in ice-ocean mode.  It is difficult to guarantee that the coupler passes perfectly homogeneous fields back to MOM, so it is more straightforward to force the IOB fields to be homogeneous inside of MOM.

Presently there are no test cases that utilize this code, so everything should be unchanged.  A new test case will be added shortly to take advantage of this new option in a SCM OM4-like configuration.

@Hallberg-NOAA , looking for some advice if the homogenize_IOB and homogenize_field routines would be better in another module, or if they are okay here.

### DIFF
--- a/config_src/drivers/FMS_cap/ocean_model_MOM.F90
+++ b/config_src/drivers/FMS_cap/ocean_model_MOM.F90
@@ -25,6 +25,7 @@ use MOM_diag_mediator, only : diag_ctrl, enable_averaging, disable_averaging
 use MOM_diag_mediator, only : diag_mediator_close_registration, diag_mediator_end
 use MOM_domains, only : MOM_domain_type, domain2d, clone_MOM_domain, get_domain_extent
 use MOM_domains, only : pass_var, pass_vector, AGRID, BGRID_NE, CGRID_NE, TO_ALL, Omit_Corners
+use MOM_domains, only : sum_across_PEs
 use MOM_error_handler, only : MOM_error, MOM_mesg, FATAL, WARNING, is_root_pe
 use MOM_error_handler, only : callTree_enter, callTree_leave
 use MOM_EOS, only : gsw_sp_from_sr, gsw_pt_from_ct
@@ -164,7 +165,8 @@ type, public :: ocean_state_type ; private
                               !! fields necessary to integrate only the tracer advection
                               !! and diffusion equation read in from files stored from
                               !! a previous integration of the prognostic model.
-
+  logical :: use_homogenized_IOB !< Creates a homogenized version of the IOB structure
+                              !! that yields homogeneous forcing/fluxes.
   logical :: single_step_call !< If true, advance the state of MOM with a single
                               !! step including both dynamics and thermodynamics.
                               !! If false, the two phases are advanced with
@@ -358,6 +360,10 @@ subroutine ocean_model_init(Ocean_sfc, OS, Time_init, Time_in, wind_stagger, gas
     use_melt_pot=.false.
   endif
 
+  call get_param(param_file, mdl, "HOMOGENIZE_IOB_FORCINGS", OS%use_homogenized_IOB, &
+                 "If true, creates a copy of all IOB fluxes and averages them "//&
+                 "horizontally across the entire domain.", default=.false.)
+
   !allocate(OS%sfc_state)
   call allocate_surface_state(OS%sfc_state, OS%grid, use_temperature, do_integrals=.true., &
                               gas_fields_ocn=gas_fields_ocn, use_meltpot=use_melt_pot)
@@ -446,6 +452,8 @@ subroutine update_ocean_model(Ice_ocean_boundary, OS, Ocean_sfc, time_start_upda
   real,    optional, intent(in)    :: cycle_length !< The duration of a coupled time stepping cycle [s].
 
   ! Local variables
+  type(ice_ocean_boundary_type) :: Ice_ocean_boundary_homogenized ! A copy of the structure containing the
+                            ! forcing fields coming from the ice and atmosphere, that is horizontally averaged.
   type(time_type) :: Time_seg_start ! Stores the dynamic or thermodynamic ocean model time at the
                             ! start of this call to allow step_MOM to temporarily change the time
                             ! as seen by internal modules.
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA (changes requested):
I like the spirit of where this PR is going, and it looks like it should work as written although I do not think that this code will be rotationally invariant.  I have two specific questions about the implementation here:

  1) Why is there the choice to homogenize the Ice_ocean_boudary_type, rather than the forces and fluxes types?  The former requires the creation of a separate structure, and it is specific to the FMS_cap, whereas if the homogenization had been to forcing and mech_forcing types, it could have been done in place, and this capability would have occurred in MOM6/src code that would have been shared across all the various drivers for MOM6.

  2) When taking the global means in `homogenize_field()`, why not use the order-invariant sums via something like the function `global_area_mean()` in src/diagnostics/MOM_spatial_means.F90, which would be rotationally invariant?

breichl:
Thanks @Hallberg-NOAA for looking at this and making comments.  I will take suggestion in the second comment and update this PR.  We can continue the conversation here or offline about your first comment.

1.  I planned to homogenize the fields in the forces/fluxes types, there were two reasons I changed my mind.  First, it requires some care to ensure that all the homogenized fields are self-consistent due to non-commuting operators in the forces/fluxes extractions (ustar/tau being a primary example where the homogenized tau would need to be computed multiple times or stored).  Second, there are many more fields that would need to be homogenized in the forces/fluxes type.  It became clear it would be significantly less effort both now and in the future to homogenize the IOB type, and it also ensures that the boundary conditions that MOM sees are self-consistent.  I'm happy to discuss more if you think the additional effort is worthwhile to find a more general approach.
2. I was not aware of this function, but will look at replacing this.  I wasn't positive that this would satisfy the rotational invariance, hopefully this simplifies that concern.  FYI, I modeled this code after a function to homogenize the T&S initialization (following the homogenize flag in MOM_state_initialization).  That code might therefore also be improved by taking advantage of this function.

config_src/drivers/FMS_cap/ocean_model_MOM.F90
```diff
@@ -1194,4 +1220,174 @@ subroutine ocean_model_get_UV_surf(OS, Ocean, name, array2D, isc, jsc)
...
+  if (associated(IOB%mi)) then
+    allocate(IOB_h%mi (isc_bnd:iec_bnd,jsc_bnd:jec_bnd) ) ; IOB_h%mi = 0.0
+    call homogenize_field(IOB%mi,IOB_h%mi, G, index_bounds)
+  endif
+
+  if (associated(IOB%ice_rigidity)) then
+    allocate(IOB_h%ice_rigidity (isc_bnd:iec_bnd,jsc_bnd:jec_bnd) ) ; IOB_h%ice_rigidity = 0.0
+    call homogenize_field(IOB%ice_rigidity,IOB_h%ice_rigidity, G, index_bounds)
+  endif
+
```
nikizadehgfdl, line 1357: @breichl , what about other IOBs? When I run this code change with "generic tracers" turned on the model crashes with 
```
FATAL: extract_coupler_values: The requested boundary condition is not associated.

Image              PC                Routine            Line        Source             
MOM6SIS2           00000000026A4A77  Unknown               Unknown  Unknown
MOM6SIS2           0000000001A421EC  mpp_mod_mp_mpp_er          68  mpp_util_mpi.inc
MOM6SIS2           00000000017767FC  fms_coupler_util_          53  FMS_coupler_util.F90
MOM6SIS2           00000000004E20B6  g_tracer_utils_mp        1429  generic_tracer_utils.F90
MOM6SIS2           00000000011995D0  ocean_model_mod_m         551  ocean_model_MOM.F90
MOM6SIS2           00000000013E3A6F  MAIN__                   1041  coupler_main.F90

```

config_src/drivers/FMS_cap/ocean_model_MOM.F90
```diff
@@ -504,9 +512,17 @@ subroutine update_ocean_model(Ice_ocean_boundary, OS, Ocean_sfc, time_start_upda
   ! Translate Ice_ocean_boundary into fluxes and forces.
   call get_domain_extent(Ocean_sfc%Domain, index_bnds(1), index_bnds(2), index_bnds(3), index_bnds(4))
 
+  if (OS%use_homogenized_IOB) &
+    call homogenize_IOB(Ice_ocean_boundary, Ice_ocean_boundary_homogenized, OS%grid, index_bnds)
```
nikizadehgfdl, line 516: @breichl , what is the reason for doing this on an extra copy of IOB? I think if we homogenize the fields on the origin Ice_ocean_boundary (rather than the new Ice_ocean_boundary_homogenized ) everything would work fine, including the (generic) tracer fluxes. The problem with  Ice_ocean_boundary_homogenized is that it does not have the hidden member "fluxes" allocated.

nikizadehgfdl: One issue with modifying IOB itself is that it is intent(in) in update_ocean_model, I had to make it intent(inout). But after that all seem to have worked fine, including the homogenization and the generic tracers.

breichl: Okay, thanks Niki.  I made the copy to avoid modifying this structure since it is intent(in).  If we are comfortable making it intent(inout) then this solution works for me.  Do the generic tracers fields also need to be homogenized?  I'm not sure how difficult that would be or if it is necessary here.

config_src/drivers/FMS_cap/ocean_model_MOM.F90
Hallberg-NOAA: I think that this issue that Niki is encountering suggests that it might be worth revisiting whether it would be simpler to homogenize the fields in the forcing and mech_forcing types, even if we had to add the option to recalculate the dependent fields (e.g., ustar) based on the primary fields (like taux and tauy).  We should not be changing the intent on the various types.

breichl:
Provided @Hallberg-NOAA thinks #1541 is a better solution I will close this PR

breichl:
Closing in favor of #1541.

Issue #1466, opened 2021-08-15
Spatially inconsistent expression for GM+E backscatter coefficient


Hallberg-NOAA:
The expression for the GM+E backscatter coefficients at the corner (q-) points at
https://github.com/NOAA-GFDL/MOM6/blob/e88cdaa785b9fab4e24286031bdd409caf673034/src/parameterizations/lateral/MOM_hor_visc.F90#L1407-L1413
is not properly collocated in space and is not rotationally invariant.  It uses the bathymetry with the same indices as for the other arrays, but this bathymetry is descretized a half-gridpoint to the south and west of the rest of this expression.

  To correct this problem, something like the maximum, the minimum, or the 4-point average of the four neighboring bathymetric depths should be used. @gustavo-marques or @sdbachman, do you have any thoughts on what transformation of the 4 neighboring bathymetries to use?

  Correcting this problem seems likely to change answers for any cases that have `USE_GME = True`.

sdbachman:
The use of the bathymetric depth in that expression is an ad-hoc way of
tapering the backscatter strength in shallow water.  The discretization
should be changed in whatever way is convenient to ensure rotational
invariance and collocation with the other arrays; there is nothing special
about the way it is done at present.

I don't think we have any important cases right now that have USE_GME =
True, so fixing this shouldn't cause us any problems.

gustavo-marques:
@Hallberg-NOAA, thanks for finding this issue and letting us know. I can confirm that we do not have any case with `USE_GME = True`.

PR #1481, opened 2021-08-26
(*)Fix discretization issues with USE_GME=True

Hallberg-NOAA:
Corrected the inconsistently staggered expressions for the total depths used
to scale away the GM+E backscatter scheme that is enabled with USE_GME.  As a
part of these changes, the nominal bathymetric depths have been replaced by the
sum of the layer thicknesses, and parentheses were added to some of the GME
expressions so that they will be rotationally invariant.  Although there are no
changes to the answers in the MOM6-examples test suite, answers will change in
any cases where USE_GME is true, and GME_H0 is larger than the minimum ocean
depth.  This commit will address MOM6 issue #1466, which can now be closed.


marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13477

marshallward (approved):

PR #1482, opened 2021-08-26
(*)Use depths in DOME and dye tracer initialization

Hallberg-NOAA:
Initialize the horizontal stripes of tracers in the DOME and dye examples
based on the depth from the (flat) sea surface rather than the height above the
seafloor.  These are mathematically equivalent, but there could be small changes
in the tracer distributions.  Because these changes only impact passive tracers,
all of the physical solutions are bitwise identical.


marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13479 :heavy_check_mark:

marshallward (approved):

PR #1483, opened 2021-08-27
(*+)New internal tide bathymetric parameters

Hallberg-NOAA:
Replaced the nominal bathymetric depth with the sum of the layer thicknesses
in the calculation of the quadratic drag acting on the internal wave field.
Also replaced the hard-coded 1 m scale at which the drag is reduced with a new
runtime parameter, INTERNAL_TIDE_DRAG_MIN_DEPTH, which is only logged if the
internal tides are used and INTERNAL_TIDE_QUAD_DRAG=True, and made the maximum
scale of the topographic roughness as a fraction of the bottom depth into the
new runtime parameter INTERNAL_TIDE_ROUGHNESS_FRAC, whose default, 0.1, is the
same as the previous hard-coded value.  Several of the comments in the
MOM_internal_tides module were also clarified or standardized.  All answers in
the MOM-examples test suite are bitwise identical, but answers could change and
there could be changes in the MOM_parameter_doc files in some cases that use the
internal tides module.


marshallward (approved):

marshallward:
Gaea Regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines :heavy_check_mark:

PR #1484, opened 2021-08-27
+(*)Add option for MEKE to calculate total depth

Hallberg-NOAA:
Added the new run-time option, MEKE_FIXED_TOTAL_DEPTH, to use the actual total
thickness instead the nominal depth in bathyT in several of the MEKE
calculations.  Also simplified and corrected a minor dimensional inconsistency
in some expressions that effectively set masks when estimating the interface
height diffusivities at tracer points for use in MEKE.  By default, the answers
in the MOM-examples test cases are bitwise identical, but answers could change
in some cases due to the proper thickness weighting in the calculation of
MEKE%Kh_diff.  There are new entries in some MOM_parameter_doc files, so a minor
spelling error correction in an unrelated MOM_parameter_doc entry was also included
in this PR.


marshallward:
Gaea Regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13495

marshallward (approved):

PR #1485, opened 2021-08-27
Testing: Disable 2018 answer flags

marshallward:
The TC tests are still using 2018 answer reproducibility flags, which
fail when aggressive initialization of allocatable arrays is enabled.
In at least one case, the `mask_z` variable of the ALE sponge is
incompletely initialized when `DEFAULT_2018_ANSWERS` is set.  When
unset, the array is fully initialized.

This patch disables the 2018 answer flags and restores reproducibility
on such platforms.


Hallberg-NOAA (approved):
These changes shift the TC testing to emphasize the newer code that we want people to be using, although it does not preclude the use of the older algorithms in existing experiments, which will be retained as there is any demand to use them. I agree with these changes.

PR #1488, opened 2021-08-28
Testing: Rotation support for MOM_read_data

marshallward:
This patch adds an interface and several implementations of
`MOM_read_data` and `MOM_read_vector` functions which support rotational
testing.

The `MOM_domain_type` now carries a scalar, `turns`, indicating the
number of quarter-turns from the input grid to the model grid, and a
pointer to the original unrotated `MOM_domain`.  If this number is
nonzero, then the input field is read into an array based on the input
grid, and then rotated to a new array based on the model grid.  This
final result is returned by the function.

If the domain's `turns` is zero, then it is assumed to be a call from a
non-rotated domain and no rotation is applied.  Functions outside of MOM
(such as calls within drivers) do not apply this rotation.

For the "domain-less" reads of 2d arrays, an explicit `turns` argument
is supported.  This only appears to be necessary in one part of grid
initialization.

This is now the third place where `turns` is tracked: the first is HI
(horizontal index tracking) of the MOM grid, the second in `restart_CS`,
and now in `MOM_Domain`.  However, I believe this is a reasonable (if
not necessary) place to track the domains while also avoiding a need for
users to explicitly rotate fields every time `read_data` is called.


marshallward:
The previous version of this did not include the pointer to the unrotated domain, so was incorrectly trying to use the rotated domain on unrotated inputs.

adcroft (approved):

Hallberg-NOAA:
The PGI compiler does not appear to be able to differentiate between the various new interfaces, which is a problem that would have to be resolved before we could merge this PR into dev/gfdl.

marshallward:
The dual definitions of `MOM_read_data` in `src/framework/MOM_io.F90` and `config_src/infra/FMSx/MOM_io_infra.F90` appear to be tripping up the PGI compiler.

Even though we locally redefine the interface to `read_data_infra`, it looks like PGI cannot quite get there and thinks we've defined two interfaces.

I guess something will need to be renamed.  I'd suggest renaming `config_src/infra` `MOM_read_data` to `read_data_infra` but I am open to any other suggestions.

marshallward:
As discussed outside the issue, I have renamed the instances of `MOM_read_data` in `MOM_io_infra` to `read_field`.  These changes appear to work with PGI.

Hallberg-NOAA:
This updated PR has passed TC testing and pipeline testing with Pipeline #13545, and I agree with the updated code after visual inspection.

PR #1490, opened 2021-08-31
(*)Offset G%bathyT by -G%Z_ref

Hallberg-NOAA:
Added code to adjust the value of G%bathyT in the ocean_grid_type by -G%Z_ref,
with changes scattered throughout the code to compensate.  This does not impact
the value of bathyT in the dyn_hor_grid.  Using a non-zero value of
REFERENCE_HEIGHT leads to mathematically equivalent solutions that differ at
round-off but are demonstrably similar for values of REFERENCE_HEIGHT ranging
from -1e4 m to 1e4 m, and in cases that do not use the less exact
..._2018_ANSWERS algorithms the solutions are qualitatively equivalent for
values ranging from -1e8 m to 1e8 m.  (For these more extreme values, the 64-bit
roundoff in the free surface height calculation starts to become physically
significant at of order 0.1 mm.)  This new capability is useful for developing
wetting and drying and for identifying algorithmic shortcomings, but because
answers are not bitwise identical for various reference heights, it is not such
a good candidate for fully automated testing.  By default, all answers are
bitwise identical and all output files are the same.


src/core/MOM_barotropic.F90
```diff
@@ -4163,31 +4174,27 @@ subroutine find_face_areas(Datu, Datv, G, GV, US, CS, MS, eta, halo, add_max)
 !$OMP do
     do j=js-hs,je+hs ; do I=is-1-hs,ie+hs
       Datu(I,j) = CS%dy_Cu(I,j) * GV%Z_to_H * &
-                 (max(CS%bathyT(i+1,j), CS%bathyT(i,j)) + add_max)
+                 max(max(CS%bathyT(i+1,j), CS%bathyT(i,j)) + (G%Z_ref + add_max), 0.0)
```
herrwang0, line 4177: Related to `Dat[uv]`, these two lines below also need to be changed. They can be problematic if both cells are active land points (negative depth) or they cancel each other out (e.g. bathyT(i,j) = 1 and bathyT(i+1,j) = -1).

https://github.com/NOAA-GFDL/MOM6/blob/060c41290bcb4a3312bdc2bab419fb59dd473c3f/src/core/MOM_barotropic.F90#L4942 

https://github.com/NOAA-GFDL/MOM6/blob/060c41290bcb4a3312bdc2bab419fb59dd473c3f/src/core/MOM_barotropic.F90#L4949

src/core/MOM_barotropic.F90
```diff
@@ -4163,31 +4174,27 @@ subroutine find_face_areas(Datu, Datv, G, GV, US, CS, MS, eta, halo, add_max)
...
       Datv(i,J) = CS%dx_Cv(i,J) * GV%Z_to_H * &
-                 (max(CS%bathyT(i,j+1), CS%bathyT(i,j)) + add_max)
+                 max(max(CS%bathyT(i,j+1), CS%bathyT(i,j)) + (G%Z_ref + add_max), 0.0)
     enddo ; enddo
   else
 !$OMP do
     do j=js-hs,je+hs ; do I=is-1-hs,ie+hs
-      Datu(I, j) = 0.0
-      !Would be "if (G%mask2dCu(I,j)>0.) &" is G was valid on BT domain
-      if (CS%bathyT(i+1,j)+CS%bathyT(i,j)>0.) &
```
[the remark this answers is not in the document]
Hallberg-NOAA: The problem here is that the barotropic domain may have much wider halos than those that are found in the ocean_grid_type, so there may be instances where we are working on points for which G%mask2dCu(I,j) is outside of the bounds of that array.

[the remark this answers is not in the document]
herrwang0: Thanks!
Sorry, I just realized it was a deleted line (was wondering about this for a while, though). The changed code is much neater.

src/core/MOM_barotropic.F90
Hallberg-NOAA: Good catch, @herrwang0 .  These lines did not need to change to incorporate G%Z_ref because the Mean_SL already included adjustments to the reference height, so they were overlooked.  Do you have a specific change that you think would work and have already tested to avoid divisions by zero or effectively negative drag?
  If not, what would you think of changing the denominator of the first line to something like `... / ((max(G%bathyT(i+1,j) + Mean_SL, 0.0) + max(G%bathT(i,j) + Mean_SL, 0.0)) + GV%H_to_Z*GV%H_subroundoff)` ?  This would at leas avoid a division by zero, but it would give very strong drag everywhere above the mean sea-level.
  This particular block of code occurs at initialization (so the instantaneous local sea level is not known) and is inside of a test for `if (.not.CS%nonlin_stress)`. Perhaps the biggest takeaway from inspecting this line is that the linearized bottom stress is simply incompatible with wetting and drying.
  Because this line is simply reproducing the previous behavior, I think that it need not be addressed in this particular PR, but it should be addressed thoughtfully in a future PR.  If you have such a suggestion to make, @herrwang0, it would be very warmly welcomed!

herrwang0: I don't really have a good change for this. For my own purpose, I just needed to add a `H_subroundoff` to make sure the denominator is not zero so that the model does not complain. I am doing tide runs and there is no surface mechanical forcing, so this part of `BT_force` is zero anyway.

I agree this is definitely not a good choice for general purposes. I guess it does make sense to use `nonlin_stress` for wetting and drying. Otherwise, maybe simply set `IDat[uv]` to zero at land points? This would be consistent with `Dat[uv]` in subroutine `find_face_areas` when there is no `eta` and `add_max` as input, and the term is calculated by harmonic means. The zero SSH approximation here does seem to imply no wetting over the land.

marshallward (approved):

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13555 :heavy_check_mark: 

This did eventually pass, despite many snags and reported failures during testing.  I believe they were all related to Lustre problems and hopefully do not reflect any problems with this PR or any other recently merged PRs, such as the rotational `read_data` support.  But let's keep an eye on it.

PR #1492, opened 2021-08-31
Testing: Set MALLOC_PERTURB_ to 1

marshallward:
The MALLOC_PERTURB_ parameter was used to initialize allocated arrays to
unphysical values.  Previously, we had set this to 256 which would
initialize bytes to 0xff, and indirectly set all floats to NaN.

However, MALLOC_PERTURB_ was undocumnted for values greater than 255,
and we were relying on undefined behavior.  In newer versions of glibc,
a value of 256 does nothing and is equivalent to a value of 0.

This patch changes the MALLOC_PERTURB_ value to 1, which sets bytes to
0xfe and tends to initialize them to an unphysically large value.

Unfortunately we have temporarily lost the ability to initialize to NaN,
but for now this is probably the best we can do.

(Note that other compilers like Intel can explicitly initialize
allocatables to NaN, so this update is more specific to the GCC
configuration of our GitHub Actions setup than the test suite.)


marshallward:
I just noticed these are still called "NaN" tests, even though we are no longer doing NaN initialization.  Also it's not a great idea to inject glibc-specific flags into a supposedly platform-agnostic Makefile.

I will think on this for a bit.  If I can't find a better solution then I'll close the PR.

Hallberg-NOAA (approved):
This is a simple enough change, and it is well motivated by the descriptive comments.  Given that this impacts only the TC testing, which has passed, I am going to approve this particular PR without requiring the redundant pipeline testing.

PR #1493, opened 2021-09-01
(*)Avoid uninitialized data use in find_interfaces

Hallberg-NOAA:
Modified find_interfaces to avoid using uninitialized data when there is no
input data in a column or only one layer of data and linear interpolation is
nonsensical.  This could change the initial conditions in some layer-mode cases,
but in practice all solutions in the MOM6-examples test suite are bitwise
identical, even though temporarily inserted error messages reveal that the
conditions that triggered the modified code are being encountered.


marshallward (approved):

marshallward:
https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13518

PR #1495, opened 2021-09-02
 ALE sponge mask_z halo fixes

marshallward:
In the ALE sponge, the `mask_u` and `mask_v` masks are constructed from
`mask_z`, but also require valid halo data on their E/W or N/S
boundaries.

Although a `pass_var` function is called on `mask_z` before computing
these masks, this function will not populate the halos of `mask_z` if
there is no adjacent data, e.g. a non-reentrant boundary or a
land-masked tile.

And even though `mask_z` was initialized to zero, this was undone by the
internal dellocation/reallocation of the array inside of
`horiz_interp_and_extrap_tracer` (although the actual result appears to
be compiler dependent).

There are two major changes in this patch:

* The FMS-based `horiz_interp_and_extrap_tracer` function no longer does
  a deallocate/reallocate of its output arrays, and now simply assumes
  they are unallocated.

  The output arrays are also explicitly declared as intent(out).

  This change clarifies that only the compute domains of `mask_z` and
  associated fields are updated, although it doesn't fully resolve the
  issue described above.

* The ALE sponge code now explicitly initializes the halo values of
  mask_z before interpolating the mask_u and mask_v masks.

  This ensures that `mask_[uv]` boundary values are disabled on
  points where no halo data is available (and hence no halo updates from
  `pass_var`.  When the data is available, sensible values will replace
  these zeros.

These changes prevent anomalous values of mask_z from entering the
halos, and ensuring that `mask_[uv]` contain sensible values.

A similar operation should not be required by the tracer fields, since
the zero-halo values in the mask will correctly disable these values
when no adjacent field is available for halo updates.


marshallward:
I suppose this works for `sp_val_[uv]` because we are using the `mask_[uv] == 1` checks, but if we were to replace this with the probably-faster `mask_[uv] * sp_val_[uv]` then we'd have to do something about `sp_val_[uv]` on those edge halos.  I'm not sure how that would look.

marshallward:
It looks like the `sp_val_[uv]` halo values are an issue if I turn on aggressive initialization.

Even though these values never get used when `mask_[uv]` is zero, the compiler still forbids their assignment with an uninitialized value.

We may want to first build `mask_[uv]` and then, depending on the value, conditionally evaluate `sp_val_[uv]`.

Hallberg-NOAA (approved):
These changes all make sense to me, and are very well explained in the comments describing this PR.

Hallberg-NOAA:
This PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13581.

PR #1497, opened 2021-09-14
Main to dev gfdl

marshallward:


Hallberg-NOAA (approved):
These changes have been thoroughly vetted before they were migrated from dev/ncar to main.

There is an change in one of the TC regressions with this case due to the correction of an understood bug.  Also, this PR has passed pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13588.  However, there are some new runtime parameters, so there will have to be a follow-up PR updating MOM6-examples.

PR #1498, opened 2021-09-16
Fix initialize_ice_thickness_from_file for ISOMIP

Hallberg-NOAA:
Adds a test and warning message to avoid a fatal error when the ISOMIP (or
other ice-shelf) test case ice shelf thicknesses are initialized from a file
that has the ice thickness but not a mask variable, correcting the behavior that
was recently changed back to its previous behavior.  If the mask variable exists
it is read, but if not it is set based on the thicknesses.  All answers are
bitwise identical in test cases that worked before, and the ISOMIP test cases
once again work with the input files that they used previously.


OlgaSergienko:
These changes work for me. Thank you, Bob.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13643 :heavy_check_mark:

PR #1499, opened 2021-09-17
Use allocate with source to initialize arrays

Hallberg-NOAA:
Use the allocate call's source optional argument to initialize 611 arrays in
62 files.  This change simplifies and shortens the code and will make it easier
to identify uninitialized allocated arrays.  This change was only made to arrays
that were already being initialized on the same line as the allocate or shortly
thereafter.  All answers are bitwise identical.


marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13646 :heavy_check_mark:

marshallward (commented):
I'll approve any PR that removes semicolons!

marshallward (approved):

PR #1500, opened 2021-09-22
Fix to sponge verbosity, also some documentation.

kshedstrom:
In a chat with @Hallberg-NOAA and @adcroft, we decided to clean up some excessive print statements from MOM6's sponge code. Here it is. Also, some added verbosity to a warning about negative water depths. Thirdly, the last of the internal vertical diffusion documentation (for now).


src/framework/MOM_horizontal_regridding.F90
```diff
@@ -764,7 +768,7 @@ subroutine horiz_interp_and_extrap_tracer_fms_id(fms_id,  Time, conversion, G, t
 
   if (.not.spongeDataOngrid) then
     if (is_root_pe()) &
-      call time_interp_external(fms_id, Time, data_in, verbose=.true., turns=turns)
+      call time_interp_external(fms_id, Time, data_in, verbose=(verbosity>2), turns=turns)
```
Hallberg-NOAA, line 771: I like and agree with this change, but I would prefer to set the threshold for getting all this output to a somewhat verbose setting, say `verbose=(verbosity>5)`.

Hallberg-NOAA (approved):
I agree with all the code changes in this PR, and I have read through the documentation changes, and believe them to be substantially correct as well.  This PR has passed the TC testing, and the pipeline testing has been triggered at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13648 .  This PR should be accepted after the pipeline testing has passed.

PR #1501, opened 2021-09-23
+(*)Add ALTERNATE_FIRST_DIRECTION

Hallberg-NOAA:
Added a new runtime parameter, ALTERNATE_FIRST_DIRECTION, to cause the first
direction in the directionally split operators (such as the continuity solver,
the barotropic solver, and the tracer advection) to alternate with every dynamic
timestep.  This includes changes to record this direction in the restart file,
so that the model will reproduce across restarts.  In addition a previously
uninitialized logical in offline_advection_layer is now being initialized
similarly to other similar variables in the MOM_offline_main module.  By
default, all answers in the MOM6-examples are bitwise identical, but there is a
new runtime parameter and a new non-mandatory field in the MOM6 restart files,
and there might be changes in offline tracer advection runs.


src/core/MOM.F90
```diff
@@ -2435,6 +2462,12 @@ subroutine initialize_MOM(Time, Time_init, param_file, dirs, CS, restart_CSp, &
       CS%tv%S => CS%S
     endif
 
+    ! Reset the first direction if it was found in a restart file.
+    if (CS%first_dir_restart > -0.5) &
```
marshallward, line 2466: Is the value of `-0.5` significant here?  Or is it just something greater than `-1.0`?  A comment might be a good idea here (although not necessary).

Hallberg-NOAA: The value was set to -1.0 before it was read from the restart file, so the comment above does accurately explain what the code is doing here, even though it does not explain why I used -0.5 instead of -1.0.  The value of -0.5 here is to avoid a comparison with exactly the same value as it was set to, but it probably could be `> -1.0` or it could be `>= 0.0` instead.  If it were an integer instead of a real value, it would have to become one of these other expressions.

marshallward: One of the inequalities might be more clear, but it's not necessary; I'll merge this.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/jobs/69534 :heavy_check_mark: :orange_circle: 

New parameter:
* `ALTERNATE_FIRST_DIRECTION`

marshallward (approved):

PR #1502, opened 2021-09-29
Testing: Generalized testing rules 

marshallward:
This PR adds new rules and features to the MOM6 testing Makefile:

* New hooks are provided to control the list of builds, configurations, and active tests.
  * `BUILDS` for executables
  * `CONFIGS` for configurations (tc1, tc2, etc)
  * `TESTS` for test types (`grid`, `rotate`, etc)
  * Tests no longer need to begin with `tc` and can be added to `CONFIGS`, although this is still the default behavior.
* Rules for new groupings of tests have also been provided:
  * `test.dim` can now be used to run all dimensional tests
  * Individual configurations can now be tested, e.g.:
    * `tc2` will run all tc2 tests
    * `tc2.grid` will run the symmetric/asymmetric grid tests on tc2
    * `tc2.dim` will run all dimension tests on tc2
    * `tc2.dim.q` will run the Q-dimensional tests on tc2
    NOTE: `tc2`will run diagnostic tests, but the others will not; use `tc2.grid.diag` to run them, for example.
* Many of the old "plural" groups have been renamed, matching the single-text format; e.g. `test.rotations` is now `test.rotate`.
* Many of the wildcard (`%`) rules have been redefined as explicitly over the tests.
* The `repro` build was incorrectly being built, even when `DO_REPRO_TESTS` was unset.  This has been fixed.
* `make clean` now deletes both `build` and `deps` (including FMS), thereby reversing the behavior of `Makefile`.
  * A new `make.build` rule is provided to only delete the MOM6 builds

The GitHub Actions tests have also been updated to use the renamed

### DIFF
--- a/.github/workflows/expression.yml
+++ b/.github/workflows/expression.yml
@@ -24,4 +24,4 @@ jobs:
       run: make run.symmetric -k -s
 
     - name: Run tests
-      run: make test.repros test.dims -k -s
+      run: make test.repro test.dim -k -s
--- a/.github/workflows/other.yml
+++ b/.github/workflows/other.yml
@@ -24,4 +24,4 @@ jobs:
       run: make run.symmetric -k -s
 
     - name: Run tests
-      run: make test.openmps test.nans test.restarts -k -s
+      run: make test.openmp test.nan test.restart -k -s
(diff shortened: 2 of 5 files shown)


adcroft (approved):
- `make clean.build` was a good compromise to our debate about targets for cleaning
- alignment of target names (removing plurals) is much nicer
- removing "repro" from BUILDS makes interactive testing quicker

PR #1503, opened 2021-09-29
User/wfc/remap scheme

wfcooke:
Update to allow the remapping scheme to be changed.
Also had to update the call to MOM_grid_init and comment out some timers.


marshallward:
DId you mean to comment out the clocks?

Also, this PR probably needs a squash or rebase.

marshallward:
Sorry @wfcooke, the suggestion to squash was meant for whoever handles the merge (e.g. Bob or Alistair, if not me).

You are of course welcome to restructure the commits though.

wfcooke:
@marshallward Yes, I needed to comment out the clocks as they were throwing errors otherwise.

Hallberg-NOAA (commented):
It would appear that these changes might make it impossible to recover the previous settings. In particular, the default for REMAPPING_SCHEME is "PPM_H4", consistent with its use elsewhere in the MOM6 code, but the hard coded value on line 303 was previously "PLM".

I do not see any way to recover the previous default settings (PLM inside of init_oda, and PPM_H4 elsewhere) with this code.  So this change will change answers for anyone using init_oda, and it will need to be treated with care.  If we are certain that no one else is using this code (e.g., it would always have led to fatal errors), and if you are good with this irreversible answer change, it might be an acceptable change. If not, we might need to add an other runtime parameter (like "REMAPPING_SCHEME_ODA", potentially taking its default from "REMAPPING_SCHEME" or with a default of "PLM" to not change answers by default) to allow the previous combination of settings to be recovered.

MJHarrison-GFDL:
The remapping scheme being set is contained within the DA regridding control structure, so should not impact the contents of the CS used by the model. However, it would make sense to use a different name here, such as "ODA_REMAPPING_SCHEME".

wfcooke:
@Hallberg-NOAA @MJHarrison-GFDL  Good point about the remapping scheme.
Updated the code to apply ODA_REMAPPING_SCHEME instead
Also removed some timers that were causing blowups (probably due to PE-list conflicts)

Hallberg-NOAA:
These changes are moving in the right direction, but there is a new issue here, in that the call to `get_param(param_file, ..., "ODA_REMAPPING_SCHEME", ...)` needs to have a description of the parameter or it will not be recorded in the MOM_parameter_doc files.  It was OK when it did not have a description before, because it was consistent with another call to get_param in another file, and that other file was taking care of documenting the parameter.

  Also, the second argument to the get_param calls here and throughout this file should be the name of the module, which in this case would be "MOM_oda_driver_mod".  "MOM" is the right value for this second parameter only for get_param calls in MOM.F90.

PR #1504, opened 2021-09-29
Particle API

cspencerjones:
This commit contains dummy interfaces for the 2D version of the drifters package. I have not yet added the actual code, but this is based on Matt and Luyu's work. I can compile it without errors. 

FAO @adcroft (also potentially @StephenGriffies and @MJHarrison-GFDL )

Perhaps this should be a draft PR until alterations to the main code are added?

There was some discussion about whether to put particles_run in the dynamics step or in the thermodynamics step. I'm not 100% sure I understand that discussion, since u and v exist both before and after the thermodynamic step (although presumably they are altered by the diabatic processes?) Any clarification would be helpful.


StephenGriffies:
MOM6 has a suite of time steps 
 thermodynamics for T/S/tracers  (T1)
 dynamics for baroclinic (T2, typically T2 < T1)
 dynamics for barotropic (T3 << T2)
 others related to coupling to atmosphere

We have no code for online time averaging the velocity (whereas we do for the transports since they are used for the tracer equation). Since it is the velocity that we need for Lagrangian particles and not transports, we need to update the particles on the velocity (baroclinic, T2) time step.

cspencerjones:
> Since it is the velocity that we need for Lagrangian particles and not transports, then we need to update the particles on the velocity (baroclinic, T2) time step.

Thanks, this helps. I will add the actual code (including particles_run in the baroclinic timestep) and comment again when I have done that.

Hallberg-NOAA:
Another option might be to optionally add the averaging of the velocities to allow the particle code to work on the longer T1 timestep, especially if we think that there may commonly be more particles than grid points.  (I.e., do we expect it to be more work to take and store an average of the velocities or to advect the particles for a step, including the cost of halo updates?)

  However, taking the average velocities and the longer timestep would require the particle code to be able to handle advective CFL numbers that are larger than 1.  From what you know about the particle code, @cspencerjones , is it restricted to work in cases where the advective CFL number (|u| * dt / dx) is less than 1?

adcroft:
Don't the particles need to sample the high frequency motions? The efficiency can be addressed by bringing in a structural change we're making in icebergs, but this won't affect the API. The purpose of @cspencerjones PR is to establish the hooks so the particles can be developed.

@cspencerjones Yes, please add the calls in src/core . You'll need to make `particles_init()` and `particles_save_restart` public, and could you add a `particle_end()`?

cspencerjones:
Having said that, I now think that I have screwed up `part_time` which means that writing particle trajectories doesn't work. Sorry, I should have realised this before pushing - I will fix now.

cspencerjones:
Having said that, I now think that I have screwed up `part_time` which means that writing particle trajectories doesn't work. Sorry, I should have realised this before pushing - I will fix now.

marshallward:
Hi @cspencerjones I've enabled automated testing for you.

You have provided most of the necessary docstrings for new functions and types, but there a few missing.  Could you include those in this PR?  The test log should indicate which ones are missing.

Also, if you are able to convert any (if not all) of the pointers inside of these  to either allocatables or locals, then that would help with some related ongoing work, but this is not absolutely necessary.

cspencerjones:
I suspect the documentation is not completely fixed, but I have done a lot on it. 

> Also, if you are able to convert any (if not all) of the pointers inside of these to either allocatables or locals, then that would help with some related ongoing work, but this is not absolutely necessary.
I can try to do a little bit of that in the morning, but I think that it won't be easy for some of these.

cspencerjones:
Checks have passed, and I have done a lot of what was suggested. It would take a lot of effort to convert the remaining pointers to allocatables (the underlying code, which is not part of this PR relies on the flexibility of pointers). If people have more feedback, now is the time. Thanks!

adcroft:
I agree with @cspencerjones : particles (and the iceberg code they are based on) use linked lists and rewriting those structures is not feasible at this stage. The conversion to allocatable for gridded data, implemented above, makes sense.

marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13767

PR #1509, opened 2021-10-06
+*Change defaults for 4 parameters to better values

Hallberg-NOAA:
Updated the defaults of 4 run-time parameters (INTERNAL_WAVE_SPEED_BETTER_EST,
PARALLEL_RESTARTFILES, EPBL_MLD_BISECTION, and BBL_USE_EOS) to more appropriate
values.  In each case, the previous default was simply the older setting, and
not the better recommendation.  It also adds logic determining whether
SIMPLE_TKE_TO_KD does anything, and only log its setting if it does.  These
default changes were discussed by the MOM6 consortium as a whole in June, 2021
and were widely agreed to. In addition this commit removes the old obsoleted
runtime parameter ORIG_MLD_ITERATION, and obsoletes the runtime parameter
LARGE_FILE_SUPPORT, and it adds comments describing several real variables and
their units.  Because this changes several default values, it will change
answers unless these parameters are explicitly set in the MOM_input files.
However, because MOM6-examples PR #344 does set these values to their old
defaults where they are used, no answers are changed in the MOM6-examples
regression suite, although there are changes to the MOM_parameter_doc files.

### DIFF
--- a/src/diagnostics/MOM_diagnostics.F90
+++ b/src/diagnostics/MOM_diagnostics.F90
@@ -1644,7 +1644,7 @@ subroutine MOM_diagnostics_init(MIS, ADp, CDp, Time, G, GV, US, param_file, diag
                  units="m s-1", default=0.0, scale=US%m_s_to_L_T)
   call get_param(param_file, mdl, "INTERNAL_WAVE_SPEED_BETTER_EST", better_speed_est, &
                  "If true, use a more robust estimate of the first mode wave speed as the "//&
-                 "starting point for iterations.", default=.false.) !### Change the default.
+                 "starting point for iterations.", default=.true.)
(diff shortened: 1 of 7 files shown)


Hallberg-NOAA:
PARALLEL_RESTARTFILES has been used extensively in production runs, and has been demonstrated to reproduce.  This may be something to do with the testing, and not with the restart capability itself.
I personally do not care about changing the default for PARALLEL_RESTARTFILES, as it is one that should give identical results with either setting, whereas I do care a great deal about correcting the other 3 defaults, which actually do matter. At this point, I am inclined to revert the default for PARALLEL_RESTARTFILES or explicitly set it to FALSE in the MOM_input files for the TC tests to clear the way to accept the fixes to the other three defaults.

PR #1515, opened 2021-10-11
fix low mode in tidal_mixing

raphaeldussin:
this PR updates some function prototypes to pass the necessary structures for the get_lowmode function to work.
The teleport function is commented out since its behavior is inadequate and might not be necessary anymore.


marshallward:
Hi @raphaeldussin , given the increased PR traffic, would it be possible to move these draft PRs to another place, such as your own repo or the Discussions tab, until they are ready for review?

PR #1523, opened 2021-10-15
+Make 37 optional arguments in src/core mandatory

Hallberg-NOAA:
Made 37 optional arguments that are always present in calls into non-optional
arguments to routines in the src/core directory.  Many of these are pointer
arguments related to things like open boundary conditions, so if these types are
not to be used, they are simply not allocated.  In several cases, this required
the order of the arguments to be shifted around, but the various types of the
arguments should prevent the model from compiling if the calls (e.g., in
user-modified code that is not in the github repository) are not changed
equivalently.  Also eliminated 3 internal arguments in MOM_barotropic.F90 that
are always hard-coded to the same values (the maximize argument to
BT_cont_to_face_areas()) or are never used (the guess arguments to uhbt_to_ubt()
and vhbt_to_vbt()), along with the code that they would exercise.  All answers
and output are bitwise identical.

### DIFF
--- a/config_src/drivers/solo_driver/MOM_driver.F90
+++ b/config_src/drivers/solo_driver/MOM_driver.F90
@@ -96,8 +96,6 @@ program MOM_main
   ! This is .true. if incremental restart files may be saved.
   logical :: permit_incr_restart = .true.
 
-  integer :: ns
-
   ! nmax is the number of iterations after which to stop so that the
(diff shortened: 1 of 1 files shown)


marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13842 :heavy_check_mark: 

(Took about 5 attempts, but it eventually passed...)

PR #1524, opened 2021-10-17
+Optional just_read_params args are now mandatory

Hallberg-NOAA:
Made 44 previously optional just_read_params arguments into mandatory
arguments and renamed them just_read to avoid duplicated variables.  This change
should be minimally disruptive, as these optional arguments were always being
provided.  Some argument documentation blocks were also reformatted or the
comments in them were corrected.  All answers are bitwise identical, but some
arguments have changed.


marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/13857 :heavy_check_mark:

PR #1534, opened 2021-10-25
Refresh attempt to get rid of NetCDF calls

MJHarrison-GFDL:


marshallward:
This is now passing for me with FMS2 but I still get the `MPI_Bcast` error with FMS1.

We may be using FMS2 operations - or perhaps just following FMS2 access patterns - in the FMS1 implementation which touch MPI resources which it ought not touch.  I'd suggest we look in `MOM_io.F90` and the associated `infra/FMS1` calls.

marshallward:
This is now passing for me with FMS2 but I still get the `MPI_Bcast` error with FMS1.

We may be using FMS2 operations - or perhaps just following FMS2 access patterns - in the FMS1 implementation which touch MPI resources which it ought not touch.  I'd suggest we look in `MOM_io.F90` and the associated `infra/FMS1` calls.

marshallward:
I don't understand what is causing the error, but I think that I have at least identified a problem.

If I comment out the `MOM_read_data()` call in [L456](https://github.com/NOAA-GFDL/MOM6/blob/221df78c6d979c33b912154f44fc44e6fc365093/src/framework/MOM_horizontal_regridding.F90#L456), then the test does not crash and reaches completion.

It is using the `MOM_read_data_2d_region()` function, which is for reading domain-decomposed 2d fields.  Since we are reading a single vertical level of a 4d field, I wonder if this is the wrong function?

marshallward:
@MJHarrison-GFDL Can you resubmit this to NOAA-GFDL/MOM6?

Also, after talking with the others, we agreed that the best short-term solution is to extend the generic `read_variables` function to support higher dimensions.

PR #1537, opened 2021-10-27
+Clarify the use of grid types

Hallberg-NOAA:
This PR simplifies the use of the dynamic and ocean-model specific grid types,
moving the routine to rotate a dynamic grid into the MOM_dyn_horgrid module
where it belongs and where it can be shared with SIS2, and reducing the
unnecessary use of elements of the grid type in initialize_MOM and the space in
code where the two grids coexist.  It also correct compile-time inconsistencies
in the use of grid types between  unit_drivers/MOM_sum_driver.F90 and the rest
of the MOM6 code, so that it once again compiles and works.  All answers that
worked before are bitwise identical with this change, although there is some
reordering of one entry in the MOM_parameter_doc files.  The commits in this PR
include:

- NOAA-GFDL/MOM6@2629a504f (*)Fix compile-time issues with MOM_sum_driver.F90
- NOAA-GFDL/MOM6@d21e65d7a +Reduce use of dyn_horgrid in initialize_MOM
- NOAA-GFDL/MOM6@4cb6948e8 +Move rotate_dyn_horgrid to MOM_dyn_horgrid module

### DIFF
--- a/config_src/drivers/unit_drivers/MOM_sum_driver.F90
+++ b/config_src/drivers/unit_drivers/MOM_sum_driver.F90
@@ -18,13 +18,14 @@ program MOM_main
   use MOM_coms, only : EFP_type, operator(+), operator(-), assignment(=), EFP_to_real, real_to_EFP
   use MOM_cpu_clock, only : cpu_clock_id, cpu_clock_begin, cpu_clock_end
   use MOM_cpu_clock, only : CLOCK_COMPONENT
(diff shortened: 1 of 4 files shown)


marshallward:
Gaea regression: https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/14068 :heavy_check_mark: :yellow_circle: 

Confirmed reordering of `REFERENCE_HEIGHT` parameter.

marshallward (approved):

PR #1539, opened 2021-11-03
Add weighted d[uv]_dt_str diagnostics

NoraLoose:
This adds four new diagnostics building on the wind stress acceleration diagnostics `du_dt_str`, `dv_dt_str` (from #1437)
1) their thickness-weighted versions: `h_du_dt_str`, `h_dv_dt_str` (completing the set of diags from #1398)
2) their viscous remnant fraction: `du_dt_str_visc_rem`,  `dv_dt_str_visc_rem` (completing the set of diags from ocean-eddy-cpt#10)

I did some quick tests with the CPT NeverWorld2 setup, which confirm that **online and offline multiplication** by 1) `h` 2) `visc_rem_[uv]` coincide (up to 1) interpolation error and 2) numerical noise, respectively). See the two figures below.

### 1. Thickness-weighted
test_h_du_dt_str

### 2. Viscous remnant fraction
test_du_dt_str_visc_rem

### DIFF
--- a/src/parameterizations/vertical/MOM_vert_friction.F90
+++ b/src/parameterizations/vertical/MOM_vert_friction.F90
@@ -1957,7 +1957,7 @@ subroutine vertvisc_init(MIS, Time, G, GV, US, param_file, diag, ADp, dirs, &
     call safe_alloc_ptr(ADp%dv_dt_visc,isd,ied,JsdB,JedB,nz)
     call safe_alloc_ptr(ADp%diag_hv,isd,ied,JsdB,JedB,nz)
   endif
- 
+
   CS%id_h_du_dt_str = register_diag_field('ocean_model', 'h_du_dt_str', diag%axesCuL, Time, &
       'Thickness Multiplied Zonal Acceleration from Surface Wind Stresses', 'm2 s-2', &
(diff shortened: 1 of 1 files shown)


Hallberg-NOAA:
These newly added diagnostics look great, thank you.

However, this is failing one of the automated checks, in a very simple way.  Please remove the trailing spaces that are causing the automated Doxygen and style check to fail, as indicated at https://github.com/NOAA-GFDL/MOM6/runs/4097277039?check_suite_focus=true.

NoraLoose:
Thanks @Hallberg-NOAA! I removed the trailing spaces, and the check is now passing.

Hallberg-NOAA (approved):
These are useful new diagnostics, and this PR has passed the TC testing and it has conditionally passed the pipeline testing at https://gitlab.gfdl.noaa.gov/ogrp/MOM6/-/pipelines/14014, although it does add new diagnostics so the available_diags files will have to be updated.

For the code itself, we should probably modify this PR to avoid doing extra allocates and deallocates of the temporary arrays for the diagnostics, but that can be addressed by adding the capability to write a diagnostic that is the product of two arrays, at which point all of the extra code surrounding the post_data calls in this PR can be replaced with a single-line call to that new routine.

PR #1541, opened 2021-11-08
Adds option to homogenize forces and fluxes fields

breichl:
- Adds functions to do global averages on U and V grids in MOM_spatial_means
- Adds functionality to average all forcing and fluxes fields in MOM_forcing_types
- Adds flag to average all forcing and fluxes in MOM.F90
- This new feature is primarily for running in single column like configurations with the coupler, which requires perfectly equal forcing across all cells.

### DIFF
--- a/src/core/MOM_forcing_type.F90
+++ b/src/core/MOM_forcing_type.F90
@@ -184,7 +184,6 @@ type, public :: forcing
                                   !! average of the gustless wind stress.
   real :: C_p                !< heat capacity of seawater [Q degC-1 ~> J kg-1 degC-1].
                              !! C_p is is the same value as in thermovar_ptrs_type.
-  real :: gust_const         !< Constant unresolved background gustiness for ustar [R L Z T-1 ~> Pa]
 
   ! CFC-related arrays needed in the MOM_CFC_cap module
   real, pointer, dimension(:,:) :: &
@@ -3532,7 +3531,7 @@ subroutine homogenize_mech_forcing(forces, G, US, Rho0, UpdateUstar)
       do j=js,je ; do i=is,ie
         if (G%mask2dT(i,j) > 0.) forces%ustar(i,j) = sqrt(sqrt(tx_mean**2 + ty_mean**2)*iRho0)
       enddo ; enddo
-   else
+    else
       call homogenize_field_t(forces%ustar, G)
     endif
   else
--- a/src/core/MOM_forcing_type.F90
+++ b/src/core/MOM_forcing_type.F90
@@ -17,6 +17,7 @@ use MOM_file_parser,   only : get_param, log_param, log_version, param_file_type
 use MOM_grid,          only : ocean_grid_type
 use MOM_opacity,       only : sumSWoverBands, optics_type, extract_optics_slice, optics_nbands
 use MOM_spatial_means, only : global_area_integral, global_area_mean
+use MOM_spatial_means, only : global_area_mean_u, global_area_mean_v
 use MOM_unit_scaling,  only : unit_scale_type
 use MOM_variables,     only : surface, thermo_var_ptrs
 use MOM_verticalGrid,  only : verticalGrid_type
@@ -35,6 +36,7 @@ public copy_common_forcing_fields, allocate_mech_forcing, deallocate_mech_forcin
 public set_derived_forcing_fields, copy_back_forcing_fields
 public set_net_mass_forcing, get_net_mass_forcing
 public rotate_forcing, rotate_mech_forcing
+public homogenize_forcing, homogenize_mech_forcing
 
 !> Allocate the fields of a (flux) forcing type, based on either a set of input
 !! flags for each group of fields, or a pre-allocated reference forcing.
@@ -3385,6 +3387,7 @@ subroutine rotate_forcing(fluxes_in, fluxes, turns)
   if (do_iceberg) then
     call rotate_array(fluxes_in%ustar_berg, turns, fluxes%ustar_berg)
     call rotate_array(fluxes_in%area_berg, turns, fluxes%area_berg)
+    !BGR: pretty sure the following line isn't supposed to be here.
     call rotate_array(fluxes_in%iceshelf_melt, turns, fluxes%iceshelf_melt)
   endif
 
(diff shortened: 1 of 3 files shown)


breichl:
This code is meant to replace #1480  @Hallberg-NOAA let me know what you think of these changes.  I still have testing and some functionality to complete, but I wanted to get your opinion on these changes compared to the original PR.

Hallberg-NOAA:
I think that this PR is a great improvement over its predecessor, and I like the general direction it is going.  There are a few comments or issues that I see with this.

1) The new routines `global_area_mean_u()` and `global_area_mean_v()` will not reproduce across PE layouts when used in  symmetric memory mode, and they will not reproduce across rotation by 180 degrees when used in non-symmetric memory mode.  The basic problem is that they will double-count the velocity points at the edges of the PEs, or only count the ones on the one edge.  In addition, it is using the integrated area of the unmasked tracer points, not the velocity points. For these routines, I think that the solution may be to do an appropriate interpolation of the velocities or stresses to the tracer points (perhaps with land masking and area-weighted averaging), and then do the global average on the tracer points.

2) What is wrong with taking the area-weighted average of ustar at the tracer points directly?  Yes, it will be larger than the ustar that would be based on the area-averaged stresses, but in the same way that the sub-gridscale gustiness is something that we regularly include when calculating ustar in the first place, this difference is the legitimate result of the averaging operator not commuting with the nonlinear conversion of the velocities into ustar.  As long as we are explicit about how the averaged ustar is calculated, why do we care if it can not be directly derived from the averaged stresses?  If we do care about ustar being exactly the average of taux and tauy, that would probably be handled best in a separate subroutine to copy the information from a `mech_forcing` type to a `forcing` type, with the call controlled by another run-time flag to determine whether to calculate ustar that way.

3) There is a lot of repetition of nearly identical blocks of about 6 lines to take the averages in the new `homogenize_forcing()` and `homogenize_mech_forcing()` routines.  (There are 53 such blocks for tracer points by my count.)  The code might be a lot shorter (by perhaps 300 lines) and less error prone if a small local subroutine were introduced to perform this task of replacing a field with its average in a single-line call.

src/core/MOM_forcing_type.F90
```diff
@@ -3490,6 +3494,233 @@ subroutine rotate_mech_forcing(forces_in, turns, forces)
...
+    enddo ; enddo
+    ty_mean = global_area_mean_v(forces%tauy, G)
+    do j=jsB,jeB ; do i=is,ie
+      if (G%mask2dCv(i,J) > 0.) forces%tauy(i,J) = ty_mean
+    enddo ; enddo
+    if (tau2ustar) then
+      do j=js,je ; do i=is,ie
+        if (G%mask2dT(i,j) > 0.) forces%ustar(i,j) = sqrt(sqrt(tx_mean**2 + ty_mean**2)*iRho0)
+      enddo ; enddo
+   else
```
breichl, line 3535: Formatting needs fixed.

src/core/MOM_forcing_type.F90
```diff
@@ -3385,6 +3388,7 @@ subroutine rotate_forcing(fluxes_in, fluxes, turns)
   if (do_iceberg) then
     call rotate_array(fluxes_in%ustar_berg, turns, fluxes%ustar_berg)
     call rotate_array(fluxes_in%area_berg, turns, fluxes%area_berg)
+    !BGR: pretty sure the following line isn't supposed to be here.
```
breichl, line 3391: I noticed this is a repeat of code above.  I think it is unintentional.

src/core/MOM_forcing_type.F90
```diff
@@ -182,6 +184,7 @@ module MOM_forcing_type
                                   !! average of the gustless wind stress.
   real :: C_p                !< heat capacity of seawater [Q degC-1 ~> J kg-1 degC-1].
                              !! C_p is is the same value as in thermovar_ptrs_type.
+  real :: gust_const         !< Constant unresolved background gustiness for ustar [R L Z T-1 ~> Pa]
```
breichl, line 187: This is left over code and needs to be removed.

src/core/MOM_forcing_type.F90
```diff
@@ -3490,6 +3494,233 @@ subroutine rotate_mech_forcing(forces_in, turns, forces)
...
+  if (associated(fluxes%buoy)) &
+    call homogenize_field_t(fluxes%buoy, G)
+
+  if (associated(fluxes%TKE_tidal)) &
+    call homogenize_field_t(fluxes%TKE_tidal, G)
+
+  if (associated(fluxes%ustar_tidal)) &
+    call homogenize_field_t(fluxes%ustar_tidal, G)
+
+  ! TODO: tracer flux homogenization
```
breichl, line 3672: I don't know if this is worth addressing before taking this PR?

src/diagnostics/MOM_spatial_means.F90
```diff
@@ -47,6 +47,52 @@ function global_area_mean(var, G, scale)
...
+  real, dimension(SZI_(G), SZJ_(G))              :: tmpForSumming
+  real :: global_area_mean_v
+  integer :: i, j, is, ie, js, je, isB, ieB, jsB, jeB
+
+  is = G%isc ; ie = G%iec ; js = G%jsc ; je = G%jec
+  isB = G%iscB ; ieB = G%iecB ; jsB = G%jscB ; jeB = G%jecB
+
+  tmpForSumming(:,:) = 0.
+  do J=js,je ; do i=is,ie
+     tmpForSumming(i,j) = G%areaT(i,j)/max(1.e-20,&
```
breichl, line 64: Fix formatting.

breichl (commented):
I caught a couple formatting issues here and wanted to raise two other issues here.

breichl:
> I think that this PR is a great improvement over its predecessor, and I like the general direction it is going. There are a few comments or issues that I see with this.
> 
>     1. The new routines `global_area_mean_u()` and `global_area_mean_v()` will not reproduce across PE layouts when used in  symmetric memory mode, and they will not reproduce across rotation by 180 degrees when used in non-symmetric memory mode.  The basic problem is that they will double-count the velocity points at the edges of the PEs, or only count the ones on the one edge.  In addition, it is using the integrated area of the unmasked tracer points, not the velocity points. For these routines, I think that the solution may be to do an appropriate interpolation of the velocities or stresses to the tracer points (perhaps with land masking and area-weighted averaging), and then do the global average on the tracer points.
> 
>     2. What is wrong with taking the area-weighted average of ustar at the tracer points directly?  Yes, it will be larger than the ustar that would be based on the area-averaged stresses, but in the same way that the sub-gridscale gustiness is something that we regularly include when calculating ustar in the first place, this difference is the legitimate result of the averaging operator not commuting with the nonlinear conversion of the velocities into ustar.  As long as we are explicit about how the averaged ustar is calculated, why do we care if it can not be directly derived from the averaged stresses?  If we do care about ustar being exactly the average of taux and tauy, that would probably be handled best in a separate subroutine to copy the information from a `mech_forcing` type to a `forcing` type, with the call controlled by another run-time flag to determine whether to calculate ustar that way.
> 
>     3. There is a lot of repetition of nearly identical blocks of about 6 lines to take the averages in the new `homogenize_forcing()` and `homogenize_mech_forcing()` routines.  (There are 53 such blocks for tracer points by my count.)  The code might be a lot shorter (by perhaps 300 lines) and less error prone if a small local subroutine were introduced to perform this task of replacing a field with its average in a single-line call.

Thanks for the review and suggestions, @Hallberg-NOAA.

I think I've fixed (1) based on our discussion offline.  It now interpolates u/v point variables to tracer points and does the averaging there.  I don't have a test for this yet.

I've added options for (2) to either compute ustar from the mean tau or as the mean of ustar.  We cannot include gustiness at this point; it could be added but would require either a new get_param or passing it to the forces module via each cap.  Since it is not clear how to do it and not clear that we need it I've put that decision off for now.

For (3) I've added three new subroutines to do the averaging now, which indeed reduces the file lengths.

src/diagnostics/MOM_spatial_means.F90
Hallberg-NOAA: In this expression, I think that we would prefer to have division at the end, or else put parentheses around the area divided by the masks, so that someone reading the code does not need to remember the order of precedence in Fortran between multiplication and division (they are the same), or whether the Fortran order of operations goes left to right or right to left.  The same comment would apply in global_area_mean_u().

breichl: The most recent code push should address this.

breichl:
@nikizadehgfdl can you try this at some point in your workflow?  It is a little different than the previous attempt, but I believe will have the same functionality.  The replacement parameter to activate homogenized forcing with this code will be "HOMOGENIZE_FORCINGS = True"

Issue #1527, opened 2021-10-19
MOM6 (with ocean only benchmark)  gets stuck with higher rank count per node


puneet336:
Hi,
I am trying to run the MOM6 with ocean only benchmark, the simulation seems to be getting stuck.
Compiler - intel2019u5
state of the directory  where simulation got stuck-
```
ocean_only_benchmark1]$ ls test/
available_diags.000000  input.nml           MOM6              MOM_parameter_doc.all        ocean.stats      time_stamp.out
change.sh               input.nml.bak       MOM_input         MOM_parameter_doc.debugging  ocean.stats.nc   Vertical_coordinate.nc
CPU_stats               input.nml.bk        MOM_memory.h      MOM_parameter_doc.layout     RESTART
diag_table              libnuma.so          MOM_override      MOM_parameter_doc.short      run.sh
env.src                 logfile.000000.out  MOM_override.bak  mon                          run.sh.bak
GOLD_IC.nc              lstopo.info         MOM_override.bk   ocean_geometry.nc            slurm-86603.out

```
contents of MOM_override file - 
```
ocean_only_benchmark1]$ cat test/MOM_override
! Blank file in which we can put "overrides" for parameters
#override NIGLOBAL = 720
#override NJGLOBAL = 360

```
contents if input.nml - 
```
ocean_only_benchmark1]$ cat test/input.nml
 &MOM_input_nml
         output_directory = './',
         input_filename = 'n'
         restart_input_dir = 'INPUT/',
         restart_output_dir = 'RESTART/',
         parameter_filename = 'MOM_input',
                              'MOM_override' /

 &diag_manager_nml
 /

 &fms_nml
        clock_grain='ROUTINE'
        clock_flags='SYNC'
        domains_stack_size = 955296
        stack_size =0 /

 &ocean_solo_nml
        months = 0
        days = 20 /
```

[slurm.out.txt](https://github.com/NOAA-GFDL/MOM6/files/7375437/slurm.out.txt)

with 56 ranks (i have 56 cores per node), the simulation seems to be getting stuck at this state (stdout) - 
```
MOM Date      1/01/01 00:00:00      0: En 5.483091E-01, MaxCFL  0.00000, Mass 7.909719100499E+19, Salt  35.00000000000, Temp   5.06383782258
    Total Energy: 4402CF01460DFD5A  4.3369709224187347E+19
[...]

PR #1542, opened 2021-11-10
Disable clock sync of in-loop BML timers

marshallward:
This patch overrides the default clock sync (configured in FMS) for the
CPU clocks inside of the bulk mixed layer's loops.

This prevents model hangs due to an inconsistent number of clock syncs
when different ranks have different size domains.


Hallberg-NOAA (approved):
This PR adroitly alleviates the anomalies addressed in Issue #1527, which can be annulled after this advance is accepted.  A-1 accomplishment, @marshallward.

marshallward:
Amazing alliteration, @Hallberg-NOAA 🤣

PR #1561, opened 2022-03-21
Removed but in DOME2D temperature initialization

hdrake:
This fixes the comparison between the four coordinates: z, sigma, rho, layer.
Screen Shot 2022-03-21 at 3 54 59 PM


hdrake:
https://github.com/mom-ocean/MOM6/issues/1560

PR #1562, opened 2022-03-25
Add missing h_neglect

NoraLoose:
Looks like the addition of `h_neglect` is missing in one line for the computation of zonal isopycnal slopes.

Note that is is correctly added in the meridional counterpart:
https://github.com/mom-ocean/MOM6/blob/399a7db1565e89e0ac8aeedd2de32bc846b1f230/src/core/MOM_isopycnal_slopes.F90#L363-L370


marshallward:
Hi Nora, could you re-send this to the NOAA-GFDL/MOM6 repository for review?  Thanks

https://github.com/noaa-gfdl/mom6

Issue #98, opened 2015-03-30
Add diagnostic: thkcello


jkrasting:
Add the 'thkcello' diagnostic.  Note that the vertical remapping process will complicate this as thkcello will become time-dependent.

For now, this diagnostic should be based on 'h' and follow the convention:

cmor_field_name='thkcello', cmor_long_name='Ocean Model Cell Thickness', cmor_units='m', cmor_standard_name='cell_thickness'

Issue #97, opened 2015-03-30
Check all existing units attributes for CMOR compliance


jkrasting:
All MOM6 diagnostics need to be reviewed so that the units are CMOR compliant.  Special attention needs to be paid to division and exponent operators.

example:  W/m**2 should be W m-2

Issue #96, opened 2015-03-30
Add diagnostic:  masso


jkrasting:
Add the 'masso' diagnostic, where:

   masso = the global sum of masscello (mass_wt)

Issue #95, opened 2015-03-30
Add diagnostic: pbo


jkrasting:
Add the 'pbo' diagnostic, where:

   pbo = col_mass + pso

PR #1563, opened 2022-04-04
GFDL to main candidate branch (2022-04-04)

marshallward:
This patch includes the usual mix of new features, bugfixes, and code refactors.

New features include internal tides updates (both solver and new diagnostics), Laplacian diffusion of interface heights, Hybgen/HYCOM compatibility, and ice shelf dynamics.

`MOM_input` now supports longer line lengths, primarily to support complex OBC specifications.

Several minor bugfixes are also included in this PR, and are detailed below.

New parameters (based on regression suite; not necessarily complete):
* `THICKNESS_TOLERANCE`    
* `KH_ETA_CONST`
* `KH_ETA_VEL_SCALE`
* `ITD_LANDFAST`
* `VELOCITY_REMAPPING_SCHEME`
* `PARTIAL_CELL_VELOCITY_REMAP`
* `REMAP_VEL_MASK_BBL_THICK`

Features:
- NOAA-GFDL/MOM6#95: (@raphaeldussin)
- NOAA-GFDL/MOM6#89: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#83: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#81: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#80: (@OlgaSergienko)
- NOAA-GFDL/MOM6#77: (@Hallberg-NOAA)

Bugfixes and Calculation Improvements:
- NOAA-GFDL/MOM6#98: (@NoraLoose)
- NOAA-GFDL/MOM6#90: (@raphaeldussin)
- NOAA-GFDL/MOM6#88: (@raphaeldussin)
- NOAA-GFDL/MOM6#84: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#82: (@marshallward)
- NOAA-GFDL/MOM6#76: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#69: (@gustavo-marques)
- NOAA-GFDL/MOM6#65: (@gustavo-marques)
- NOAA-GFDL/MOM6#73: (@Hallberg-NOAA)

Refactoring and Testing:
- NOAA-GFDL/MOM6#97: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#96: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#94: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#93: (@marshallward)
- NOAA-GFDL/MOM6#87: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#86: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#85: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#78: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#74: (@marshallward)
- NOAA-GFDL/MOM6#64: (@MJHarrison-GFDL)
- NOAA-GFDL/MOM6#63: (@Hallberg-NOAA)
- NOAA-GFDL/MOM6#49: (@Hallberg-NOAA)

Contributors:
  - @Hallberg-NOAA: 16
  - @raphaeldussin: 3
  - @marshallward: 3
  - @gustavo-marques: 2
  - @NoraLoose: 1
  - @OlgaSergienko: 1
  - @MJHarrison-GFDL: 1


marshallward:
Thanks @jiandewang, that is useful information.  We did fix some netCDF issues on our end, so it's very possible we created a new problem.

Also reassuring it is restricted to the `infra/FMS2` section, and (hopefully) unrelated to the mixed precision changes.

I can see the file.  I don't see anything odd, but noticed it does not have a time axis.  Maybe we are doing something which assumes an unlimited time axis.  I have seen similar errors in the past.

Do you know where this file is read in the IAU model?

jiandewang:
> Thanks @jiandewang, that is useful information. We did fix some netCDF issues on our end, so it's very possible we created a new problem.
> 
> Also reassuring it is restricted to the `infra/FMS2` section, and (hopefully) unrelated to the mixed precision changes.
> 
> I can see the file. I don't see anything odd, but noticed it does not have a time axis. Maybe we are doing something which assumes an unlimited time axis. I have seen similar errors in the past.
> 
> Do you know where this file is read in the IAU model?

yes it is not related to FMS mixed mode
code: https://github.com/NOAA-GFDL/MOM6/blob/dev-gfdl-main-candidate-2022-04-04/src/initialization/MOM_state_initialization.F90#L2220

I also put MOM_input at the same location on GAEA, see line 82

kshedstrom (approved):

klindsay28:
Not a full-fledged review...
I noticed in a quick skim some new unused local and imported variables.
These are not errors per se, but do lead to code that is less tidy than necessary.
I attempted to look more methodically and came up with the following.
I wouldn't be surprised if I missed some.

new unused local variables

src/ALE/MOM_ALE.F90
subroutine ALE_main
ntr

src/framework/MOM_error_handler.F90
subroutine disable_fatal_errors
rc
subroutine enable_fatal_errors
rc

variables newly imported via use, but not used

src/core/MOM_check_scaling.F90
MOM_error
FATAL
WARNING
assert
MOM_get_verbosity

src/framework/MOM_unique_scales.F90
MOM_error
FATAL
WARNING

marshallward:
@klindsay28 Thanks for catching those.  We need to investigate the netCDF issue anyway, so we have an opportunity to also clean these up.

jiandewang:
> > Thanks @jiandewang, that is useful information. We did fix some netCDF issues on our end, so it's very possible we created a new problem.
> > Also reassuring it is restricted to the `infra/FMS2` section, and (hopefully) unrelated to the mixed precision changes.
> > I can see the file. I don't see anything odd, but noticed it does not have a time axis. Maybe we are doing something which assumes an unlimited time axis. I have seen similar errors in the past.
> > Do you know where this file is read in the IAU model?
> 
> yes it is not related to FMS mixed mode code: https://github.com/NOAA-GFDL/MOM6/blob/dev-gfdl-main-candidate-2022-04-04/src/initialization/MOM_state_initialization.F90#L2220
> 
> I also put MOM_input at the same location on GAEA, see line 82

@marshallward it's possible this is the cause:
https://github.com/NOAA-GFDL/MOM6/pull/64/files
removing the added part here made "iau" job retained result. I am testing all other jobs but our machine is saturated at this moment, will provide update to you tomorrow.

jiandewang:
> > > Thanks @jiandewang, that is useful information. We did fix some netCDF issues on our end, so it's very possible we created a new problem.
> > > Also reassuring it is restricted to the `infra/FMS2` section, and (hopefully) unrelated to the mixed precision changes.
> > > I can see the file. I don't see anything odd, but noticed it does not have a time axis. Maybe we are doing something which assumes an unlimited time axis. I have seen similar errors in the past.
> > > Do you know where this file is read in the IAU model?
> > 
> > 
> > yes it is not related to FMS mixed mode code: https://github.com/NOAA-GFDL/MOM6/blob/dev-gfdl-main-candidate-2022-04-04/src/initialization/MOM_state_initialization.F90#L2220
> > I also put MOM_input at the same location on GAEA, see line 82
> 
> @marshallward it's possible this is the cause: https://github.com/NOAA-GFDL/MOM6/pull/64/files removing the added part here made "iau" job retained result. I am testing all other jobs but our machine is saturated at this moment, will provide update to you tomorrow.

update: all 27 jobs passed RT

marshallward:
@jiandewang We've looked over the problem, and agreed that it is due to `field_size`, which calls the code block which you linked.

This function is receiving an array of size four (`siz(4)`) and is reading 3d fields (e.g. `h_fg`).  Because `siz` is one larger, it is inserting a third axis, as if it is inserting a single vertical level.  So something like this:

>  (x, y, k) -> (x, y, 1, k)
  
This feature is needed for other parts of the model (e.g. OBCs) and works if we are reading a 2d+time array, but not a 3d array with no time axis.

We believe the issue is that FMS1 used to check the axes and its metadata to decide if this sort of resize should occur.  But FMS2 no longer provides this feature, so we need to figure it out on the MOM6 side.  Since the axis metadata was not based on CF or any other standard, we were reluctant to reintroduce a similar check in MOM6.

We clearly need a better solution for `field_size`, but it's not yet clear how we can fix this in a way which preserves the old FMS1 behavior.

However, none of this happens if `siz` matches the size of the fields.  So we can fix this in the ODA by passing `siz(:3)` into your `field_size()` calls in `initialize_oda_incupd_file`.  But this also assumes that this file will always contain 3d fields, and will never contain a time axis.

I don't know what the format of these file should be.   Can we assume that `ODA_INCUPD_FILE` always contains 3d (x,y,k) fields?  Or could it ever vary?

jiandewang:
> 

@marshallward Phil is going to chime in for the format of that file.

jiandewang:
> > 
> 
> @marshallward Phil is going to chime in for the format of that file. I think a simple solution is adding t-axis in the fixed file, Phil: what do you think ?

pjpegion:
@marshallward  I created the increment without a time-dimension, but I see that other people have included a time dimension. Would it make it easier if I created the increment file with a time-dimension of size 1?

pjpegion:
@abozec Should also have a say in this

marshallward:
I don't know if the file needs to be changed.  If the current file works, then I think we have some obligation to continue supporting it.

The bigger question to me is what other possible formats the file could take?  If we know that, then we can modify the ODA code to support these other possibilities.

If the answer is "all of the above", then it will be a messier solution, but we still ought be able to solve it.

marshallward:
@pjpegion To answer your specific question: If a time axis had been in the file, then I believe it would have worked.

It looks like we should be prepared to handle both 3d and 3d+time.  I will look into a solution.

marshallward:
@pjpegion @jiandewang Can one of you try this patch?  I don't have any ODA configurations and am not set up to test this one, but I think it will fix your issue.

It uses the `ndims` output argument of `field_size` which (hopefully) reports that the fields are 3d and will revert any modifications to the `siz` array.

We need to address the underlying problem at some stage, and perhaps more generally how we want to handle inputs with different axes, but I think this will address your particular problem.
```
diff --git a/src/initialization/MOM_state_initialization.F90 b/src/initialization/MOM_state_initialization.F90
index 4e8dd4f18..459fc6641 100644
--- a/src/initialization/MOM_state_initialization.F90
+++ b/src/initialization/MOM_state_initialization.F90
@@ -2198,6 +2198,7 @@ subroutine initialize_oda_incupd_file(G, GV, US, use_temperature, tv, h, u, v, p
   integer :: isd, ied, jsd, jed
 
   integer, dimension(4) :: siz
+  integer :: inc_ndims
   integer :: nz_data  ! The size of the sponge source grid
   logical :: oda_inc  ! input files are increments (true) or full fields (false)
   logical :: save_inc ! save increments if using full fields
@@ -2265,7 +2266,16 @@ subroutine initialize_oda_incupd_file(G, GV, US, use_temperature, tv, h, u, v, p
   if (.not.file_exists(filename, G%Domain)) &
     call MOM_error(FATAL, " initialize_oda_incupd: Unable to open "//trim(filename))
 
-  call field_size(filename,h_var,siz,no_domain=.true.)
+  call field_size(filename,h_var,siz,no_domain=.true., ndims=inc_ndims)
+  ! NOTE: if siz is larger than h_var, then field_size assumes that h_var is
+  !   2d+time and pads siz with a single vertical level.  Currently, increment
+  !   fields are either 3d or 3d+time, so this correction is an error.
+  ! Here, we revert this correction and move siz(4) back to siz(3).
+  ! TODO: Remove this if the `field_size` behavior is resolved.
+  if (inc_ndims == 3) then
+    siz(3:4) = [siz(4), 0]
+  endif
+
   if (siz(1) /= G%ieg-G%isg+1 .or. siz(2) /= G%jeg-G%jsg+1) &
          call MOM_error(FATAL,"initialize_oda_incupd_file: Array size mismatch for oda data.")
   nz_data = siz(3)
```

jiandewang:
@marshallward I will test that. Where is your branch ?

marshallward:
Sorry, there was no branch, just that patch.

I have added the patch to this branch: https://github.com/marshallward/MOM6/tree/incsize_patch

jiandewang:
> Sorry, there was no branch, just that patch.
> 
> I have added the patch to this branch: https://github.com/marshallward/MOM6/tree/incsize_patch

got it, will test it and get back to you

jiandewang:
> > Sorry, there was no branch, just that patch.
> > I have added the patch to this branch: https://github.com/marshallward/MOM6/tree/incsize_patch
> 
> got it, will test it and get back to you

@marshallward the patch caused the iau job failed when using infra/FMS1 code.
FATAL from PE     0: fms_io(read_data_3d_new), field h_fg in file INPUT/mom6_increment.nc: field size mismatch 1
job using infra/FMS2 code is still waiting on the queue

abozec:
sorry to be late in the conversation... we had a discussion with @awallcraft  about that time dimension when we were implementing the incup and at that time, we did not see an application that would use a incremental update that was changing in time ... Usually, assimilation frameworks work sequentially with model running then assimilation software creating an increment at a specific time that is implemented once the model runs again ... If you have a way to seamlessly accommodate 3 and 4 dims, that would be fine but if it is too much problem, I would stick with 3 dimensions (x,y,z).. That is my 2-cents.

marshallward:
> @marshallward the patch caused the iau job failed when using infra/FMS1 code. FATAL from PE 0: fms_io(read_data_3d_new), field h_fg in file INPUT/mom6_increment.nc: field size mismatch 1 job using infra/FMS2 code is still waiting on the queue

Right, of course the FMS1 would already be correct.  Sorry about that... hopefuly FMS2 answer is OK.

Let me try to come up with something (and perhaps will try to test myself this time).

marshallward:
@jiandewang I have finished looking into this.  If the `ndims` argument is provided, then the behavior is completely different.

When using FMS2 on a 3d array (nx, ny, nz): 
* `call field_size(filename, var, size)` sets `size` to (nx, ny, 1, nz)
* `call field_size(filename, var, size, ndims=ndims) sets `size` to (nx, ny, nz, 0) and `ndims` to 3.

For FMS1, the behavior is largely the same: (nx, ny, nz, _).  (Sometimes the final value is set to 0, but not relevant here.)

We could "fix" this by adding the `ndims` argument, but I am more concerned that this change to `field_size` is too volatile to keep in the model.  The reshaping of (nx, ny, nz) to (nx, ny, 1, nz) is just too particular.

I will discuss this with the others on Monday, but I expect we will have to revert this change before this PR can be considered.

marshallward:
After some discussion, we've decided that this change to `field_size()` is a bit too volatile to patch over, and we'd rather try to find a more robust fix before merging to `main`, so we're going to withdraw this PR and will re-submit it soon.

Apologies to those who have already reviewed it, but the work is appreciated.

PR #1568, opened 2022-05-16
Fix data read for on-grid interpolation

### LINKED ISSUE #1567: Field size error with on-grid temp-salt initialisation
When using `INIT_LAYERS_FROM_Z_FILE` with `TEMP_SALT_INIT_VERTICAL_REMAP_ONLY`, an error is raised when reading the field:

```
FATAL from PE     0: fms_io(read_data_3d_new), field thetao in file INPUT/ocean_daily_z_t0_theta.nc: field size mismatch 1
         131         179           1         131         179          75
```
It seems that the call to `MOM_read_data` doesn't match the `MOM_read_data_2d_region` interface, because `start` and `count` aren't passed through: https://github.com/mom-ocean/MOM6/blob/bac8031b662560066a950bd3c9849ea306abcc25/src/framework/MOM_horizontal_regridding.F90#L438-L440

angus-g:
The wrong `MOM_read_data` interface was being used: a 2D slice of a 3D field was expected, but the interface for a 2D field was being called.

Closes #1567.


marshallward:
Hi Angus, we're still unsure how to handle PRs to this repo.  Could you resend it to noaa-gfdl?

Sorry for the trouble, I know it's not very intuitive.

PR #1575, opened 2022-06-22
Improved MacOS support

marshallward:
This PR makes several changes to the preprocessors and makedep script to
provide better support for MacOS and BSD-like systems.

GitHub CI support for MacOS has also been included in this pull request.

Autoconf now includes macros to test the following:

* The name of sigsetjmp: `sigsetjmp` (BSD) or `__sigsetjmp` (Linux)

* The size of `jmp_buf` and `sigjmp_buf`.  Also renamed to `SIZEOF_*` to
  align with autoconf macro name conventions.

The Linux defaults are retained in `posix.h`, but autoconf will now
override these values.

The makedep script was also modified to run on a wider range of MacOS
environments.

* The shebang was changed from `/bin/bash` to `/usr/bin/env bash` to
  allow for a user-defined bash implementation.

  This is because the script requires Bash 4.x associative arrays, and most
  Mac environments use 3.x.

* Preprocessing of C files now uses `cc` in place of `cpp`.  This is due
  to errors emerging from certain XCode headers (`Availability.h`).

* `dirname` is now only called for one path at a time, rather than
  multiple paths in a single call.  Support for multiple paths is a GNU
  extension and not POSIX compliant.

* Older versions of GNU make do not support use of -t (touch target) and
  -n (no output) at the same time, causing the target file to be created
  despite `-n` enabled.

  To compensate for this, we sweep all object files of zero size.

Despite these changes, there are still significant hurdles to building
on MacOS:

* Users must install a newer bash (4.x) to use makedep.

* When using mkmf, `-D\"sigsetjmp\"` must be explicitly added to
  `FFLAGS`.

The first problem is not solvable without completely rewriting makedep
(probably in another language).  The second could be addressed by
changing the default tempalates in the `mkmf` repository.


marshallward:
oops

PR #1579, opened 2022-07-29
add ML tracer to ideal age module

sdbachman:
This PR adds a new tracer that keeps track of "mixed layer age", which has been added to the ideal age module. This PR also adds the ability to use the actual BL depth that is diagnosed by the active BL scheme inside the ideal age module (for all ideal age tracers).  To turn on this tracer set "DO_ML_RESIDENCE" to True.  To use the real BL depth, set "USE_REAL_BL_DEPTH" to True.


marshallward:
Hi @sdbachman we handle PRs through the major forks, which is mostly likely NCAR in your case.  Can you resubmit there?

If you're unsure which hub then we can talk it over.

PR #1588, opened 2022-12-08
+(*)Fix numerous issues with MOM_stoch_eos and Stanley parameterizations

Hallberg-NOAA:
This PR corrects several bugs that would prevent cases with STOCH_EOS = True or that set a non-negative value of STANLEY_COEF from running at all, and then when it runs it corrects other bugs that cause it to give incorrect answers, as evident in its inability to reproduce across processors, and to fail to pass dimensional consistency testing.  There was also extensive refactoring of the MOM_stoch_eos.F90 code.  Specifically this PR makes the following changes:

- The new routine stoch_eos_register_restarts was added to register the restart
  field associated with the MOM_stoch_EOS module before the restarts are read
  and before restart_registry_lock is called, which was causing any test case
  with STOCH_EOS=True or a positive value of STANLEY_COEF to have a fatal error.

- Added a missing dimensional rescaling factor in the get_param call for
  KD_SMOOTH in MOM_stoch_eos_init, which would have caused cases exercising the
  MOM_stoch_eos options to fail the dimensional consistency tests.

- MOM_stoch_eos_init was changed from a subroutine into a function that returns
  a logical value indicating whether this routine is used further.  The order of
  the arguments was modified to match the order used in every other init call.

- The new routine post_stoch_eos_diags was added to write diagnostics associated
  with the stoch_eos module.

- The MOM_stoch_eos_CS type was made opaque.

- The unused diag argument was removed from MOM_stoch_eos_run.

- Unit arguments were added to the get_param calls for STANLEY_COEFF and
  STANLEY_A.

- Four arrays in the MOM_stoch_eos_CS type that had been declared to optionally
  use static memory allocation were modified to be simple allocatables, as they
  are not used in the vast majority of MOM6 cases, and there is no reason to
  always assign memory to them.

- The register_restart_field call for "stoch_eos_pattern" was revised to use the
  newer, more direct form rather than working via a vardesc type.

- Return statements were added to several of the MOM_stoch_eos routines in the
  cases where they are not supposed to do anything.

- The comments describing several real variables in the MOM_stoch_eos module
  were added or modified to describe their units using the standard format.

- The module use statements at the start of the MOM_stoch_eos module were
  updated to reflect these changes.

- Added use_stochastic_EOS element to the MOM_control_struct to indicate whether
  the stoch_eos calls are to be used.

- MOM_stoch_eos_init is only called if temperature and salinities are state
  variables, as it makes no sense to call it otherwise.

- Calls to stoch_eos routines were updates to reflect the new interfaces.

- The contents of the MOM_stoch_eos_CS type are no longer used in MOM.F90.

- Corrected the domain extents used in EOS calls that are triggered with
  USE_STANLEY_GM and USE_STANLEY_ISO.  These bugs were accidentally introduced
  when the changes adding the MOM_stoch_eos code to the main branch of MOM6 were
  merged with changes on the dev/gfdl branch.  

- Added tests for cases when USE_STANLEY_GM, USE_STANLEY_PGF,USE_STANLEY_ISO or
  USE_STANLEY_ML are set to true but STANLEY_COEF is negative to reset the
  internal versions of this flag to false with a sensible warning message rather
  than encountering segmentation faults.

- Corrected the diagnostics rho_pgf, rho_stanley_pgf and p_stanley to only be
  calculated if they would be written, give identical output when dimensional
  rescaling is applied, and be documented with the right units.  

- Added a missing dimensional rescaling factor in the get_param call for the
  recently added variable DT_OBC_SEG_UPDATE_OBGC in initialize_MOM, which would
  have caused certain cases to fail the dimensional consistency tests.

- Use dimensionally rescaled units when preparing the interface height fields to
  write to the MOM_IC file, and then use the conversion argument to the
  register_restart_field call to undue this scaling.

Several public interfaces to MOM_stoch_eos routines were altered, and there are changes to multiple MOM_parameter_doc files due to the new units, and due to the fact that stoch_EOS parameters are no longer being logged in cases where they are meaningless.  All answers are bitwise identical in cases that do not use dimensional rescaling or the USE_STANLEY parameters, but answers will change (i.e. are corrected) in some cases that do use dimensional consistency tests or the STOCH_EOS or USE_STANLEY parameterizations.

  This commits in this PR include:

- NOAA-GFDL/MOM6@e62ebb7fe Rescale interface heights for MOM_IC
- NOAA-GFDL/MOM6@10c507295 (*)Correct scaling of DT_OBC_SEG_UPDATE_OBGC
- NOAA-GFDL/MOM6@2ebb731a1 (*)Test for inconsistent USE_STANLEY flags
- NOAA-GFDL/MOM6@194077c08 (*)Correct scaling of USE_STANLEY_PGF diagnostics
- NOAA-GFDL/MOM6@be923c764 (*)Correct USE_STANLEY domain extents in EOS calls
- NOAA-GFDL/MOM6@47cc14971 +(*)Fix numerous issues with MOM_stoch_eos


Hallberg-NOAA:
This PR was intended to go in to dev/gfdl.

Issue #270, opened 2016-04-25
Failure to compile with PGI


kshedstrom:
I'm getting this problem with PGI on our systems:
```
mpif90 -Duse_libMPI -Duse_netCDF -DSPMD -Duse_LARGEFILE   -Kieee -I/usr/mpi/gcc/openmpi-1.8.5/include -I/usr/include -I/usr/local/pkg/netcdf/netcdf-4.3.0.pgi-13.4/include -DPGI -O0 -g -C   -c -I/center/d/kate/ESMG/ESMG-configs/src/FMS/include -I/center/d/kate/ESMG/ESMG-configs/src/FMS/mpp/include	/center/d/kate/ESMG/ESMG-configs/src/FMS/mpp/mpp.F90
PGF90-S-0155-Could not resolve generic procedure mpp_min (/center/d/kate/ESMG/ESMG-configs/src/FMS/mpp/include/mpp_comm_mpi.inc: 220)
PGF90-S-0155-Could not resolve generic procedure mpp_max (/center/d/kate/ESMG/ESMG-configs/src/FMS/mpp/include/mpp_comm_mpi.inc: 221)
PGF90-S-0155-Could not resolve generic procedure mpp_sum (/center/d/kate/ESMG/ESMG-configs/src/FMS/mpp/include/mpp_comm_mpi.inc: 222)
PGF90-S-0155-Could not resolve generic procedure mpp_sum (/center/d/kate/ESMG/ESMG-configs/src/FMS/mpp/include/mpp_comm_mpi.inc: 223)
PGF90-S-0155-Could not resolve generic procedure mpp_min (/center/d/kate/ESMG/ESMG-configs/src/FMS/mpp/include/mpp_comm_mpi.inc: 247)
PGF90-S-0155-Could not resolve generic procedure mpp_max (/center/d/kate/ESMG/ESMG-configs/src/FMS/mpp/include/mpp_comm_mpi.inc: 248)
PGF90-S-0155-Could not resolve generic procedure mpp_sum (/center/d/kate/ESMG/ESMG-configs/src/FMS/mpp/include/mpp_comm_mpi.inc: 249)
PGF90-S-0155-Could not resolve generic procedure mpp_sum (/center/d/kate/ESMG/ESMG-configs/src/FMS/mpp/include/mpp_comm_mpi.inc: 250)
[...]

PR #1591, opened 2023-02-01
ungridded dimension for pstokes

jiandewang:
this is a minor PR to implements use of ungridded dimensions for the partitioned stokes drift by removing the exchange of 6 individual fields in favor of 2 fields containing 3 ungridded dimensions each.

The WW3 mesh cap has the capability to use ungridded dimensions to transfer fields. This allows the 6 fields (3 each for x,y) used to exchange the partitioned stokes drift between WW3 and MOM6 to be simplified to only 2 fields. The MOM6 and WW3 caps used in UFS can be simplified as a result.


sanAkel (commented):
@jiandewang These files are outside the scope of what we use (since they pertain to nuopc). For that reason, I'll decline reviewing this PR. 

@marshallward would you please "adjust" that from 5 reviews policy if not enough folks review it. Thank you!

jiandewang:
> @jiandewang These files are outside the scope of what we use (since they pertain to nuopc). For that reason, I'll decline reviewing this PR.
> 
> @marshallward would you please "adjust" that from 5 reviews policy if not enough folks review it. Thank you!

@sanAkel fully understand, I think @kshedstrom will be on the same boat as you

marshallward:
@sanAkel your review is primarily an acknowledgement that the change will not impact your runs.  Critical feedback is greatly appreciated, of course, but it's not necessary to understand everything coming in.

Which is to say, just click "approved" if it doesn't affect you :).

jiandewang:
> The CI failures appear to be from mom-ocean/main, not EMC. Most likely due to software stack changes on GitHub's end.
> 
> Merging dev/gfdl into this PR does appear to work, so it's likely that the problem will go away after merging dev/gfdl into main. I would feel more comfortable fixing this before accepting this PR though, so we may want to cherry-pick a fix into either main dev/emc before accepting.
> 
> I can't see what the actual problem could be, however, so I'm not yet sure how to proceed.
> 
> Currently, the problem is a FMS library link failure. The FMS library can be built, and module include test passes (`fms_mod.mod` can be found and read by the compiler), but the link test fails because it cannot find any of the symbols.

@marshallward it might be simple to do cherry-pick into this PR branch but I need your instruction on how to do the cherry-pick

marshallward:
The fix appears to be the change from MPICH to OpenMPI.  I don't recall why this was needed, but I didn't think it was related to FMS linking.

Anyway, I agree with your suggestion to cherry-pick into the EMC pull request, I'll send you some instructions if needed.

jiandewang:
> The fix appears to be the change from MPICH to OpenMPI. I don't recall why this was needed, but I didn't think it was related to FMS linking.
> 
> Anyway, I agree with your suggestion to cherry-pick into the EMC pull request, I'll send you some instructions if needed.

@marshallward you mean https://github.com/NOAA-GFDL/MOM6/pull/270, right ?

sanAkel (approved):
These changes fall **outside** the scope of what we use. Therefore these changes do not impact our answers, though they look okay and I am happy to see that `if (cesm_coupled)` has been removed. 

These sort of blocks (or equivalent `ifdef` directives always come in the way of making our couplers more _uniform_ and _seamless_!

jiandewang:
> Yes, thats the one. (Looks like Alistair did not know why either.)
> 
> My guess: The MPI calls in FMS have some type mismatches, which is why we need the `-fallow-argument-mismatch` flag. Perhaps these still work in OpenMPI but not in MPICH.
> 
> I've sent some basic instructions, but get in touch if it doesn't work.

thanks for Marshall's help, this PR branch has cherry-picked Alistair's fixing on CI issue

kshedstrom (approved):
I can confirm that this does not break my code since I'm not linking in the changed files.

abozec (approved):
COAPS does not use the nuopc_cap either so it won't have any effect on our configurations. I approve.

alperaltuntas (approved):

Hallberg-NOAA:
I am not seeing anything here that I expect would present a problem from the GFDL side, especially as all these changes are confined to the NUOPC coupler, but given the potentially strong interactions with the wave interfaces, I would like to defer formal approval from GFDL until we hear what @breichl thinks about this PR.

breichl:
> I am not seeing anything here that I expect would present a problem from the GFDL side, especially as all these changes are confined to the NUOPC coupler, but given the potentially strong interactions with the wave interfaces, I would like to defer formal approval from GFDL until we hear what @breichl thinks about this PR.

I don't foresee any conflicts with our use of the wave interfaces from the proposed code updates.  I recommend approving from GFDL.

marshallward:
This has passed our CI and review; GFDL approves this PR.

marshallward (approved):

jiandewang:
since we got green lights from all reviewers, I am going to merge it soon.

PR #1593, opened 2023-02-28
update how UFS names restart files

jiandewang:
in order to have an unified style of restart file naming among UFS sub components, this mini PR 

- write all restarts as YYYYMMDD.HHMMSS.MOM.resX
- includes stoch restarts
- write final restart with timestamp
- change nems=>ufs

the code changes are only within ufs loop in nuopc drive so it's unlikely to have impact for other users. 
As usual we ask reviewers to help us check the code to make sure it won't break your current settings.


sanAkel (approved):
As noted, these changes are limited in their scope to the NUOPC coupler- that seem reasonable for the said purpose.

(However, if I may suggest, rather than doing this way, perhaps best to have flexibility in your workflow tool- anyway do as you wish!)

alperaltuntas (approved):
No impact on CESM.

abozec (approved):
COAPS approves

kshedstrom (approved):
The indentation is not according to MOM6 guidelines in config_src/drivers/nuopc_cap/mom_cap.F90, lines 1702-1705

I'll approve subject to this being fixed.

marshallward (approved):
GFDL approves this PR.

We agree with @kshedstrom with respect to whitespace in the `write()` statements (specifically after commas), but we're OK with fixing this up in a later PR.

jiandewang:
thanks for all reviewers' quick action, we will fix the extra whitespace in a later PR.

PR #1604, opened 2023-07-28
Entrain denom patch

MJHarrison-GFDL:
This prevents possible division by zero in the layered entrainment module. This situation was encountered during ice shelf model development. This appears to be occurring in ice cavities where the top-most layers are vanished.


Hallberg-NOAA:
This patch should come in via one of the institutional development branches (e.g, dev/gfdl or dev/ncar).  Regardless, a 2-line description is an insufficient for a PR with 109 commits, changing 150 files, with 12874 lines of code changed!

PR #1608, opened 2023-09-08
add run time information in NUOPC cap

jiandewang:
this mini PR added some run time information in NUOPC cap for us to monitor the UFS running speed. A logic variable "write_runtimelog" is added here with default value of False. As usual I am asking MOM6 major users for the review although only NCAR may use this feature (and we asked @alperaltuntas and @marshallward for a pre-review before this formal PR). Thanks for your very helpful suggestions.


sanAkel (approved):
Changes are in the NUOPC driver that is not used in [GEOS](https://github.com/GEOS-ESM/) application.

marshallward (approved):
I am approving this, but I would like to repeat my concern about using `MPI_Wtime` in this driver.  The function can become blocking if certain flags are set, which will block across the entire `MPI_COMM_WORLD`, not just MOM6.  If other ranks are not calling `MPI_Wtime`, then the model will hang.

We recognize that your groups are maintaining the NUOPC cap, so I won't let this concern hold up the patch.  But if a future NUOPC user encounters problems with this, then we may want to revisit this change.

abozec (approved):
We don't use the NUOPC cap in our applications, so it won't affect us.

kshedstrom (approved):
This code doesn't affect me either.

alperaltuntas (approved):

jiandewang:
thanks for everybody's time and effect, I am going to merge it

PR #1613, opened 2023-10-26
Update .readthedocs.yml configuration

marshallward:
Newest Read the Docs configuration file requires explicit specification of the environment (using `build:`).  This patch includes this section.

Upgrading to newer Python environments has also forced us to specify an older Jinja2 that works with our legacy Sphinx module.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,4 +1,13 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
 version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
 
 # Extra formats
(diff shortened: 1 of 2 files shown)


marshallward:
This is a very minor PR which only modifies Read The Docs configuration.  Currently, our documentation generation does not work and gives an error.

Although documentation is produced from the NOAA-GFDL fork, it is generated from our copy of `main`, so we need this to go through the MOM-ocean fork before we can update our own copy.

This should not affect the others in any way and should be quick to approve.

marshallward:
Reminder to have a look at this PR.  It is only to fix a minor documentation issue.

jiandewang (approved):

kshedstrom (approved):

gustavo-marques (approved):

sanAkel (approved):

PR #1617, opened 2023-12-15
Disable FPEs in MacOS testing

marshallward:
Due to poor handling of floating point in HDF5 1.14.3, it is currently not possible to use floating point exceptions (FPEs) whenever this version is present.

The GitHub Actions CI nodes would randomly select either 1.14.2 or 1.14.3, and would raise an FPE error if 1.14.3 was selected. Additionally, the homebrew installation does not provide a clean method for selecting a different version of HDF5.

Thus, for now we disable FPEs in the MacOS testing, and hope to catch any legitimate FP errors in the Ubuntu version.  We will restore these tests as soon as this has been fixed in an easily-accessible version of HDF5.

As part of this PR, I have also moved the FCFLAGS configuration to the platform specific Actions files, allowing for independent compiler configuration for each platform.

### DIFF
--- a/.github/actions/macos-setup/action.yml
+++ b/.github/actions/macos-setup/action.yml
@@ -16,3 +16,18 @@ runs:
         brew install netcdf-fortran
         brew install mpich
         echo "::endgroup::"
+
+    # NOTE: Floating point exceptions are currently disabled due to an error in
+    # HDF5 1.4.3.  They will be re-enabled when the default brew version has
+    # been updated to a working version.
+
+    - name: Set compiler flags
+      shell: bash
+      run: |
+        cd .testing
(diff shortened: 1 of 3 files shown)


marshallward:
This is an "urgent" PR which fixes the random MacOS CI failures.  It should not affect the pending PR or any ongoing work by any of the groups.

It would be great if we can get this merged quickly, but if not then it's also possibly to cherry-pick this one into your forks.

sanAkel (approved):

kshedstrom (approved):
I approve this PR.

jiandewang (approved):

abozec (approved):

marshallward:
Thanks all for expediting this PR.

PR #1619, opened 2024-02-01
User/wfc/restart ensemblename fix

wfcooke:
Fix issue with restart filenames when using ensembles.
    
    Moved call for filename_appendix outside of loop.
    
    Fixes issue where restart filename was
    ./MOM.res.ens_01.nc
    ./MOM.res.ens_01.ens_01_1.nc
    ./MOM.res.ens_01.ens_01.ens_01_2.nc
    ./MOM.res.ens_01.ens_01.ens_01.ens_01_3.nc
    
    instead of
    ./MOM.res.ens_01.nc
    ./MOM.res.ens_01_1.nc
    ./MOM.res.ens_01_2.nc
    ./MOM.res.ens_01_3.nc

This only happened when restart files were larger than some limit (4GB?) so ensembles of 1/4 degree ocean will show this.


wfcooke:
Added update to wrong repository

PR #1621, opened 2024-03-19
NCAR to main 2024-03-19

alperaltuntas:
This PR contains enhancements and bug fixes from NCAR fork as listed below. In summary, these include (1) automated runtime land block elimination based on automatically generating a mask_table during domain initialization at runtime, (2) rotational symmetry fixes and efficiency improvements in Leith+E, and misc bug fixes. 

- https://github.com/NCAR/MOM6/pull/262
- https://github.com/NCAR/MOM6/pull/263
- https://github.com/NCAR/MOM6/pull/267
- https://github.com/NCAR/MOM6/pull/268

Since this PR includes changes in the NUOPC cap, I'd recommend EMC to first run their tests and approve, before others run their full test suites.


kshedstrom (approved):
This doesn't change answers for me, so I approve. However, it changed the MOM_domains_init interface which requires PR #207 to SIS2. Also, a case of mine which is coupled to ice, but has no ice, now needs to have TOPO_CONFIG and/or TOPO_FILE set in SIS_input.

src/parameterizations/lateral/MOM_hor_visc.F90
```diff
@@ -298,8 +298,9 @@ subroutine horizontal_viscosity(u, v, h, diffu, diffv, MEKE, VarMix, G, GV, US,
     dudx, dvdy, &    ! components in the horizontal tension [T-1 ~> s-1]
     dudx_smooth, dvdy_smooth, & ! components in the horizontal tension from smoothed velocity [T-1 ~> s-1]
     GME_effic_h, &  ! The filtered efficiency of the GME terms at h points [nondim]
-    htot, &       ! The total thickness of all layers [H ~> m or kg m-2]
-    m_leithy      ! Kh=m_leithy*Ah in Leith+E parameterization [L-2 ~> m-2]
+    m_leithy, &   ! Kh=m_leithy*Ah in Leith+E parameterization [L-2 ~> m-2]
+    Ah_sq, &      ! The square of the biharmonic viscosity [L8 T-2 ~> m8 s-2]
+    htot          ! The total thickness of all layers [Z ~> m]
```
Hallberg-NOAA, line 303: The units of `htot` in this comment should be `[H ~> m or kg m-2]`, as they were before.

src/framework/MOM_domains.F90
```diff
@@ -60,11 +63,12 @@ module MOM_domains
 !> MOM_domains_init initializes a MOM_domain_type variable, based on the information
 !! read in from a param_file_type, and optionally returns data describing various
 !! properties of the domain type.
-subroutine MOM_domains_init(MOM_dom, param_file, symmetric, static_memory, &
```
Hallberg-NOAA, line 63: Adding a mandatory `unit_scaling_type` argument to `MOM_domains_init()` is going give us problems with SIS2 being unable to compile with successive versions of MOM6, as noted by @kshedstrom.  Were MOM6 and SIS2 under the same version control, this would not be an issue, but unfortunately they are.  There is a simple solution, though.  If `US` were made into an optional argument to `MOM_domains_init()` and `gen_auto_mask_table()`, with `m_to_Z` defaulting to 1.0 when `US` is not present, everything here should work exactly as it does now, but SIS2 would be able to work across versions of the code and have the new `US` type argument added in an orderly way.
I would be happy to prepare a pull request to dev/ncar-candidate-240319 making these changes if it would help.

alperaltuntas: Thanks, @Hallberg-NOAA! I'd appreciate it if you could submit a PR to dev/ncar-candidate-240319 with your proposed change.

Hallberg-NOAA: @alperaltuntas, I have submitted the PR to dev/ncar with my suggested changes at https://github.com/NCAR/MOM6/pull/272.  If these modifications are acceptable, I do not see any other issues within the larger PR from dev/ncar to main that I think will need to be considered.  I have run these changes through my version of the GFDL regression tests, and they are passing for me, although of course we will have to run the final revised version through our full suite of regression tests once it is available.

alperaltuntas: Thanks! I just merged the changes.

sanAkel:
@alperaltuntas what's the highest (finest) resolution that has been tested for the `automatic land mask generation`? Any performance issues? (just asking!)

alperaltuntas:
> @alperaltuntas what's the highest (finest) resolution that has been tested for the `automatic mask table generation`? Any performance issues? (just asking!)

The highest I tried was 0.25 deg. If you are asking about the performance of the iteration, it's quite fast for both 0.66 and 0.25 degrees (on the order of 0.1sec). If we discover in the future that it is slow for finer resolutions (e.g., 1/12 deg), I have plans to speed up the iteration, but for now, I opted for a simpler implementation for ease of maintenance. Other than that, I have not encountered any performance issues.

abozec (approved):
COAPS approves.

sanAkel:
> > @alperaltuntas what's the highest (finest) resolution that has been tested for the `automatic mask table generation`? Any performance issues? (just asking!)
> 
> The highest I tried was 0.25 deg. If you are asking about the performance of the iteration, it's quite fast for both 0.66 and 0.25 degrees (on the order of 0.1sec). If we discover in the future that it is slow for finer resolutions (e.g., 1/12 deg), I have plans to speed up the iteration, but for now, I opted for a simpler implementation for ease of maintenance. Other than that, I have not encountered any performance issues.

That ( `on the order of 0.1sec`) sounds great! Thank you!

jiandewang:
@alperaltuntas what is automatic mask table generation ?

jiandewang:
@alperaltuntas if we use land mask, then the lat/lon information on masked domain will be written as undefined value, is that true ? See an example at
https://www.emc.ncep.noaa.gov/gc_wmb/wd20xw/JD/ocn_2013_04_01_06-geolon.txt

jiandewang:
@alperaltuntas one more minor point: in restart file variable "h" are all 1e-10 on land points but using land mask approach pure land domian have some points (but not all of them) with value of zero. All ocean points have identical values which is what we expected.
see samples files at https://www.emc.ncep.noaa.gov/gc_wmb/wd20xw/JD/NCAR/

jiandewang (approved):
finally machine is back to normal. Sorry for the delay

marshallward (approved):
This is good on the GFDL side now, and this version passes our regression testing.

sanAkel (approved):

PR #1659, opened 2025-04-02
Docs: Force CMake 3.5 compatibility

marshallward:
The Doxygen CMake build compatibility is increase to 3.5 in order to enable usage of newer versions of CMake.

This is not ideal, and is an indication that we need to overhaul and update our documentation software stack, but this is a short-term solution.

Thanks to @mnlevy1981 for identifying and investigating this issue.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -195,8 +195,8 @@ doxygen:
 	git clone -b $(DOXYGEN_RELEASE) --depth 1 https://github.com/doxygen/doxygen
 
 doxygen/bin/doxygen: doxygen
-	(cd $(<); cmake -G "Unix Makefiles" .)
-	(cd $(<); make)
+	(cd $(<); cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -G "Unix Makefiles" .)
+	(cd $(<); $(MAKE))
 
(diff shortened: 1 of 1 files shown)


marshallward:
This is an emergengy fix to restore the documentation stage of the CI.  Review should be very quick, assuming there aren't any objections.

@dougiesquire It looks like ACCESS-NRI has not yet been added as a reviewer (which we can change at any time) but your review would also be very appreciated.

jiandewang (approved):
since all checks passed, approve

dougiesquire (approved):
> @dougiesquire It looks like ACCESS-NRI has not yet been added as a reviewer (which we can change at any time) but your review would also be very appreciated.

Thanks @marshallward. Please do add us as a reviewer.

kshedstrom (approved):
I approve this PR.

abozec (approved):
COAPS approves

marshallward:
@dougiesquire I've added you as the ACCESS reveiwer, and the total number of required reviewers has been increased to six.

@alperaltuntas Have you had a chance to look at this?

gustavo-marques (approved):

marshallward:
Thanks all.  This should fix the doxygen/style test failures.  And is a reminder that we need to clean up the documentation build system.

PR #1661, opened 2025-04-15
Use sorted chksum_diag for diagnostic regression

marshallward:
This PR has two commits:

* A timestamp is appended to the checksum output of `chksum_diag`, to facilitate sorting.
* CI testing sorts this output in order to eliminate false-negative diagnostic regressions.


marshallward:
oops, on many levels.

PR #1670, opened 2025-08-05
+*Update parameter defaults for 2024 and 2025, including EQN_OF_STATE

Hallberg-NOAA:
This series of 3 commits updates the defaults for a total of 21 runtime parameters, and obsoletes another 7 runtime parameters, as agreed to in MOM6 consortium-wide meetings on July 29, 2024, December 16, 2024 and  June 2, 2025.

 The default equation of state is now `EQN_OF_STATE = "WRIGHT_FULL"`, in place of the buggy previous default of `"WRIGHT"`.

  The 8 answer date parameters `REGRIDDING_ANSWER_DATE`, `TIDES_ANSWER_DATE`, `WAVE_INTERFACE_ANSWER_DATE`, `MEKE_GM_SRC_ANSWER_DATE`, `NDIFF_ANSWER_DATE`, `KPP%ANSWER_DATE`, `HOR_DIFF_ANSWER_DATE` and `LOTW_BBL_ANSWER_DATE` all now take their default values from `DEFAULT_ANSWER_DATE`.

  The bug-retention parameters `MEKE_GM_SRC_ALT_SLOPE_BUG`, `HOR_DIFF_LIMIT_BUG`, `BACKSCATTER_UNDERBOUND`, `DETERMINE_TEMP_CONVERGENCE_BUG`, `LA_MISALIGNMENT_BUG`, `IDL_HURR_SCM_EDGE_TAPER_BUG`, `VISC_REM_BUG` and `FRICTWORK_BUG` are now false by default.

  With these changes, `MASS_WEIGHT_IN_PRESSURE_GRADIENT_TOP` now takes its default value from the setting for `MASS_WEIGHT_IN_PRESSURE_GRADIENT`.  Similarly, the default for `DRAG_DIFFUSIVITY_ANSWER_DATE` is now set to follow `SET_DIFF_ANSWER_DATE`. `KELVIN_WAVE_VEL_NUDGING_TIMESCALE` is now set to a negative value by default, forcing the user to provide a valid value at runtime. `SHELFWAVE_CORRECT_AMPLITUDE` now defaults to True.

  The parameters `BETTER_BOUND_KH`, `BETTER_BOUND_AH`, `USE_DIABATIC_TIME_BUG`, `FIX_UNSPLIT_DT_VISC_BUG`, `CFL_BASED_TRUNCATIONS` and `KD_BACKGROUND_VIA_KDML_BUG` have all been obsoleted and are no longer valid MOM6 runtime parameters.

[...]

PR #1671, opened 2025-08-05
GH Actions PR test

marshallward:
Testing if GitHub Actions test launch is working in mom-ocean.


marshallward:
Actions are launching, but does not yet resolve the mystery of absent actions in #1670

PR #1675, opened 2025-08-14
DEV-EMC-20250813 (convert missing value to zero in Stokes drift)

jiandewang:
The nuopc_cap and mom_cap_methods.F90 are revised to convert imported Stokes drift 
components from a missing value (9.99e20) to zero. This update doesn't have any impacts on 
global applications and are only needed when the ocean model domain is greater 
than the wave model domain while the wave model cannot provide Stokes drift components
for the entire ocean model domain.

Thanks for @alperaltuntas to do a pre-test for this PR.


jiandewang:
We apologize for some extra commits in this PR. What happened is we had an internal updating on ISO code which happened right before GFDL's 20250423 PR (https://github.com/mom-ocean/MOM6/pull/1664). In GFDL's PR one variable was removed and unfortunately that variable was being used/modified in EMC's ISO code. During the test and merging time we found there is no simply way to solve the code conflict. As a consequence we had to manually changed back (git revert is not allowed in UFS as it will have big negative impact for all UFS developers' onging work) our ISO code in order to merge GFDL's PR. We added ISO code updating back after GFDL's PR was merged in dev/emc.

we had an internal discussion on this extra commits issue but found there is no simply way to handle it, so please treat this case as an exception and we will make sure to follow "one updating, one hash" style as close as possible.

mathomp4 (approved):
GEOS does not even compile these files, so easy approval from GMAO.

abozec (approved):
We do not use these routines either at COAPS so we approve.

kshedstrom (approved):
I'm not using these files either. Approve!

jiandewang:
@alperaltuntas can you review it ?

alperaltuntas (approved):

marshallward:
There is no problem merging this commit.  We often have to make revisions to the PRs which are not part of our development branches.

EMC will pick up d6d211f when you merge `main` back into `dev/emc`.

PR #1690, opened 2026-02-27
Add CLAUDE.md guidance for agentic development with Claude Code

glwagner:
This PR adds a `CLAUDE.md` document to help guide development with [Claude Code](https://code.claude.com/docs/en/overview). This document is loaded into the context of the model Claude is using when its started up inside the MOM6 directory. An introduction to the basic purpose of `CLAUDE.md` is written here:

https://www.humanlayer.dev/blog/writing-a-good-claude-md

The reason to add this to the repo itself is because, in addition to providing MOM6-specific guidance to Claude for any development task, the `CLAUDE.md` can also be used to catch "common mistakes" that Claude makes (eg it can say "make sure you don't do _this_"). This is actually not foolproof; sometimes Claude will make a mistake even if it is explicitly told not to. However, this can often be fixed with something like "review the code you've written and make sure you don't make any of the mistakes in CLAUDE.md". The common mistakes are things that have to be found empirically by using CLAUDE.md, so including it in the repo allows all MOM6 develops to collaborate on strengthening and improving the Claude dev workflow by contributing to `CLAUDE.md`.

Including this was an idea that came out of a conversation with @StephenGriffies !

I'll note that one can go much further with guidelines for Claude; for example many .md files can be developed for specific tasks. And so on.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -105,6 +105,8 @@ Key rules:
 - **No global/module data** -- all state lives in control structures passed as arguments
 - **All arguments must have declared `intent`** (`in`, `out`, or `inout`); pointers exempt
 - **Every module has `_init` and `_end` subroutines** for lifecycle management
+- **`! Local variables`** comment separates dummy arguments from local declarations in subroutines
+- **Prefer `allocatable` over `pointer`** for control structure members
 
 ### Naming Conventions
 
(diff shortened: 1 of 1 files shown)


glwagner:
I'll reopen this pointing to dev/gfdl (where I started from) but happy to take advice where this should point (it just adds one file so it should be trivial...)

PR #1692, opened 2026-02-28
Fixes for flang in MOM_remapping.F90

### LINKED ISSUE #1689: Fortran Standards violation in MOM_remapping.F90
Today I was trying to build GEOS with LLVM Flang 22.1.0. In doing so, it fails when trying to build `MOM_remapping.F90`. It turns out there is violation of the Fortran Standards that, apparently, `ifort`, `ifx` and `gfortran` seem to allow *in a very specific case*. But `flang` (and `nagfor`) do not like it ever.

The issue is with lines like:
```fortran
   type(PCM) :: PCM
```
where you have a variable with the same name as a type. Every compiler does not like this:
```
> cat test.F90
type :: PCM
   integer :: a
end type PCM

type(PCM) :: PCM

end

> ifx test.F90
test.F90(5): error #6406: Conflicting attributes or multiple declaration of name.   [PCM]
type(PCM) :: PCM
-------------^
compilation aborted for test.F90 (code 1)
```

But, as I said, `ifx` (and GNU) seem to allow it in the case of how MOM is doing it. I've "reported" it to Intel here:

https://community.intel.com/t5/Intel-Fortran-Compiler/Possible-Fortran-Standards-Violation-in-ifx-2025-3/m-p/1738576#M178401

Now the fix to this is pretty simple: rename something. And I could easily make a PR for this. 

My question to @Hallberg-NOAA et al is what would be the "MOM standard" way of doing this? For example, one could rename the type:

```fortran
use Recon1d_PCM, only : PCM_t => PCM
...
      allocate( PCM_t :: CS%reconstruction )
...
  type(PCM_t) :: PCM
```
or you could rename the variable. Not sure which is best. 

And, what to call the rename is another decision. `PCM_t`? `PCM_type`? `PCM_local` (for the variable)...

mathomp4:
Closes #1689 

This PR fixes #1689 by renaming the instance of a type. Fortran does not allow (or is unhappy with):
```fortran
type(PCM) :: PCM
```
Both `ifort`/`ifx` and GNU seem to allow this, but stricter compilers like `nagfor` and `flang` do not.

So we do:
```fortran
type(PCM) :: PCM_instance
```
and then rename the variable use below, e.g.:
```fortran
  call test%test( PCM_instance%unit_tests(verbose, test%stdout, test%stderr), 'PCM unit test')
```

NOTE: GEOS does not build *all* of MOM6, for example, the NUOPC code, but we do a good chunk. This was the only error Flang flagged.

### DIFF
--- a/src/ALE/MOM_remapping.F90
+++ b/src/ALE/MOM_remapping.F90
@@ -19,22 +19,22 @@ use PQM_functions, only : PQM_reconstruction, PQM_boundary_extrapolation_v1
 use MOM_hybgen_remap, only : hybgen_plm_coefs, hybgen_ppm_coefs, hybgen_weno_coefs
 
 use Recon1d_type, only : Recon1d
-use Recon1d_PCM, only : PCM
-use Recon1d_PLM_CW, only : PLM_CW
-use Recon1d_PLM_hybgen, only : PLM_hybgen
-use Recon1d_PLM_CWK, only : PLM_CWK
-use Recon1d_MPLM_CWK, only : MPLM_CWK
-use Recon1d_EMPLM_CWK, only : EMPLM_CWK
-use Recon1d_MPLM_WA, only : MPLM_WA
-use Recon1d_EMPLM_WA, only : EMPLM_WA
-use Recon1d_MPLM_WA_poly, only : MPLM_WA_poly
-use Recon1d_EMPLM_WA_poly, only : EMPLM_WA_poly
-use Recon1d_PPM_CW, only : PPM_CW
-use Recon1d_PPM_hybgen, only : PPM_hybgen
-use Recon1d_PPM_CWK, only : PPM_CWK
-use Recon1d_EPPM_CWK, only : EPPM_CWK
-use Recon1d_PPM_H4_2019, only : PPM_H4_2019
-use Recon1d_PPM_H4_2018, only : PPM_H4_2018
+use Recon1d_PCM, only : PCM_t => PCM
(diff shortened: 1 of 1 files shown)


adcroft (changes requested):
I agree that it fixes #1689 and I get the gist of the change. It's short and sweet, but it could be even shorter. How about if renaming the local variables in `remapping_unit_tests()`? There, they are declared once and used once, whereas this PR currently renames the import at the top of the module, changes the allocation in `setReconstructionType()`, and then changes the local declaration `remapping_unit_tests()`.

marshallward:
> How about if renaming the local variables in remapping_unit_tests()?

When something like this came up for AMD a year or so ago, this is how we decided to resolve the issue.

As for the naming, I'd prefer not to see `_t` suffix, and instead consider names that describe themselves as types.  There is also the fact that `_t` is reserved for POSIX types, albeit in C.  (Linux forbids `_t` IIRC.)

mathomp4:
That is, I can test whatever you'd like as a rename:

* `PCM_r`
* `PCM_instance`
* `PCM_local_variable`

I just wasn't sure if there was an "official" MOM standard of how to name instances of types, etc. 🙂

adcroft (approved):
We like this solution.

alperaltuntas (approved):

jiandewang (approved):
works fine in UFS

dougiesquire (approved):

abozec (approved):
COAPS approves.

kshedstrom (approved):

Issue #86, opened 2015-01-23
need check for npes > ni, nj


StephenGriffies:
There should be an "idiot" check for the case where npes is greater than the number of grid points in the model.  For example, with the SCM tests, the model crashes when npe>1, but there is no error message that tells the user what went wrong.  It is obvious after-the-fact, but for new users (e.g., Griffies), it would prove very useful to have this sort of error message, particularly since the example run command on the wiki has "aprun -n 8".

PR #1700, opened 2026-04-22
NCAR to main (2026-06-09)

chengz2:
### Extensions / Improvements

- NCAR/MOM6#340 @mnlevy1981
- NCAR/MOM6#342 @alperaltuntas
- NCAR/MOM6#347 @gustavo-marques
- NCAR/MOM6#349 @gustavo-marques
- NCAR/MOM6#353 @alperaltuntas
- NCAR/MOM6#358 @iangrooms
- NCAR/MOM6#359 @iangrooms
- NCAR/MOM6#364 @gustavo-marques
- NCAR/MOM6#366 @alperaltuntas
- NCAR/MOM6#368 @Hallberg-NOAA
- NCAR/MOM6#374 @mnlevy1981
- NCAR/MOM6#376 @iangrooms
- NCAR/MOM6#378 @iangrooms
- NCAR/MOM6#379 @gustavo-marques
- NCAR/MOM6#387 @gustavo-marques
- NCAR/MOM6#388 @iangrooms
- NCAR/MOM6#389 @gustavo-marques
- NCAR/MOM6#392 @mnlevy1981
- NCAR/MOM6#415 @gustavo-marques

### Bug fixes / Code optimizations

- NCAR/MOM6#357 @alperaltuntas 
- NCAR/MOM6#367 @Hallberg-NOAA 
- NCAR/MOM6#371 @gustavo-marques 
- NCAR/MOM6#375 @phall-brown 
- NCAR/MOM6#377 @gustavo-marques 
- NCAR/MOM6#390 @gustavo-marques 
- NCAR/MOM6#395 @alperaltuntas 
- NCAR/MOM6#396 @alperaltuntas 
- NCAR/MOM6#398 @alperaltuntas 
- NCAR/MOM6#401 @mnlevy1981 
- NCAR/MOM6#403 @iangrooms 
- NCAR/MOM6#409 @alperaltuntas 
- NCAR/MOM6#414 @alperaltuntas 
- NCAR/MOM6#418 @alperaltuntas 
- NCAR/MOM6#419 @mnlevy1981 
- NCAR/MOM6#421 @gustavo-marques 
- NCAR/MOM6#423 @klindsay28 
- NCAR/MOM6#424 @gustavo-marques 
- NCAR/MOM6#425 @alperaltuntas 
- NCAR/MOM6#427 @alperaltuntas 
- NCAR/MOM6#429 @alperaltuntas 
- NCAR/MOM6#430 @Hallberg-NOAA 
- NCAR/MOM6#431 @Hallberg-NOAA 
- NCAR/MOM6#438 @chengz2  

### New input parameters

```
USE_UH_PARTICLES = False        !   [Boolean] default = False
                                ! If true, use the uh velocity in the particles package.
USE_STANLEY_ML = False          !   [Boolean] default = False
                                ! If true, turn on Stanley SGS T variance parameterization in ML restrat code.
STOCHASTICS_ANSWER_DATE = 20250101 ! default = 20250101
                                ! The vintage of the order of arithmetic in the stochastics calculations.
                                ! Values below 20250701 recover the answers from early in 2025, while higher
                                ! values use expressions that have been refactored for rotational symmetry.
ENSEMBLE_APPENDIX_PREFIX = ""   ! default = ""
                                ! If set to a non-empty string, this value specifies the substring after which
                                ! the ensemble appendix is inserted in restart, initial conditions, and ocean
                                ! geometry file names. If the specified substring is not found in any of those
                                ! output file names, the model terminates with an error.
KPP%KPP_ER_Cb = 0.96                !   [nondim] default = 0.96
                                ! Entrainment Rule TKE buoyancy production weight
KPP%KPP_ER_Cs = 0.038               !   [nondim] default = 0.038
                                ! Entrainment Rule TKE Stokes production weight
KPP%KPP_ER_Cu = 0.023               !   [nondim] default = 0.023
                                ! Entrainment Rule TKE shear production weight
KPP%KPP_OBL_DEPTH_BOUNDS_BUG = False !   [Boolean] default = False
                                ! If true, limit the KPP boundary layer depth relative to the top of the
                                ! bottommost layer instead of the seafloor.
KPP%ANSWER_DATE = 20251231          ! default = 99991231
                                ! The vintage of the order of arithmetic in the CVMix KPP calculations.  Values
                                ! below 20240501 recover the answers from early in 2024, while higher values use
                                ! expressions that have been refactored for rotational symmetry.  Values of
                                ! 20260101 or higher replace some divisions with multiplication by reciprocals.
```

### Contributors:

- @alperaltuntas : 13
- @gustavo-marques : 12
- @Hallberg-NOAA : 4
- @iangrooms : 6
- @mnlevy1981 : 5
- @phall-brown : 1
- @klindsay28 : 1
- @chengz2 : 1


chengz2:
@alperaltuntas, @gustavo-marques and @mnlevy1981 please review this PR and add a description. I'm not sure if any commits should be considered as new features, so I put everything other than bug fixes under Extensions / Improvements.

src/core/MOM.F90
```diff
@@ -395,6 +401,15 @@ module MOM
   type(diag_grid_storage)  :: diag_pre_sync !< The grid (thicknesses) before remapping
   type(diag_grid_storage)  :: diag_pre_dyn  !< The grid (thicknesses) before dynamics
 
+  logical :: accumulate_resolved_flux = .false. !< If true, accumulate resolved flux for tracers for diagnostics
+                                                !! separating the tracer flux due to resolved and parameterized flow
+  logical :: do_resolved_advection = .false.    !< If true, calculate advection by resolved flow
+  logical :: do_param_advection = .false.       !< If true, calculate advection by parameterized flow
+  real ALLOCABLE_, dimension(NIMEMB_PTR_,NJMEM_,NKMEM_) :: &
```
Hallberg-NOAA: These two declarations use the correct macros to work in either static or dynamic memory mode (the code is correct), but in static memory mode these two 3-d arrays will always be given memory as a part of this type, even though they are only used when certain diagnostics are being used.  In the interest of memory efficiency it might be better to simply declare these as:

```
  real, allocatable, dimension(:,:,:) :: &
    uhtr_resolved   !< accumulated zonal thickness fluxes due to resolved flow to advect tracers [H L2 ~> m3 or kg]
  real, allocatable, dimension(:,:,:) :: &
    vhtr_resolved   !< accumulated meridional thickness fluxes due to resolved flow to advect tracers [H L2 ~> m3 or kg]
```
so that memory is only assigned to them if they are being used. (In dynamic memory mode all the macros would resolve to this suggested version.  Note that if this change is made, there would also have to be changes to avoid using macros on the allocate and deallocate calls on lines 3726, 3727 and 4625. 

Note that this is not a bug and should not hold up this PR, but only a suggestion!

src/parameterizations/stochastic/MOM_stochastics.F90
```diff
@@ -94,10 +106,13 @@ subroutine stochastics_init(dt, grid, GV, CS, param_file, diag, Time)
   integer :: pe_zero           !  root pe
   integer :: nxT, nxB          ! number of x-points including halo
   integer :: nyT, nyB          ! number of y-points including halo
+  integer :: default_answer_date ! The default setting for the various ANSWER_DATE flags.
   integer :: i, j, k           ! loop indices
-  real    :: tmp(grid%isdB:grid%iedB,grid%jsdB:grid%jedB) ! Used to construct tapers
+  real    :: tmp(grid%isdB:grid%iedB,grid%jsdB:grid%jedB) ! Used to construct tapers and weights
```
Hallberg-NOAA: There should be a description of the units of the real variables, even including this `tmp` array, per the MOM6 style guide.  If it is used in different places with different units, it should have units of `[various]` or a list of the units that it is used with.

src/parameterizations/vertical/MOM_CVMix_KPP.F90
```diff
@@ -149,39 +155,49 @@ module MOM_CVMix_KPP
...
   real, allocatable, dimension(:,:)   :: Usurf     !< i-velocity of surface layer [L T-1 ~> m s-1]
   real, allocatable, dimension(:,:)   :: Vsurf     !< j-velocity of surface layer [L T-1 ~> m s-1]
   real, allocatable, dimension(:,:,:) :: EnhK      !< Enhancement for mixing coefficient [nondim]
   real, allocatable, dimension(:,:,:) :: EnhVt2    !< Enhancement for Vt2 [nondim]
+  real, allocatable, dimension(:,:)   :: La_SL     !< Langmuir number used in KPP [nondim]
+  real, allocatable, dimension(:,:)   :: OBLdepth_original  !< Depth (positive) of OBL [Z ~> m] without smoothing
+  real, allocatable, dimension(:,:)   :: ERdepth   !< Percent use ER boundary layer depth [nondim]
+  real, allocatable, dimension(:,:)   :: RNdepth   !< Percent use Ri Number boundary layer depth [nondim]
+  real, allocatable, dimension(:,:)   :: StokesXI  !< Stokes similarity parameter  [nondim]
+  real, allocatable, dimension(:,:)   :: BEdE_ER   !< Enrtainment Rule's Parameterized BEdE [ m3 s-3 ]
```
Hallberg-NOAA: I think that "Enrtainment" should be "Entrainment" here, unless this rule is named after the mysterious Dr. Enrtainment.

src/parameterizations/vertical/MOM_CVMix_KPP.F90
```diff
@@ -1239,24 +1311,18 @@ subroutine KPP_compute_BLD(CS, G, GV, US, h, Temp, Salt, u, v, tv, uStar, buoyFl
...
-          !surfUs   = surfHus   * I_hTot
-          !surfVs   = surfHvs   * I_hTot
-
-          surfTemp = surfHtemp / hTot
-          surfSalt = surfHsalt / hTot
-          surfU    = surfHu    / hTot
-          surfV    = surfHv    / hTot
-          surfUs   = surfHus   / hTot
-          surfVs   = surfHvs   / hTot
+          I_hTot = 1./hTot
```
Hallberg-NOAA: Although these code changes are correct and will be more efficient, changing the divisions here into a multiplication by a reciprocal will change answers at roundoff.  Perhaps an answer date flag here would be a good idea.

src/parameterizations/vertical/MOM_CVMix_KPP.F90
```diff
@@ -1317,7 +1382,7 @@ subroutine KPP_compute_BLD(CS, G, GV, US, h, Temp, Salt, u, v, tv, uStar, buoyFl
         if (GV%Boussinesq .or. GV%semi_Boussinesq) then
           deltaBuoy(k) = GoRho*(rho_1D(kk+2) - rho_1D(kk+1))
         else
-          deltaBuoy(k) = (US%Z_to_m*US%s_to_T**2) * GV%g_Earth_Z_T2 * &
+          deltaBuoy(k) = (US%Z_to_m*US%s_to_T**2) * (US%L_to_Z**2 * GV%g_Earth) * &
```
Hallberg-NOAA: The change on this line is correct, but unnecessary and more complicated than it needs to be.  An examination of MOM_vertical_grid.F90 will reveal that `GV%g_Earth_Z_T2 = (US%L_to_Z**2 * GV%g_Earth)`.

src/parameterizations/vertical/MOM_CVMix_KPP.F90
```diff
@@ -1096,9 +1152,9 @@ subroutine KPP_compute_BLD(CS, G, GV, US, h, Temp, Salt, u, v, tv, uStar, buoyFl
   ! some constants
   GoRho = US%Z_to_m*US%s_to_T**2 * (GV%g_Earth_Z_T2 / GV%Rho0)
   if (GV%Boussinesq) then
-    GoRho_Z_L2 = GV%Z_to_H * GV%g_Earth_Z_T2 / GV%Rho0
+    GoRho_Z_L2 = US%L_to_Z**2 * GV%Z_to_H * GV%g_Earth / GV%Rho0
```
Hallberg-NOAA: The changes on lines 1155 and 1157 are not wrong, but they are unnecessary and more complicated than they need to be. An examination of MOM_vertical_grid.F90 will reveal that `GV%g_Earth_Z_T2 = (US%L_to_Z**2 * GV%g_Earth)`, and because `US%L_to_Z` is always an integer power of 2 the order of the multiplication does not matter.

src/parameterizations/stochastic/MOM_stochastics.F90
iangrooms: Thanks for catching this. Looks like it's just used to construct tapers, not weights, and it's nondimensional. So it should be
```suggestion
  real    :: tmp(grid%isdB:grid%iedB,grid%jsdB:grid%jedB) ! Used to construct tapers [nondim]
```

src/parameterizations/vertical/MOM_CVMix_KPP.F90
```diff
@@ -1337,45 +1402,73 @@ subroutine KPP_compute_BLD(CS, G, GV, US, h, Temp, Salt, u, v, tv, uStar, buoyFl
...
       enddo
 
-      ! CVMix_kpp_compute_turbulent_scales_1d_OBL computes w_s velocity scale at cell centers for
-      ! CVmix_kpp_compute_bulk_Richardson call to CVmix_kpp_compute_unresolved_shear
-      ! at sigma=Vt_layer (CS%surf_layer_ext or 1.0) for this calculation.
-      ! StokesVt_1d controls Stokes enhancement  (= 0 for none)
-      Vt_layer = 1.0      ! CS%surf_layer_ext
-      call CVMix_kpp_compute_turbulent_scales( &                             ! 1d_OBL
+      ! Use CS%deepOBLoffset (<-0.1*iFaceHeight(GV%ke+1)) to avoid vanishingly small layers near the bottom.
+      zBottomMinusOffset = iFaceHeight(GV%ke) + min( max(CS%deepOBLoffset,0.0), -0.1*iFaceHeight(GV%ke+1))
```
Hallberg-NOAA: When this line was moved here as a part of https://github.com/NCAR/MOM6/pull/379, the k-index of the first instance of `iFaceHeight` was changed from `GV%ke+1` to `GV%ke`.  I strongly suspect that line 1406 should read
```
zBottomMinusOffset = iFaceHeight(GV%ke+1) + min( max(CS%deepOBLoffset,0.0), -0.1*iFaceHeight(GV%ke+1))
```
Correcting this bug is necessary (but not sufficient) to recover the previous answers in cases using KPP.

src/parameterizations/vertical/MOM_CVMix_KPP.F90
```diff
@@ -1387,86 +1480,65 @@ subroutine KPP_compute_BLD(CS, G, GV, US, h, Temp, Salt, u, v, tv, uStar, buoyFl
...
-                    bfsfc=surfBuoyFlux2, & ! surface buoyancy flux [m2 s-3]
-                    uStar=surfFricVel,  & ! surface friction velocity [m s-1]
-                    CVmix_kpp_params_user=CS%KPP_params ) ! KPP parameters
-        CS%Vt2(i,j,:) = US%m_to_Z**2*US%T_to_s**2 * Vt2_1d(:)
       endif
 
-      ! recompute wscale for diagnostics, now that we in fact know boundary layer depth
+      ! recompute unresolved squared velocity, wscale  and  BulkRi  for known boundary layer depth
+      ! compute unresolved squared velocity for diagnostics
+      ! recompute wscale for diagnostics,
```
Hallberg-NOAA: Line 1539 should be removed or reverted to read "recompute wscale for diagnostics, now that we in fact know boundary layer depth" or otherwise revised to remove the nonsensical trailing comma.

src/parameterizations/vertical/MOM_CVMix_KPP.F90
```diff
@@ -1387,86 +1480,65 @@ subroutine KPP_compute_BLD(CS, G, GV, US, h, Temp, Salt, u, v, tv, uStar, buoyFl
...
+                  Xi = StokesXI_1d,       & ! (in) Stokes similarity parameter Lmob limit (1-Xi)
+                  zBottom = zBottomMinusOffset,     & ! (in) Numerical limit on OBLdepth
+                  CVMix_kpp_params_user=CS%KPP_params ) ! KPP parameters
+
+          CS%OBLdepth(i,j) = US%m_to_Z * KPP_OBL_depth
+          CS%RNdepth(i,j) = 100. ! check and diagnostic
+        endif  ! KPP_OBL_depth
+
+      endif  ! fixedOBLdepth
+      CS%OBLdepth(i,j) = min( CS%OBLdepth(i,j), -zBottomMinusOffset ) ! no deeper than deepOBLoffset off bottom
```
Hallberg-NOAA, line 1556: For reasons that I do not understand, the lower bound of `-iFaceHeight(2)` appears to be above `zBottomMinusOffset` sometimes, so to recover the previous answers we would have to add the line
```
      CS%OBLdepth(i,j) = max( CS%OBLdepth(i,j), -iFaceHeight(2) )     ! no shallower than top layer
```
below line 1502 and above lines 1503 that sets `CS%kOBL`.  If this is done here, the equivalent line 1414 that applies this bound can be removed.

jiandewang:
the newly added "module stochy_data_mod" in https://github.com/chengz2/MOM6/blob/ncar-to-main_2604/config_src/external/stochastic_physics/stochy_data_mod.F90#L3

 share the same name as

https://github.com/NOAA-PSL/stochastic_physics/blob/ae192f119b8614b4476a1b04ba55f6fb93e2953d/stochy_data_mod.F90#L3

which caused UFS compiling failure

src/parameterizations/vertical/MOM_CVMix_KPP.F90
Hallberg-NOAA: This problem can also be addressed by adding the suggested line bounding `CS%OBLdepth` by `-iFaceHeight(2)`  at the end of the `KPP_OBL_depth` block at about line 1498 above, and by bounding `zBottomMinusOffset` not to be shallower than `iFaceHeight(2)` when a nonzero value of `CS%deepOBLoffset` is used.

mnlevy1981:
> the newly added "module stochy_data_mod" in https://github.com/chengz2/MOM6/blob/ncar-to-main_2604/config_src/external/stochastic_physics/stochy_data_mod.F90#L3
> 
> share the same name as
> 
> https://github.com/NOAA-PSL/stochastic_physics/blob/ae192f119b8614b4476a1b04ba55f6fb93e2953d/stochy_data_mod.F90#L3
> 
> which caused UFS compiling failure

@jiandewang UFS will need to update its build system so `config_src/externals/stochastic_physics` is not included when compiling the model. NCAR is also using the NOAA-PSL stochastic physics package, but we included dummy caps in `config_src/externals` in this PR so that users who do not want to build that package can still run MOM6 (provided `DO_SKEB=False`,
`SKEB_USE_GM=False`, and `SKEB_USE_FRICT=False`; those are the default values for those parameters)

src/core/MOM.F90
chengz2: Resolved in PR [#430](https://github.com/NCAR/MOM6/pull/430) to dev/ncar

jiandewang:
@mnlevy1981 my bad, I shouldn't include those files in my cmake list

src/parameterizations/stochastic/MOM_stochastics.F90
chengz2: This is addressed in this [PR](https://github.com/chengz2/MOM6/pull/1).

chengz2: This is addressed in this [PR](https://github.com/chengz2/MOM6/pull/1).

jiandewang:
does this MOM6 PR comes with an associated stochastic_physics PR ?

iangrooms:
> does this MOM6 PR comes with an associated stochastic_physics PR ?

No. (At least not that I recall.) In the future we will want to make some arguments optional in calls to stochastic_physics subroutines, but we are not including that as part of this PR.

mnlevy1981:
@jiandewang @iangrooms it's worth pointing out that CESM uses [a fork of `stochastic physics`](https://github.com/ESCOMP/stochastic_physics/), though it sounds like the three commits on that fork and not on the main NOAA-PSL version are only necessary for running the CESM implementation.

jiandewang:
@mnlevy1981 thanks for the information. This explains why my compiling is failing. In UFS stochastic_physics is using

commit ae192f119b8614b4476a1b04ba55f6fb93e2953d (HEAD, origin/master, origin/HEAD, master)
Merge: aea3150 956d374
Author: JONG KIM <jong.kim@noaa.gov>
Date:   Wed Dec 3 15:27:55 2025 -0500

    Merge pull request #86 from BrianCurtis-NOAA/nco_debug_changes

    Bring in nco DEBUG flag change for ops

mnlevy1981:
@jiandewang I'll talk to @iangrooms about updating the CESM fork of stochastic physics to include more recent changes. Maybe we are calling into a function that has a different API in newer commits, so the code we added in this PR doesn't compile against `ae192f1`?

jiandewang:
@mnlevy1981 this is what I got:

ufs-weather-model/MOM6-interface/MOM6/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90(67): error #6580: Name in only-list does not exist or is not accessible.   [STOCH_RESTFILE]
use stochy_data_mod,         only : stoch_restfile
------------------------------------^
/work/noaa/marine/Jiande.Wang/HERCULES/MOM6-update/NCAR-20260422/ufs-weather-model/MOM6-interface/MOM6/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90(295): error #6404: This name does not have a type, and must have an explicit type.   [STOCH_RESTFILE]
         stoch_restfile = input_restart_file(1:i)//'r_stoch'//input_restart_file(i+2:k)
---------^
/work/noaa/marine/Jiande.Wang/HERCULES/MOM6-update/NCAR-20260422/ufs-weather-model/MOM6-interface/MOM6/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90(295): error #6303: The assignment operation or the binary expression operation is invalid for the data types of the two operands.
         stoch_restfile = input_restart_file(1:i)//'r_stoch'//input_restart_file(i+2:k)
------------------------------------------------------------^
compilation aborted for /work/noaa/marine/Jiande.Wang/HERCULES/MOM6-update/NCAR-20260422/ufs-weather-model/MOM6-interface/MOM6/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90

mnlevy1981:
`stoch_restfile` is something we added so that CESM would remain bit-for-bit, so we may need to push our changes to PSL...

mnlevy1981:
@jiandewang If you add `#ifdef CESMCOUPLED` around the following blocks in `config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90`, does that let you build?

```fortran
use stochy_data_mod,         only : stoch_restfile
```

is [line 67](https://github.com/chengz2/MOM6/blob/97d5eb8f601706c446e312d3df35bd32b3c8ded2/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90#L67) on our branch, and

```fortran
  if(present(input_restart_file)) then
      k = index(input_restart_file, ' ')
      if (k==0) k = len_trim(input_restart_file)
      i = index(input_restart_file, '.r.')
      if (i>0) then
         stoch_restfile = input_restart_file(1:i)//'r_stoch'//input_restart_file(i+2:k)

         if (is_root_pe()) then
           write(stdout,*) 'input_restart_file =', input_restart_file
           write(stdout,*) 'stoch_restfile =', stoch_restfile
         endif
      endif
  endif
```

Are [lines 290-302](https://github.com/chengz2/MOM6/blob/97d5eb8f601706c446e312d3df35bd32b3c8ded2/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90#L290-L302)

src/parameterizations/vertical/MOM_CVMix_KPP.F90
gustavo-marques: I am reviewing this code with Bill Large. Bill's reason for going only to `iFaceHeight(GV%ke)` is to allow one to use `ENHANCE_DIFFUSION = True`, in which case the shear across' iFaceHeight(GV%ke+1) ' is needed and won't be available. So, if we use 

`zBottomMinusOffset = iFaceHeight(GV%ke+1) + min( max(CS%deepOBLoffset,0.0), -0.1*iFaceHeight(GV%ke+1))`

and `ENHANCE_DIFFUSION = True`, there could be a potential for the code to break.

jiandewang:
> @jiandewang If you add `#ifdef CESMCOUPLED` around the following blocks in `config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90`, does that let you build?
> 
> ```fortran
> use stochy_data_mod,         only : stoch_restfile
> ```
> 
> is [line 67](https://github.com/chengz2/MOM6/blob/97d5eb8f601706c446e312d3df35bd32b3c8ded2/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90#L67) on our branch, and
> 
> ```fortran
>   if(present(input_restart_file)) then
>       k = index(input_restart_file, ' ')
>       if (k==0) k = len_trim(input_restart_file)
>       i = index(input_restart_file, '.r.')
>       if (i>0) then
>          stoch_restfile = input_restart_file(1:i)//'r_stoch'//input_restart_file(i+2:k)
> 
>          if (is_root_pe()) then
>            write(stdout,*) 'input_restart_file =', input_restart_file
>            write(stdout,*) 'stoch_restfile =', stoch_restfile
>          endif
>       endif
>   endif
> ```
> 
> Are [lines 290-302](https://github.com/chengz2/MOM6/blob/97d5eb8f601706c446e312d3df35bd32b3c8ded2/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90#L290-L302)

@mnlevy1981 this made my compiling works. However I noticed that for those tests that stochastoc physics is involved, final restart file "ocn_stoch.res.nc" are written outside of RESTART directory which shouldn't.

iangrooms:
@jiandewang @mnlevy1981 the problem seems to be [this line](https://github.com/NCAR/MOM6/blob/e10fe20a9531440cdc56efa00aaa294ed85e199a/config_src/drivers/nuopc_cap/mom_cap.F90#L2046-L2047)
```
        call ocean_model_restart(ocean_state, restartname=restartname, &
                                stoch_restartname=stoch_restartname, num_rest_files=num_rest_files)
```

When we made the original changes to write restart files in CESM we explicitly added the 'RESTART/' at the beginning to the stochastic restart file name so that it would get written in the RESTART directory. The code, which can be found [here](https://github.com/NCAR/MOM6/pull/376/changes), read
```
        call ocean_model_restart(ocean_state, restartname=restartname, &
                                stoch_restartname='RESTART/'//stoch_restartname)
```
In [the latest merge from mom-ocean to NCAR](https://github.com/NCAR/MOM6/pull/417) the RESTART/ part of the filename got wiped out.  I'm guessing that the code above won't work for a multi-instance case though, so possibly the right solution is
```
        call ocean_model_restart(ocean_state, restartname=restartname, &
                                stoch_restartname='RESTART/'//stoch_restartname, num_rest_files=num_rest_files)
```
That said, the code on mom-ocean currently doesn't have an explicit 'RESTART/' in the file name, so I don't know why we need to add it in for this PR.

Separate from the problem with the stochastic restart being written outside the RESTART directory: While @mnlevy1981's `#ifdef` fix is a short-term solution, I think a better solution is to send our updates to the stochastic physics package in a PR to  NOAA (which should be bit-for-bit). When I said earlier ' does this MOM6 PR comes with an associated stochastic_physics PR ? No. (At least not that I recall.)' I was wrong.

jiandewang:
_In https://github.com/NCAR/MOM6/pull/417 the RESTART/ part of the filename got wiped out. I'm guessing that the code above won't work for a multi-instance case though_

right, this is how I found out the issue. In UFS we have a pair of "restart" tests in which 2nd job relies on the restart file from 1st job.

I think the stochastic physics branch being used in testing for this MOM6 PR on NCAR is out of dated though 3 new commits were being added. We need a clean sotchastic physics branch to test in UFS and NCAR side.

src/parameterizations/vertical/MOM_CVMix_KPP.F90
Hallberg-NOAA: Thanks for that explanation, @gustavo-marques.

I have to admit that this argument does not make sense to me, mostly because I do not understand the detailed inner workings of `CVMix_kpp_compute_OBL_depth()` or `cvmix_kpp_compute_enhanced_diff()` inside of CVmix_kpp.F90.  However, I can definitely say that (1) the suggested reversion of the `iFaceHeight` index to `iFaceHeight(GV%ke+1)` in the definition of `zBottomMinusOffset` is necessary to recover the previous answers, and (2) the addition of a massless layer at the bottom of MOM6 should never change answers substantially.  Is it possible that the real problem here stems from inadequate error handling inside of CVmix_kpp.F90 that would calculate the shear with a one-sided calculation as though a massless zero velocity (0 because of the no-slip BC) layer did exist below the OBL in cases where it is necessary to do so?

If there is a compelling reason for `zBottomMinusOffset` as passed to `CVMix_kpp_comput_OBL_depth` to be redefined as here to be higher by the thickness of the bottommost layer, we should perhaps reconsider the use of this same variable to impose a limit on the boundary layer depth on line 1502 so that it does not change answers even in cases that are not using Langmuir turbulence.

It is still not clear in my mind whether this change of indexing is actually fixing a bug or introducing a bug, or perhaps both. However, now that we know that we can recover the previous answers with changes inside of MOM6, I think that this PR should respect the usual MOM6 practice of not changing other people's answers without their consent and that we should provide a runtime flag to recover past answers whenever possible and those past answers are well defined.  If it really is the case that this is a bug that is being fixed, the correct procedure would be to introduce a bug-fix flag to correct that bug (keeping the previous answers by default), then change the default via the MOM6 parameter review process, and then (after perhaps a year) obsolete the bug-fix flag altogether if no one objects in a second pass through the parameter review process.

sanAkel:
> _In [NCAR#417](https://github.com/NCAR/MOM6/pull/417) the RESTART/ part of the filename got wiped out. I'm guessing that the code above won't work for a multi-instance case though_
> 
> right, this is how I found out the issue. In UFS we have a pair of "restart" tests in which 2nd job relies on the restart file from 1st job.
> 
> I think the stochastic physics branch being used in testing for this MOM6 PR on NCAR is out of dated though 3 new commits were being added. We need a clean sotchastic physics branch to test in UFS and NCAR side.

@jiandewang and others, we really need to step back and re-think why we must have any restart issues at all?

Since this PR is probably _ready_ for review:
- Leaving a comment (not raising anything that needs to be resolved before merging). 
- We should (later) revisit this entire sto-phy handle at the top,
- So that if we build with a stub (as @mnlevy1981 said) we must [not have anything to fuss with](https://github.com/mom-ocean/MOM6/pull/1700#issuecomment-4452521022).

kshedstrom (approved):

iangrooms:
@jiandewang I was wrong when I said that I didn't think we needed a stochastic physics PR. NCAR is using the ESCOMP fork of the stochastic physics package, which as you and @mnlevy1981 say has diverged from the version that UFS is using. I pulled all the latest updates from NOAA-PSL/stochastic_physics into the NCAR version, then verified that it still passes NCAR's tests. The code is [here](https://github.com/ESCOMP/stochastic_physics/pull/6). If you test with the version of stochastic-physics in my PR it should hopefully reproduce your current answers. (You won't be able to use my PR directly: you'll need to change `halo_exchange_fv3.F90` back to `halo_exchange.fv3.F90` in a few places.)

If you compare my `noaa-sync` branch to the master branch of NOAA-PSL, you'll see that I have not changed the API of the stochastic physics package nor its internal functionality. The only change that I made is in how the restart filename is handled, specifically the location where the stochastic physics package _reads_ its restart file. It used to be hard-coded within the package with a filename that doesn't work for CESM. I changed it so that the restart filename is now a public module variable. The default value is the value that used to be hard-coded, but now CESM can 'use' and change the restart filename to whatever we need it to be.

My stochastic-physics PR code should at least allow you to compile. I'm not 100% sure that the stochastic physics package will _write_ its restart file in the right place, but that's a different part of the code.

jiandewang:
@iangrooms Now I point to your stochastic physics branch in my UFS testing, can you tell me the exact code/places that I need to make change ?

sanAkel:
> If you test with the version of stochastic-physics in my PR it should hopefully reproduce your current answers. (You won't be able to use my PR directly: you'll need to change `halo_exchange_fv3.F90` back to `halo_exchange.fv3.F90` in a few places.)


@iangrooms what is the connection with `halo_exchange_fv3.F90`? If it a long answer, we can take it offline.

iangrooms:
The only connection is that the CESM build system doesn't like the file name `halo_exchange.fv3.F90` so we change the dot to an underscore. It's the same file.

sanAkel:
> The only connection is that the CESM build system doesn't like the file name `halo_exchange.fv3.F90` so we change the dot to an underscore. It's the same file.

Thanks @iangrooms Hopefully they can teach their file parser some new tricks!

chengz2:
Thanks everyone for the help, this PR is now ready for review. The code can now recover the previous answers, and has passed all dimensional consistency tests.

jiandewang:
> @iangrooms Now I point to your stochastic physics branch in my UFS testing, can you tell me the exact code/places that I need to make change ?

@iangrooms I saw "halo_exchange_fv3.F90" in your branch (only one file), why do you say "you'll need to change halo_exchange_fv3.F90 back to halo_exchange.fv3.F90 **in a few places**." ?

iangrooms:
@jiandewang I think this is the complete list:
- in CMakeLists.txt at line 30
- in makefile at line 47
- in unit_tests/compile_standalone_ca.hera.gnu at line 36
- in unit_tests/compile_standalone_ca.hera.intel at line 32
Also you should rename the file `halo_exchange_fv3.F90` to `halo_exchange.fv3.F90`

jiandewang:
@iangrooms in UFS we don't have fork for stochastic physics, instead we are using PSL's master, see https://github.com/ufs-community/ufs-weather-model/blob/develop/.gitmodules#L11. So I am going to use your PR branch for my test which will make things easy as there is no need for me to modify cmake list et al as all changes are contained in your branch.

jiandewang:
@iangrooms use your stochastic physics branch made compiling working fine but "ocn_stoch.res.nc" files are written outside of RESTART file directory. I recall about half year ago you, me and Phil worked together and fixed that problem. Apparently your fixing was being removed here.

jiandewang:
a typo at https://github.com/chengz2/MOM6/blob/ncar-to-main_2604/config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90#L455 (Thanks @DeniseWorthen)
not worth to have a hash for this but can be done while you fix the RESTART dir issue.

iangrooms:
I fixed the typo and put the 'RESTART/' back. Code is [here](https://github.com/iangrooms/MOM6/tree/ncar-to-main_2604).

jiandewang:
@iangrooms can you have a PR to merge your fixing to @chengz2  [branch](https://github.com/chengz2/MOM6/tree/ncar-to-main_2604) ? I will run full UFS regression test after that.

chengz2:
> @iangrooms can you have a PR to merge your fixing to @chengz2 [branch](https://github.com/chengz2/MOM6/tree/ncar-to-main_2604) ? I will run full UFS regression test after that.

It has been merged.

jiandewang:
it passed UFS regression test. @iangrooms note we have neither code manger for stochastic physics nor fork of it. So in case you need to make last minute change in stochastic physics, be sure to let me know and I will re-test it.

jiandewang (approved):

config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90
```diff
@@ -283,6 +287,19 @@ subroutine ocean_model_init(Ocean_sfc, OS, Time_init, Time_in, gas_fields_ocn, i
   call time_interp_external_init
 
   OS%Time = Time_in
+  if(present(input_restart_file)) then
```
Hallberg-NOAA: MOM6 uses a 2-space indenting convention, rather than the 3- or 4-spaces used here, as documented in the MOM6 style guide at https://github.com/mom-ocean/MOM6/wiki/Code-style-guide.  Also from the style guide, `if (present(input_restart_file)) then` would be strongly preferred over `if(present(input_restart_file)) then`.

config_src/drivers/nuopc_cap/mom_ocean_model_nuopc.F90
```diff
@@ -283,6 +287,19 @@ subroutine ocean_model_init(Ocean_sfc, OS, Time_init, Time_in, gas_fields_ocn, i
...
+      k = index(input_restart_file, ' ')
+      if (k==0) k = len_trim(input_restart_file)
+      i = index(input_restart_file, '.r.')
+      if (i>0) then
+         stoch_restfile = input_restart_file(1:i)//'r_stoch'//input_restart_file(i+2:k)
+
+         if (is_root_pe()) then
+           write(stdout,*) 'input_restart_file =', input_restart_file
```
Hallberg-NOAA: For most output, the `MOM_mesg()` or `MOM_error()` interfaces are strongly preferred, both because they facilitate the redirection of the output and (more importantly) because they work with the verbosity settings to regulate the volume of output.  It would be strongly preferable to revise this block to read
```
      call MOM_mesg("input_restart_file = "//trim(input_restart_file), 3)
      call MOM_mesg("stoch_restfile  = "//trim(stoch_restfile), 3)
```
where the 3 in these lines indicates the verbosity level of these messages to be typical of informational messages.  These messages are only issued from the root PE (unless an `all_print` optional argument indicates otherwise), so these calls do not need to be inside of an `if (is_root_pe()) then` test.  If this revision is adopted, `MOM_mesg` will also need to be added to the module use statement for `MOM_error_handler` on line 27 of this file.

src/parameterizations/lateral/MOM_hor_visc.F90
```diff
@@ -2664,6 +2699,19 @@ subroutine hor_visc_init(Time, G, GV, US, param_file, diag, CS, ADp)
                  "If true, Ah and m_leithy are smoothed within Leith+E.  This requires "//&
                  "lots of blocking communications, which can be expensive", &
                  default=.true., do_not_log=.not.CS%use_Leithy)
+  call get_param(param_file, mdl, "TAPER_LEITHY", CS%taper_leithy, &
+                 "If true, Leith+E c_K coefficient is tapered to zero"//&
```
Hallberg-NOAA: A space should be added to `zero "//&` so that the parameter description reads "If true, Leith+E c_K coefficient is tapered to zero below a threshold depth" instead of "If true, Leith+E c_K coefficient is tapered to zerobelow a threshold depth"

src/parameterizations/lateral/MOM_hor_visc.F90
```diff
@@ -2951,7 +3002,9 @@ subroutine hor_visc_init(Time, G, GV, US, param_file, diag, CS, ADp)
       if (CS%use_Leithy) then
         CS%biharm6_const_xx(i,j) = Leith_bi_const * max(G%dxT(i,j),G%dyT(i,j))**6
         CS%m_const_leithy(i,j) = 0.5 * sqrt(CS%c_K) * max(G%dxT(i,j),G%dyT(i,j))
-        CS%m_leithy_max(i,j) = 4. / max(G%dxT(i,j),G%dyT(i,j))**2
+        CS%m_leithy_max(i,j) = 4. / max(G%dxT(i,j),G%dyT(i,j))**2 * &
```
Hallberg-NOAA: I think that it would be better to write this line as `CS%m_leithy_max(i,j) = (4. / max(G%dxT(i,j),G%dyT(i,j))**2) * &` so that a person reading it does not have to remind themselves whether multiplication in Fortran is left-associative or right-associative, and hence whether the masks are in the numerator or denominator.

src/parameterizations/stochastic/MOM_stochastics.F90
```diff
@@ -242,54 +265,71 @@ subroutine stochastics_init(dt, grid, GV, CS, param_file, diag, Time)
...
+  ! Initialize the smoothing weights
+  if ((CS%do_skeb) .and. CS%skeb_npass >= 1) then
+    ALLOC_(CS%area_wt(grid%isd:grid%ied,grid%jsd:grid%jed)) ; CS%area_wt(:,:) = 0.0
+    ALLOC_(CS%Isum_area_wts(grid%isd:grid%ied,grid%jsd:grid%jed)) ; CS%Isum_area_wts(:,:) = 0.0
+    do j=grid%jsc-2,grid%jec+2 ; do i=grid%isc-2,grid%iec+2
+      CS%area_wt(i,j) = grid%mask2dT(i,j)*grid%areaT(i,j)
+    enddo ; enddo
+    do j=grid%jsc-1,grid%jec+1 ; do i=grid%isc-1,grid%iec+1
+      sum_area_wts = CS%area_wt(i,j) + &
+       (((CS%area_wt(i-1,j) + CS%area_wt(i+1,j)) + (CS%area_wt(i,j-1) + CS%area_wt(i,j+1))) + &
```
Hallberg-NOAA: The MOM6 style guide at https://github.com/mom-ocean/MOM6/wiki/Code-style-guide strongly encourages an indent by at least 4 spaces for continuation lines.  Lines 277 and 278 should be indented by another 3 spaces to adhere to this convention.

chengz2:
Thanks @Hallberg-NOAA, the formatting issues you mentioned are all fixed.

Hallberg-NOAA (approved):
I have run the latest version through the GFDL test suite, and all solutions are identical, provided that KPP%ANSWER_DATE is set between 20240501 and 20251231.  After this PR is merged into the main branch of MOM6, we plan to update our KPP-based test solutions to use the latest answer dates.

mathomp4:
In re NASA/GMAO, I did some testing at using our very-low-res case and found it to be zero-diff. I'll let @sinakhani comment if he saw anything at higher resolution

abozec (approved):
COAPS approves

dougiesquire (approved):
Apologies for the hold up. I've reviewed the code changes and confirmed that they do not change answers for our configurations. ACCESS-NRI approves.

sinakhani:
> In re NASA/GMAO, I did some testing at using our very-low-res case and found it to be zero-diff. I'll let @sinakhani comment if he saw anything at higher resolution

I tested this PR in a higher-resolution GEOS-MOM6 configuration. We approve it (this does not change our answers).

mathomp4 (approved):

marshallward:
@chengz2 We have approvals from everyone so you are welcome to merge when you are ready.

Hallberg-NOAA:
The two checks that are flagged as failing are due to changes in diagnostics, and not any changes in the solutions themselves.  Most of these changes are due the addition of new diagnostics, but in particular the S_SURF diagnostic is changing because the point where this diagnostic is posted has been moved from `post_tracer_transport_diagnostics()`, which is called after tracer advection but before diabatic forcing is applied, into the new routine `post_tracer_integral_diagnostics()`, which is called after tracer advection, diabatic forcing and vertical remapping have all been applied.  I am convinced that these changes in diagnostics are correct and should not be a barrier to the acceptance of this PR.

We are still waiting for an assessment of this PR from the Korean MOM6 User's Group (K-MUG).  This evaluation is underway, and assuming that no new issues are identified, this PR will be ready to be merged into the main branch of MOM6 once we have the results of this assessment.

Hallberg-NOAA:
Please hold on merging in this PR.  We have answer changes in the Korean NW Pacific regional configuration that we are tracking down.

theresa-cordero:
We now have confirmation that there are no answer changes with the Korean NW Pacific regional configuration and approval from K-MUG to merge this PR.

Hallberg-NOAA:
@chengz2, would you like to please do the honor of clicking the "Merge pull request button" for https://github.com/mom-ocean/MOM6/pull/1700?

chengz2:
> @chengz2, would you like to please do the honor of clicking the "Merge pull request button" for #1700?

Sorry, I don't think I can do it on my end. Please feel free to merge it if you don't see any problem.

PR #1707, opened 2026-07-23
Add ability to smoothly turn off the MLE scheme for deep mixed layers

iangrooms:
This PR adds the ability to 'taper' the mixed layer and boundary layer depths in the MLE scheme. This is controlled by two new parameters: `MLE_TAPER_FN_DEPTH` and `MLE_TAPER_FN_POWER`.  When both new parameters are positive, the depth that the MLE scheme sees, denoted $H$, is replaced by

$$\frac{H}{1 + (H/H_0)^P}$$

where $P$ is `MLE_TAPER_FN_POWER` and $H_0$ is `MLE_TAPER_FN_DEPTH`. In the OM4 MLE scheme this is applied to $H=$ `MLD_fast`. In the Bodner MLE scheme this is applied to both `big_H` and `little_h`. For very deep mixed layers $H\gg H_0$ this effectively turns off the MLE restratification scheme, both by reducing the amplitude of the overturning (which scales with $H$) and by reducing the depth $H$ over which the parameterization is applied.

Note that this function does not directly modify any state variables, nor does it directly change the mixed layer depth. It only changes what the MLE scheme thinks is the mixed layer depth.

### DIFF
--- a/src/parameterizations/lateral/MOM_mixed_layer_restrat.F90
+++ b/src/parameterizations/lateral/MOM_mixed_layer_restrat.F90
@@ -68,6 +68,11 @@ type, public :: mixedlayer_restrat_CS ; private
                                    !! the mixed-layer [nondim].
   real    :: MLE_MLD_stretch       !< A scaling coefficient for stretching/shrinking the MLD used in
                                    !! the MLE scheme [nondim]. This simply multiplies MLD wherever used.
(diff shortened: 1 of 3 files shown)


iangrooms:
Sorry, meant to point this to NCAR, not mom-ocean.

PR #1709, opened 2026-08-20
GitHub action version update

marshallward:
This should eliminate the Node 20 warnings in the Actions CI.

This only affect GitHub Actions so does not need regression testing.


marshallward:
oops

PR #1714, opened 2026-09-10
Remove `to` from `!$omp declare target` for mapping parameter arrays

edoyango:
This fixes compilation errors with `amdflang`.

I initially thought this was a bug, but flang has rightfully dropped support of `!$omp declare target to` since it's been dropped in the OpenMP standard. 

Hence, this PR drops the `to` and uses the form that is accepted by both AMD and NVIDIA. 

Confirmed to work (i.e. produce correct anwers) with NVIDIA. For AMD, only checked that it passes the ICE that was being encountered. 

Should also mention that `!$omp declare target` is being used here to map `parameter` variables to the device, but we shouldn't need to. There was a similar question on the NVIDIA forums and Mat opened a bug report for the compiler team. https://forums.developer.nvidia.com/t/offloading-declare-target-and-parameter-in-fortran/342119/2


edoyango:
Opened PR with wrong repo as base

